PR #10, opened 2016-08-13
Don't catch and handle errors in the controllers

danzel:
At the moment if a controller throws an error, tsoa catches this and returns some details of the error to the client.
This stops express error loggers from catching it, and makes customising this behaviour difficult to impossible.
Instead I think we should use the standard express error handling, and then the client app can override the error output if they'd like.

ref: https://expressjs.com/en/guide/error-handling.html

Maybe we should change the getValidatedParams wrapper too, but this is enough for me for now :)

### DIFF
--- a/src/routeGeneration/templates/express.ts
+++ b/src/routeGeneration/templates/express.ts
@@ -2,7 +2,7 @@ export const expressTemplate = `
 export function RegisterRoutes(app: any) {
     {{#each controllers}}
     {{#each actions}}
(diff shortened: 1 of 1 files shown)


lukeautry:
@danzel Thanks for this, it makes a lot of sense.

I merged these changes and made some additional changes on top so that validation errors get passed to the custom error handler as well: https://github.com/lukeautry/tsoa/commit/a9128b7596f8c2abc587412a09607747afd31b38

Issue #13, opened 2016-09-11
Support for class-based model


gms1:
part of my model is class-based ( because it need to be decorated for the object-relational mapping and should have a default constructor ) and 
although I would be able to declare additional interfaces, this would lead to duplicating all the property definitions of the classes to the corresponding interfaces.

Do you have any plan to support a class-based model?

lukeautry:
@gms1 I've opened up a pull request to support this behavior: #14 

Since you opened this issue, please take a look at those changes and provide any feedback. I've added a suite of tests for these changes and done some manual testing, but I'd like your input.

Thanks!

lukeautry:
@gms1 This has been released and merged in 0.0.61.

PR #14, opened 2016-09-12
Class-based model generation

lukeautry:
Today, only interfaces can be used as controller method return types or param types. As #13 points out, there is a valid use case for using classes for definitions.

This revision implements class-based model generation in the following ways.

TypeScript classes can be used as return type models, e.g.

```
@Route('GetTest')
export class GetTestController {

  @Get('ClassModel')
  public async getClassModel(): Promise<TestClassModel> {
    return new ModelService().getClassModel();
  }
}

export class TestClassBaseModel {
  public id: number;
}

/**
 * This is a description of TestClassModel
 */
export class TestClassModel extends TestClassBaseModel {
  /**
   * This is a description of a public string property
   */
  public publicStringProperty: string;
  public optionalPublicStringProperty?: string;
  protected protectedStringProperty: string;

  /**
   * @param publicConstructorVar This is a description for publicConstructorVar
   */
  constructor(
    public publicConstructorVar: string,
    protected protectedConstructorVar: string,
    public optionalPublicConstructorVar?: string
  ) {
    super();
  }
}
```

Classes can also be used (and thus validated) in requests to the server:

```
@Route('PostTest')
export class PostTestController {
  @Post('WithClassModel')
  public async postClassModel(model: TestClassModel): Promise<TestClassModel> {
    const augmentedModel = new TestClassModel('test', 'test2', 'test3');
    augmentedModel.id = 700;

    return augmentedModel;
  }
}
```

Generation of swagger descriptions works in similar fashion to interfaces, where jsdoc comments are parsed and added to the spec.

Constructor variables (e.g. `constructor(public readonly myProperty: string)` are treated as properties and are added to the spec. Only public properties and constructor properties are added.

## Caveats

One 'gotcha' is, when making a request to the server using the generated routes where the contract requires an instance of a class in the request (see the second example above), that class is *not* instantiated, which could lead to some deceptive behavior. From the body of your controller method, you may actually work with your instance as if it were a fully bonified class. In reality, you're working with a POJO with no methods or other class behavior.

I've updated the readme to state that interfaces as models are _preferred_. Classes may make sense to use in some cases, but in the majority of cases, I think interfaces are a better representation of the data exchange that occurs between client and server.

### DIFF
--- a/README.MD
+++ b/README.MD
@@ -19,7 +19,7 @@
 - If regular type annotations aren't an appropriate way to express metadata, use decorators
 - Use jsdoc for pure text metadata (e.g. endpoint descriptions)
 - Minimize boilerplate
-- Models are always represented by interfaces; the contract between API client is best represented through a data structure, not a class or object
+- Models are best represented by interfaces (pure data structures), but can also be represented by classes
 
 ## How it works
 
--- a/src/decorators/example.ts
+++ b/src/decorators/example.ts
@@ -1,3 +1,3 @@
 export function Example<T>(exampleModel: T): any {
-  return null;
+  return () => { return; };
 }
--- a/src/decorators/methods.ts
+++ b/src/decorators/methods.ts
@@ -1,19 +1,19 @@
 export function Get(value?: string): any {
-  return;
+  return () => { return; };
 }
 
 export function Post(value?: string): any {
-  return;
+  return () => { return; };
 }
 
 export function Patch(value?: string): any {
-  return;
+  return () => { return; };
 }
 
 export function Put(value?: string): any {
-  return;
+  return () => { return; };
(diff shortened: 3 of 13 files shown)


gms1:
Thank you very much! I will test this further on weekend.
Anyway, the generated swagger.json and routes.ts are looking promising, but properties without explicit modifier are currently not included in the swagger definition, although they are public by default

lukeautry:
Good find, didn't think about that. Will update this PR tonight probably.

lukeautry:
@gms1 Pushed additional updates to treat non-access modifier properties as public as expected.

gms1:
tested; please pull your changes
Thanks!

PR #17, opened 2016-09-14
Return functions from decorator factories

lukeautry:


gms1:
please pull this changes ( e.g. reflect-metadata library requires this )

Thanks!

PR #21, opened 2016-10-19
add an opportunity to add various tags to routes methods of swagger.json

strictlogistic:
Hi, Luke.

I added an opportunity to add various tags to swagger.json file methods.
For example there was not any ways to add "tags" property to any methods to separate our api methods in swagger. Now we can just add @tags ["users", "accounts"] to our controller method jsdoc description and they will appear into swagger.json.

Regards, Anatoly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "sl-tsoa",
   "description": "Build swagger-compliant REST APIs using TypeScript and Node",
-  "version": "0.0.63",
+  "version": "0.1.0",
   "main": "./dist/index.js",
   "typings": "./dist/index.d.ts",
   "keywords": [
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "sl-tsoa",
   "description": "Build swagger-compliant REST APIs using TypeScript and Node",
-  "version": "0.1.0",
+  "version": "0.1.1",
   "main": "./dist/index.js",
   "typings": "./dist/index.d.ts",
   "keywords": [
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "tsoa",
+  "name": "sl-tsoa",
   "description": "Build swagger-compliant REST APIs using TypeScript and Node",
   "version": "0.0.63",
   "main": "./dist/index.js",
@@ -33,7 +33,7 @@
     "handlebars": "^4.0.5",
     "mkdirp": "^0.5.1",
     "moment": "^2.14.1",
-    "typescript": "^2.1.0-dev.20160910",
+    "typescript": "^2.0.3",
     "typescript-formatter": "^2.3.0",
     "yargs": "^4.8.1"
   },
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
-  "name": "sl-tsoa",
+  "name": "tsoa",
   "description": "Build swagger-compliant REST APIs using TypeScript and Node",
-  "version": "0.1.1",
+  "version": "0.0.63",
   "main": "./dist/index.js",
   "typings": "./dist/index.d.ts",
   "keywords": [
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
-  "name": "tsoa",
+  "name": "sl-tsoa",
```
lukeautry: Let's keep this just 'tsoa' since once it gets merged in, it won't be in your fork

package.json
```diff
@@ -33,7 +33,7 @@
     "handlebars": "^4.0.5",
     "mkdirp": "^0.5.1",
     "moment": "^2.14.1",
-    "typescript": "^2.1.0-dev.20160910",
+    "typescript": "^2.0.3",
```
lukeautry: Prefer to keep this at typescript@next (i.e. the version that was there)

src/metadataGeneration/extraTagsGenerator.ts
```diff
@@ -0,0 +1,69 @@
+import { Type, Parameter } from './metadataGenerator';
+import * as ts from 'typescript';
+import * as tsserver from '../tsserver';
+
```
lukeautry: Looks like an extra space is here; make sure to run the lint task to catch any other problems.

src/routeGeneration/routeGenerator.ts
```diff
@@ -31,18 +31,18 @@ export class RouteGenerator {
   private buildContent(middlewareTemplate: string) {
     let canImportByAlias: boolean;
     try {
-      require('tsoa');
+      require('sl-tsoa');
```
lukeautry: Just 'tsoa'

src/routeGeneration/routeGenerator.ts
```diff
@@ -31,18 +31,18 @@ export class RouteGenerator {
...
     }
 
     const routesTemplate = handlebars.compile(`
+            /* tslint:disable */
             /**
              * THIS IS GENERATED CODE - DO NOT EDIT
              */
-            /* tslint:disable */
-            import {ValidateParam} from '${canImportByAlias ? 'tsoa' : '../../src/routeGeneration/templateHelpers'}';
+            import {ValidateParam} from '${canImportByAlias ? 'sl-tsoa' : '../../src/routeGeneration/templateHelpers'}';
```
lukeautry: Just 'tsoa'

src/tsserver.d.ts
```diff
@@ -0,0 +1,12 @@
+import * as ts from 'typescript';
```
lukeautry: I'm surprised that these typings don't already exist somewhere in the TypeScript project. Do we have to use these custom ambient types?

lukeautry:
@strictlogistic Thanks for the PR - I made a few comments.

I haven't used tags w/ swagger before - can you attach a copy of the swagger.json that these changes allow you to generate? I'd like to understand a bit more about the final output.

It would also be nice to include some tests with these - should be really simple to put something in definitions.spec.ts - you'll see either tests in there with a pattern you can follow quite easily.

Let me know if you have any questions and thanks again.

lukeautry:
@strictlogistic  Just wanted to follow up on my previous comments. Actually what would be better than a copy of swagger.json is a short addition to the README explaining how to use this feature with examples

Thanks again!

package.json
andreasrueedlinger: Why are you insisting on typescript@next? When I checked out your project I realized that it is quite unstable if not using the exact version you specified (typescript and typescript-formatter). Did you have similar issues or do you just have one version of your project installed which builds ever since?

andreasrueedlinger: Here are my observations:
* typescript newer than 2.1.0-dev.20160910 at some time changes the formatOptions from uppercase to lowercase which breaks with certrain version combination of typescript-formatter and `unformats` all the source files
* typescript newer than 2.1.0-dev.20160910 introduced a bug in jsdoc parsing (https://github.com/Microsoft/TypeScript/issues/10987) with multiline comments which makes your multiline test break

Issue #23, opened 2016-11-09
Typescript Union Types Fail Model Validation


shaun-mason:
If I declare a property in a model with a union type:

```
export class Example {
     public $key?: string | null = null;
}
```

The resulting model is generated as such:

` '$key': { typeName: 'object', required: false }`

Since there is no corresponding 'object' model definition, an error results at run time.

lukeautry:
Merged pull request, live in version 1.0.2.

PR #24, opened 2016-11-09
Added guard to handle union types.

shaun-mason:
Added a guard to bypass validation of union types that result in type 'object'.

See Issue #23


lukeautry:
Good find @shaun-mason

AmazingTurtle:
@shaunlmason the guard doesn't seem to work with `UnionType[]`

Issue #7, opened 2016-07-26
Providing configuration to controllers


danzel:
I'd like to pass some settings to my controllers to configure them correctly (db connection details as an example).

Is there a nice way to do this? My current thought is to have them require a static settings object that I set up externally.

Ideally I'd like to do this by passing an object to their constructor, but if there was a post construction init method that'd probably be fine.

Thanks :)

lukeautry:
@danzel I've thought about this, and there's more than db connection details that this might be useful for.

Really, what we're talking about is dependency injection. This could be the raw node.js request, it could be a data access service, anything. Either way, the controller should be fully testable without any knowledge of your middleware or w/e. It also allows you to do things like oauth without having to bake it into the library.

I would like to implement decorator based injection, something like this:

`
@Inject(UserService, Request)
@Route('Users')
export class MyController {

     constructor(private UserService: UserService, private Request: Request) {

     }

     @Get()
     public async getUsers(): Promise<User[]> {
          return this.UserService.getUsers();
     }
}
`

Interested to hear your thoughts on this.

danzel:
That looks great!

lukeautry:
@danzel I looked into this a bit, and it quickly became obvious that I'd basically be creating a dependency injection framework, which isn't the road I want to go down.

That said, I think it's very possible to get an existing DI framework to work within our context. Based on a little bit of research, [Inversify](http://inversify.io/) seems like a robust option.

My plan is to implement something like Inversify in [the example project](https://github.com/lukeautry/tsoa-example) that we can reference going forward.

danzel:
Awesome.
My static vars work fine for now (and are fine for production TBH)

andreasrueedlinger:
I am also very interested in IOC-support. Have you looked more into it?
What would be the suggested way to go? Write an express-inversify template for the routes where the controllers are created by the kernel of inversify?

andreasrueedlinger:
Also, I would like to get a possibility to get access to the request object in the controller methods somehow. For example when using some kind of authorization which stores something in the request context or some other context information.

lukeautry:
@andreasrueedlinger Since two folks are asking for it, I think it's worth taking another stab at it.

> Write an express-inversify template for the routes where the controllers are created by the kernel of inversify

I think so.

> Also, I would like to get a possibility to get access to the request object in the controller methods somehow.

The way to do this now is to set up a singleton-type object (probably an exported namespace/variable) that can be accessed statically.

[...]

PR #25, opened 2016-11-10
Add support for injected express request and other values

andreasrueedlinger:
Here is a proposition how to implement support for express request parameter in controllers as well as other custom values injected by a custom template.
See also my proposition/question [here](https://github.com/lukeautry/tsoa/issues/7)

### DIFF
--- a/README.MD
+++ b/README.MD
@@ -104,6 +104,28 @@ app.listen(3000);
 
 ```
 
+### Get access to the request object of express in Controllers
+
+To access the request object of express in a controller method use the `@Request`-decorator:
+```typescript
+// controllers/usersController.ts
+
+import * as express from 'express';
+import {Get, Route, Request} from 'tsoa';
+import {User, UserCreationRequest} from '../models/user';
+
+@Route('Users')
+export class UsersController {
+    @Get('{id}')
+    public async getUser(id: number, @Request() request: express.Request): Promise<User> {
+        // TODO: implement some code that uses request as well
+    }
+}
+```
+Note that the parameter `request` does not appear in your swagger definition file. 
+Likewise you can use the decorator `@Inject` to mark a parameter as being injected manually and should be omitted in swagger generation.
+In this case you should write your own custom template where you inject the needed objects/values in the method-call. 
+
 ### Use awesome Swagger tools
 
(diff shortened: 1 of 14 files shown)


lukeautry:
@andreasrueedlinger Thanks for this change, I will review some time in the next 24 hours.

lukeautry:
@andreasrueedlinger Sorry, took longer to get to this than I thought.

Two things:

1) Would you mind adding a quick section to the readme with usage for Request/Inject?
2) Let's expose the new decorators in https://github.com/lukeautry/tsoa/blob/master/src/index.ts

Thanks!

andreasrueedlinger:
Thanks for the feedback. I added your 2 suggestions.

lukeautry:
@andreasrueedlinger Thanks; this is live in 1.0.3.

danzel:
Awesome!
I'm not quite sure how to use the `@Inject` property however, do I need to edit this template:
https://github.com/lukeautry/tsoa/blob/master/src/routeGeneration/templates/express.ts
to do the injecting?

andreasrueedlinger:
@danzel If you want to use `@Inject` you need to write your own template (using the one you referenced as base).
Then you can perform your injection in this section:
```ts
      function getValidatedParams(params: any, request: any, bodyParamName?: string): any[] {
        const requestParams = getRequestParams(request, bodyParamName);

        return Object.keys(params).map(key => {
            if (params[key].injected === 'inject') {
              if (key === 'nameOfYourInjectedParameter') {
                return 'yourValueOrObjectHere';
              }
              return undefined;
            } else if (params[key].injected === 'request') {
              return request;
            } else {
              return ValidateParam(params[key], requestParams[key], models, key);
            }
        });
```

What is your use-case of parameter injection? Normally you should be fine to use normal dependency injection with inversify or similar. In my case I adapted the express template as follows:
```ts
// import the inversify kernel (this is unfortunately custom code, otherwise we could add an inversify-js template)
import { kernel } from './ioc/ioc';

export function RegisterRoutes(app: any) {
  // ...

  // instead of `const controller = new {{../name}}();`
  const controller = kernel.get<{{../name}}>({{../name}});
  // ...
}

danzel:
Awesome thanks, yeah that is what I was trying to achieve :)

PR #33, opened 2017-01-03
fix for more recent typescript version + stricker checks

egandro:
Recent versions of typescript (+it's typings) have stricker checks on handling nulls for strings.

### DIFF
--- a/src/routeGeneration/templateHelpers.ts
+++ b/src/routeGeneration/templateHelpers.ts
@@ -17,11 +17,11 @@ export function ValidateParam(typeData: any, value: any, generatedModels: any, n
     case 'string':
       return validateString(value, name);
     case 'boolean':
-      return validateBool(value, name);
+      return validateBool(value, <any>name);
     case 'number':
-      return validateNumber(value, name);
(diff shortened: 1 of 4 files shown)


lukeautry:
tslint disable tags look good.

I'm confused about the any casts; the name value should always be a string (it's an optional parameter, defaulting to ''), and the functions we're calling into from that switch require strings. strictNullChecks is on for the project, so if there were a problem there it should have surfaced during the build.

What were you seeing?

egandro:
Yes: "strictNullChecks": true

lukeautry:
Right - [as you can see](https://github.com/lukeautry/tsoa/blob/master/tsconfig.json#L11), strictNullChecks is on for this project, so any violations of that rule get caught in the build.

Are you seeing some compiler error requiring the any casts?

egandro:
So yes ;)

I can't use the generated code - when i go with strictNullChecks. 

1) I can remove this from my tsconfig file (which is a bad idea)
2) I can use a custom template for the generator.

Cool with 2). Can you please point me to an example on how I can use my own template generator?

Issue #6, opened 2016-07-25
Support enums


lukeautry:
Acceptance Criteria:
- Enum properties/return types should generate valid swagger enum types
- Routing layer should validate enums

danzel:
For my uses I'd like to declare an enum and have the field be the string values

```ts
enum ResultType {
    Success,
    FailureNoPermission
}
```

Should result in swagger like:
```js
"type": "string",
"enum": [ "Success", "FailureNoPermission" ],
```

Not sure if other people would like the numeric values.

lukeautry:
I think the string representation that you've proposed is ideal. The number
representation is basically meaningless from a human readability standpoint.

lukeautry:
@danzel Just wanted to provide an update here; enum support hasn't been straightforward, at least for the routing side of things. Generating the correct swagger spec is a straightforward, but wiring up the autogenerated routes to 1) convert strings to their enum values and 2) validate enum values takes a bit more work. Hope to have something in the next week or so.

danzel:
Thanks, no hurry the string work around is fine :)

egandro:
You can try using this:

`export type CaseEnum = "Unchanged" | "Capitalized" | "FirstLetterUppercase" | "CamelCase" | "PascalCase" | "Lower" | "Upper";

export const CaseEnum = {
    Unchanged: "Unchanged" as CaseEnum,
    Capitalized: "Capitalized" as CaseEnum,
    FirstLetterUppercase: "FirstLetterUppercase" as CaseEnum,
    CamelCase: "CamelCase" as CaseEnum,
    PascalCase: "PascalCase" as CaseEnum,
    Lower: "Lower" as CaseEnum,
    Upper: "Upper" as CaseEnum
};`

It feels like an enum - but it's actually a class/string

maxcan:
@lukeautry  would it be easier to support string literal types instead of enums?

for example:
```typescript
type Auth = "Facebook" | "Google" 
```

lukeautry:
@maxcan Yeah, I have a branch going to implement that exact thing; the issue you opened #39 is caused by the inability to parse a type alias/string literal type.

Feel free to take a stab at it though, it'll probably be next week sometime before it can be completed.

maxcan:
@lukeautry which branch is it? None of the branch names seemed obvious

lukeautry:
Sorry Max, didn't mean to imply it was a remote branch, just a local branch
that's exploring the idea of parsing string literal types.

[...]

PR #41, opened 2017-01-09
string literal enum support

maxcan:
Hacked up support for string literal enums.

This partially addresses #6

### DIFF
--- a/src/metadataGeneration/resolveType.ts
+++ b/src/metadataGeneration/resolveType.ts
@@ -10,6 +10,7 @@ syntaxKindMap[ts.SyntaxKind.VoidKeyword] = 'void';
 const localReferenceTypeCache: { [typeName: string]: ReferenceType } = {};
 const inProgressTypes: { [typeName: string]: boolean } = {};
 
+type UsableDeclaration = ts.InterfaceDeclaration | ts.ClassDeclaration | ts.TypeAliasDeclaration;
 export function ResolveType(typeNode: ts.TypeNode): Type {
   const primitiveType = syntaxKindMap[typeNode.kind];
   if (primitiveType) {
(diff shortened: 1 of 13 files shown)


src/metadataGeneration/resolveType.ts
```diff
@@ -62,6 +62,13 @@ function generateReferenceType(typeName: string, cacheReferenceType = true): Ref
     name: typeName,
     properties: properties
   };
+  if (modelTypeDeclaration.kind === ts.SyntaxKind.TypeAliasDeclaration) {
+    let innerType = modelTypeDeclaration.type;
```
lukeautry: Can this be a const?

src/metadataGeneration/resolveType.ts
```diff
@@ -62,6 +62,13 @@ function generateReferenceType(typeName: string, cacheReferenceType = true): Ref
...
+    if (innerType.kind === ts.SyntaxKind.UnionType && (innerType as any).types) {
+      let unionTypes = (innerType as any).types;
```
lukeautry: const here too

lukeautry:
@maxcan  Other than the couple of comments I made, this looks good to me. Thanks for digging in.

As far as I know, we could support literal types for numbers, but it would never make sense to supported mixed types based on the swagger specification.

src/metadataGeneration/resolveType.ts
maxcan: done

src/metadataGeneration/resolveType.ts
maxcan: done

maxcan:
does it need a version bump?

lukeautry:
I'll take care of the version bump.

PR #45, opened 2017-01-17
Custom generated spec behavior configuration

AntoineABI:
Allow to alter the way custom spec (from tsao.json by default) is merged to generated swagger spec by a configuration option in SwaggerConfig interface named specMerging. The default value is meant to not break existing projects meaning this update is harmless.
```typescript
 /**
   * Alter how the spec is merged to generated swagger spec.
   * Possible values:
   *  - 'immediate' is overriding top level elements only thus you can not append a new path or alter an existing value without erasing same level elements.
   *  - 'recursive' proceed to a deep merge and will concat every branches or override or create new values if needed. @see https://www.npmjs.com/package/merge
   * The default is set to immediate so it is not breaking previous versions.
   * @default 'immediate'
   */
  specMerging?: 'immediate' | 'recursive';
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "app-root-path": "^1.4.0",
     "express": "^4.14.0",
     "handlebars": "^4.0.6",
+    "merge": "^1.2.0",
     "mkdirp": "^0.5.1",
     "moment": "^2.17.1",
     "tslint": "^4.1.1",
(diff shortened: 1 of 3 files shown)


lukeautry:
Nice @AntoineABI , I really like this approach.

PR #48, opened 2017-01-30
Support dependency injection for controllers

andreasrueedlinger:


lukeautry:
@andreasrueedlinger This looks great, will merge once I give it a test drive.

I don't necessarily want to make it a requirement for this pull request, but is it feasible to write integration tests for this? I assume we'd just need to pull in inversify as a dev dependency.

andreasrueedlinger:
@lukeautry I added integration tests for express server. I think that should suffice? Was not very inventive tough writing good tests

lukeautry:
Looks good to me. Thanks for all the effort, once I verify everything later today I'll package it into a release.

PR #61, opened 2017-02-21
Manage 'enum' as if they were 'number'

gigagg:
I had to manage enum. 
The easiest way to do that is to treat them as if they were 'numbers'.
It's not perfect, but it works for me.


gigagg:
ooups I didn't push the correct branch.
Closing this merge request

PR #64, opened 2017-02-25
find tsoa.json from working directory and authenticationModule relative path

isman-usoh:


lukeautry:
I'm trying to remember why I added this in the first place. I want to say it was because people wanted to be able to use tsoa as a globally installed package; I'm not sure if this would break that or if it's even a use case that should be supported.

Interested in your thoughts there, but in any case, we should remove the package.json/yarn dependency for app-root-path if we're not going to use it.

PR #67, opened 2017-03-01
Manage 'enum' as if they were 'number'

gigagg:
I have 'enum' in some of my interfaces.
I did some modifications to make tsoa think enum are numbers.
It works for me


isman-usoh:
#76

Issue #27, opened 2016-12-14
Support generic type parameter and return type


DuckMingBrother:
Return type:
Current only support return `Promise<XXX>`, when return `Promise<Response<XXX>>` or `Response<XXX>`, cli throw error "No matching model found for referenced type T".

Paramter type:
use generic type in parameter can generate definition file, but swagger show parameter's schema is undefined.

Can you support it? thank you.

lukeautry:
I think this can be supported and it's a great idea. I'll look into it soon.

alexey-pelykh:
@lukeautry one thing I'm looking to address as well, as soon as other PRs are processed

nenadalm:
I see that support for this was merged, but if I try to use `Promise<ResultList<Resource>>` I get an error:
```shell
$ ./node_modules/.bin/tsoa swagger
There was a problem resolving type of 'T'.
There was a problem resolving type of 'ResultListResource'.
Error: No matching model found for referenced type T
```

with `ResultList` being
```typescript
interface ResultList<T> {
    data: T[],
}
```

version: `tsoa@1.1.8`

judenaveenraj:
@nenadalm +1 same here

judenaveenraj:
Just something i came across while fiddling with the above
**The following seems to work:**
interface ResultList<T> {
    data: T,
}

**The following doesn't, gives the above error that @nenadalm mentioned:**
interface ResultList<T> {
    data: T[],
}
**The following doesn't as well**
interface ResultList<T> {
    data: Array<T>,
}

AndyGura:
+1 on this! I can't make something like 
```
export interface IListPage<T> {
    pageIndex: number;
    totalPages: number;
    list: T[];
}
```
And use as result Promise<IListPage<MyCoolItem>>

amozh:
+1, i have the same issue

AndyGura:
@lukeautry Hello, anytning on this? Its been sitting there dormant for a while

fhewitt:
+1, same issue here. This seems a particularly important feature for code maintenance.

ChamNouki:
Hello, still have a kind a similar issue :
```sh
There was a problem resolving type of 'T'.
There was a problem resolving type of 'JsonApiResourceT'.
There was a problem resolving type of 'JsonApiWrapperUserAttributes'.
Generate routes error.
 Error: No matching model found for referenced type T.
[...]

PR #71, opened 2017-03-05
Added genericType handling for request/response objects

Dirrk:
This PR implements this feature request https://github.com/lukeautry/tsoa/issues/27.

It adds the ability to return and take in generic types.

Example:

```javascript
@Route('user')
export class UserController {

    @Get('{id}')
    public async getUser(@Path() id: string): Promise<IResponse<User>> {
        return {
            success: true,
            result: new User(id, 'Foo Bar')
        };
    }
}

export interface IResponse<T> {
    success: boolean;
    result: T;
}

export class User {
    public id: string;
    public name: string;

    constructor(id, name) {
        this.id = id;
        this.name = name;
    }
}
```

### DIFF
--- a/src/metadataGeneration/resolveType.ts
+++ b/src/metadataGeneration/resolveType.ts
@@ -1,5 +1,6 @@
 import * as ts from 'typescript';
 import { MetadataGenerator, Type, ReferenceType, Property } from './metadataGenerator';
+import * as _ from 'lodash';
 
 const syntaxKindMap: { [kind: number]: string } = {};
 syntaxKindMap[ts.SyntaxKind.NumberKeyword] = 'number';
@@ -40,26 +41,35 @@ export function ResolveType(typeNode: ts.TypeNode): Type {
     return ResolveType(typeReference);
   }
 
+  if (typeReference.typeArguments && typeReference.typeArguments.length === 1) {
+    let typeT: ts.TypeNode[] = typeReference.typeArguments as ts.TypeNode[];
+    return generateReferenceType(typeReference.typeName.text, typeT);
+  }
+
   return generateReferenceType(typeReference.typeName.text);
 }
 
-function generateReferenceType(typeName: string, cacheReferenceType = true): ReferenceType {
+function generateReferenceType(typeName: string, genericTypes?: ts.TypeNode[], cacheReferenceType = true): ReferenceType {
   try {
-    const existingType = localReferenceTypeCache[typeName];
+
+    const typeNameWithGenerics = getTypeName(typeName, genericTypes);
+    const existingType = localReferenceTypeCache[typeNameWithGenerics];
     if (existingType) { return existingType; }
 
-    if (inProgressTypes[typeName]) {
-      return createCircularDependencyResolver(typeName);
+    if (inProgressTypes[typeNameWithGenerics]) {
+      return createCircularDependencyResolver(typeNameWithGenerics);
     }
 
-    inProgressTypes[typeName] = true;
+    inProgressTypes[typeNameWithGenerics] = true;
 
     const modelTypeDeclaration = getModelTypeDeclaration(typeName);
-    const properties = getModelTypeProperties(modelTypeDeclaration);
+
+
+    const properties = getModelTypeProperties(modelTypeDeclaration, genericTypes);
 
     const referenceType: ReferenceType = {
       description: getModelDescription(modelTypeDeclaration),
-      name: typeName,
+      name: typeNameWithGenerics,
       properties: properties
     };
     if (modelTypeDeclaration.kind === ts.SyntaxKind.TypeAliasDeclaration) {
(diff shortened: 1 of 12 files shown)


lukeautry:
Thanks for the pull request.

The generated routes here look good. The generated swagger spec looks a little strange - excerpt here:

```
"GenericModel<TestModel>": {
			"description": "",
			"properties": {
				"result": {
					"$ref": "#/definitions/TestModel"
				}
			},
			"required": [
				"result"
			],
			"type": "object"
		},
		"GenericModel<TestModel[]>": {
			"description": "",
			"properties": {
				"result": {
					"items": {
						"$ref": "#/definitions/TestModel"
					},
					"type": "array",
					"description": ""
				}
			},
			"required": [
				"result"
			],
			"type": "object"
		},
```

I would expect that the generated models would be flattened out; the fact that the requests or responses have generic type arguments is purely a back-end concern. Plus, the generic type argument syntax has no meaning within the swagger world, it's a TypeScript thing.

I believe the solution here would be to make use of Swagger's "allOf" composition property. See http://swagger.io/specification/ under "Composition and Inheritance (Polymorphism)".

If you're not interested in the swagger side of things (i.e. you're just using generated routes), I can look into generating the correct swagger spec.

Dirrk:
@lukeautry I was wondering if you had a planned solution in mind for this?  I provide one below but I don't think its better than whats in the PR.  When you have time could you go over it and let me know?

## Controller
```javascript
// UserController.ts
@Route('user')
export class UserController {

    @Get('{id}')
    public async getUser(@Path() id: string): Promise<IResponse<User>> {
        return {
            success: true,
            result: new User(id, 'Foo Bar')
        };
    }
}

export interface IResponse<T> {
    success: boolean;
    result: T;
}

export class User {
    public id: string;
    public name: string;

    constructor(id, name) {
        this.id = id;
        this.name = name;
    }
}
```

## Swagger Definition
```yaml
// swagger.yml

definitions:
  User:
    properties:
      id:
        type: string
        description: 'user id'
      name:
        type: string
        description: 'user name'
    required:
      - id
      - name
    type: object
  IResponse<User>:
    properties:
      success:
        type: boolean
        description: 'status'
      result:
        $ref: '#/definitions/User'
    required:
      - success
      - result
    type: object
```

## Overview
In the below example (from swagger.io) on how to use allOf.  

### Swagger docs on it (http://swagger.io/specification/#composition-and-inheritance--polymorphism--83)
> Swagger allows combining and extending model definitions using the allOf property of JSON Schema, in effect offering model composition. allOf takes in an array of object definitions that are validated independently but together compose a single object.

```yaml
// Example from
definitions:
  Pet:
    type: object
    discriminator: petType
    properties:
      name:
        type: string
      petType:
        type: string
    required:
    - name
    - petType
  Cat:
    description: A representation of a cat
    allOf:
    - $ref: '#/definitions/Pet'
    - type: object
      properties:
        huntingSkill:
          type: string
          description: The measured skill for hunting
          default: lazy
          enum:
          - clueless
          - lazy
          - adventurous
          - aggressive
      required:
      - huntingSkill
  Dog:
    description: A representation of a dog
    allOf:
    - $ref: '#/definitions/Pet'
    - type: object
      properties:
        packSize:
          type: integer
          format: int32
          description: the size of the pack the dog is from
          default: 0
          minimum: 0
      required:
      - packSize
```
I think the above shows a great example of doing polymorphism and inheritance but the same wouldn't work well with the Users model provided above and that generics give.


My solution using AllOf was to define the object IResponse<User> into different models to accomplish using inheritance

Below is an example of how to first break this into two objects that have all the properties except the result: User object
```yaml
definitions:
  User:
    properties:
      id:
        type: string
        description: 'user id'
      name:
        type: string
        description: 'user name'
    required:
      - id
      - name
    type: object
  IResponse:
    properties:
      success:
        type: boolean
        description: 'status'
    required:
      - success
    type: object
```

A 3rd definition
```yaml
# ...
  IResponse<User>:
    allOf:
      - $ref: '#/definitions/IResponse'
      - type: object
        properties:
          result:
            $ref: '#/definitions/User'
    required:
     - result
```

Personally this seems like its more complex and less readable. It becomes even harder when you add in a subgeneric type or an array.  But most importantly I think having documentation like this makes api documentation easier to read. 

Thanks!

Dirrk:
@lukeautry any update on this?

lukeautry:
@Dirrk Need to take some more time to look at it. Will try to look this week; @isman-usoh if you're interested in this may be worth your time to look as well.

lukeautry:
@Dirrk I finally sat down and started working on this.

What if, for the swagger spec, we normalized the response name, rather than using the generic syntax? e.g. IResponse<User> becomes IResponseUser. I'd be okay with that as a solution here - no need to bring in the complexity of allOf.

Dirrk:
@lukeautry that sounds good to me, I will remove the brackets and push that up. Thanks for taking the time to review it!

Dirrk:
@lukeautry Fixed the generic type string and resolved the merge conflicts.  Should be ready when you get a chance.  Thanks

lukeautry:
More merge conflicts, I'm taking care of them.

Issue #70, opened 2017-03-04
RoutesConfig error


Dirrk:
When providing iocModule in the RoutesConfig it checks to see if the file exists at the path inside the string name but the docs say:

> Note that the path is relative to the routesDir and does not contain the extension.

I can fix this, but I am not sure if you want to fix it via the docs or code.  

Also as a heads up this bug: https://github.com/vvakame/typescript-formatter/issues/87 is causing a build failure for your code.  Potential solutions would be to lock down to version (4.1.2) until its fixed or pass in vscode param.

lukeautry:
@Dirrk I think we should just fix it in the docs, some of the work @isman-usoh did on working path probably caused this.

Regarding typescript-formatter, I think this isn't an issue if you use "yarn" to install, which I have been. That said there's nothing in the project docs that would indicate to use yarn for dev environment.

Dirrk:
What about yarn is not causing the issue?  Wouldn't a new user, installing with yarn to this repo have problems? Maybe I am just missing something, and its possible because I don't use yarn.  I will put up a pr to fix the docs if thats cool

lukeautry:
Well, I only say that because I'm using yarn to install deps locally and haven't seen the problem; yarn generates a lock file, which prevents weird npm behavior of auto-updated to minor/patch versions with breaking changes.

That said I have no problem pinning the version, will look at your change.

PR #72, opened 2017-03-05
ISSUE-70

Dirrk:
See https://github.com/lukeautry/tsoa/issues/70

### DIFF
--- a/src/routeGeneration/routeGenerator.ts
+++ b/src/routeGeneration/routeGenerator.ts
@@ -22,7 +22,8 @@ export class RouteGenerator {
         tsconfig: true,
         tsfmt: true,
         tslint: true,
-        verify: true
+        verify: true,
+        vscode: true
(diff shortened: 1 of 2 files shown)


src/routeGeneration/routeGenerator.ts
```diff
@@ -22,7 +22,8 @@ export class RouteGenerator {
         tsconfig: true,
         tsfmt: true,
         tslint: true,
-        verify: true
+        verify: true,
+        vscode: true
```
lukeautry: What does vscode: true do?

Dirrk: This was related to the issue I was talking about in typescript-formatter.  They just expect vscode to be defined.  The default is true, which allows it to use .vscode/settings.json https://github.com/vvakame/typescript-formatter/issues/87

PR #78, opened 2017-03-10
Allow non-production builds

alexey-pelykh:
To allow
```
npm install -S ../tsoa
```


.npmignore
```diff
@@ -1,2 +1,2 @@
-*
 !dist/**
+typings
```
lukeautry: I don't think this should be necessary; typings directory shouldn't be there anyway.

alexey-pelykh:
@lukeautry updated

lukeautry:
@alexey-pelykh Looks like a little more got merged into this branch than you intended.

alexey-pelykh:
@lukeautry indeed, sorry. Updated

PR #81, opened 2017-03-16
(feat) refactored to allow route templates to be overridden

cadriel:
First stab at refactoring the way route templates are handled.

Note: I'm very tempted to move the hardcoded handlebar content in `routeGenerator.ts` on around `line: 50` to the templates too, but wanted your thoughts first @lukeautry.

Otherwise - all tests are passing and this seems to be working ok.


cadriel:
Encountered a few pathing issues - sorting them out.

src/routeGeneration/routeGenerator.ts
```diff
@@ -38,57 +35,68 @@ export class RouteGenerator {
...
+    const iocModule = this.options.iocModule ? this.getRelativeImportPath(this.options.iocModule) : undefined;
+
+    // If we're working locally then tsoa won't exist as an importable module.
+    // So, we reference its helpers and base controller via path instead of alias.
+    // Note: require.resolve('tsoa') works under normal circumstances, but when using npm link
+    // to develop locally, it breaks. Keep this in mind.
     let canImportByAlias: boolean;
     try {
-      require('tsoa');
+      require.resolve('tsoa');
```
cadriel: This feels problematic if you're trying to test with a project that is linked locally to tsoa. I tested, and if using link - require always throws an error.

If this is only there to support the tests running - could we perhaps use NODE_ENV instead to set the environment to test - and change the imports that way instead?

cadriel:
Other than my comment re: require, I think this is good to go.

cadriel:
Final update;

- Moved all template code to the template files, no more concat.
- Added handlebars-helpers. For those that want to have basic comparison operators in their templates.
- Changed to the code above whereby we check environment rather than availability of a module.

lukeautry:
@cadriel Thanks for doing this work! We've needed something like this for a while.

One issue with the PR right now is the use of `NODE_ENV=test` in Windows. I think maybe we can use something like https://github.com/kentcdodds/cross-env to get the same effect in a cross-platform way.

cadriel:
Sure, np. :)

I'll push in an update that makes use of cross-env momentarily.

I also have some more work that reduces the amount of code produced in the routes.ts - and instead defer's it to two middleware functions. One handles the parameter checking in the same way (but passes the validated args in the request) - the other loads the controller by way of a factory service.

Not quite ready yet, but if you were interested in those changes I could push them in too when ready.

cadriel:
Updated to use cross-env.

lukeautry:
Thanks @cadriel . Sure, I'll wait to merge until those changes come up.

cadriel:
Looks like there's still some issues building on Windows. I'll look into these and let you know once I have it running there.

cadriel:
Ok @lukeautry;

This is now building and testing fine in both MacOS and Windows 10 for me.

cadriel:
You can probably merge this one now if you're happy with it.

The other work I mentioned above I'll create another PR for in the future.

cserron:
Hi,
When do you think this will be ready to be merged in?
We are actually waiting for something like this to be implemented in our TSOA project.
Thanks.

lukeautry:
Should be able to merge it in this weekend.

lukeautry:
Merged, released in 1.1.7. Thanks @cadriel!

cadriel:
np :)

PR #86, opened 2017-03-28
FEATURE: CLI support override host and basePath

isman-usoh:


lukeautry:
Why not just override this in the config file? What's the use case?

isman-usoh:
case 1. use subdoamin defined api version e.g. http://localhost/v1 for dev, http://v1.domain.com/ for production
case 2. code deploy multi tier e.g. http://dev.domain.com/v1, http://staging.domain.com/v1
case 3. if run on cloud every create new instances, host ip has changed.

PR #100, opened 2017-05-16
FEATURE: Support for modifying the internal compiler options

### LINKED ISSUE #99: The internal metadataGenerator doesn't allow compiler options to be set.
Currently, the internal TypeScript program used during the route generation process is instantiated with an empty CompilerOptions object.  When attempting to generate routes in a project that utilizes path based module resolution (example follows) the generator is unable to walk dependencies to generate models.

tsconfig.json:
```json
{
"compilerOptions": {
     "baseUrl": "./path/to/base/url",
     "paths": {
          "exampleLib": "./path/to/example/lib"
     }
}
```

The result is that during type resolution no model is available and an error occurs:
> No matching model found for referenced type...

I intend to submit a pull request that utilizes the RoutesConfig interface to pass in TypeScript compiler options.  I'm not sure of the consequences of passing options in besides the one necessary to make path resolution work, so I haven't documented them in the README.

shaunlmason:
Fixes #99

### DIFF
--- a/README.MD
+++ b/README.MD
@@ -318,6 +318,42 @@ export class SecureController {
 }
 ```
 
+### Path Mapping
+
+Per the [TypeScript Handbook](https://www.typescriptlang.org/docs/handbook/basic-types.html) under [module resolution](https://www.typescriptlang.org/docs/handbook/module-resolution.html):
+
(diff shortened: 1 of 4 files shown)


shaunlmason:
My PR doesn't fix the same issue occurring during swagger generation.  I'll be back.

Issue #108, opened 2017-06-15
Duplicate required properties in swagger in definitions.


luminblade:
When a method returns a class defined with a property that is defined both in the class and the constructor, the property is duplicated in the required section of the generated swagger.json, making it invalid.

One solution is to only define the property in the constructor, but this can reduce readability, especially on large classes.

This code produces duplicate required properties for "prop1".

```
export class ExampleModel {
	public prop1: string;
	public prop2?: string;

	constructor(prop1: string) {
		this.prop1 = prop1;
	}
}

@Route('example')
export class ExampleController extends Controller {

	@Get('duplicate')
	public async duplicate(@Query() summary?: string): Promise<ExampleModel> {
		let ex = new ExampleModel('value');
		return Promise.resolve(ex);
	}
}
````

```
"ExampleModel": {
	"description": "",
		"properties": {
		"prop1": {
			"type": "string",
				"description": ""
		},
		"prop2": {
			"type": "string",
				"description": ""
		}
	},
	"required": [
		"prop1",
		"prop1"
	],
		"type": "object"
}
````

shaunlmason:
@luminblade This has been a consistent nagging issue in my use as well.  I *think* I have it fixed, just waiting on a review.

Thanks for the feedback.

PR #109, opened 2017-06-15
fix(duplicate-required-props): Resolve issue #108.

shaunlmason:
As described in issue #108, constructor properties intended for initialization during instantiation would result in duplicate property metadata being created when naming had a collision.  

This commit adds a initially failing test to catch the issue, and a dedupe method called during type resolution.


shaunlmason:
@lukeautry I can merge this, but I'd appreciate a second set of eyes to ensure I didn't introduce any issues due to my relative lack of familiarity.

isman-usoh:
i fix in https://github.com/lukeautry/tsoa/commit/c1cd34f824236232268322e52de3a587c920d73a#diff-d0ebfd6b9cb2fdea327d4f71b33a282eR67

shaunlmason:
Underlying issues is fixed, per @isman-usoh.

PR #116, opened 2017-06-28
fix for typescript 2.4.x

egandro:
Some data structures in TS 2.4 have been changed.

This results in a compiler error.


lukeautry:
It would probably be worth upgrading the TypeScript dependency in the project to 2.4 while we're at it.

egandro:
Maybe we can do more brain storming to this - here my thoughts so far.

Tsoa has some outstanding architecture when it comes to dependencies.

1. There is a "runtime" package for the attributes e.g. Route, Path, etc.
2. There is a "development" package at development time for creating the routes

If the runtime has zero use when the routes are builded. I just decorates the controllers and isn't required at all.

Maybe we could think of a nice concept how to separate this - as the main project can have no "runtime" depencencies to the main tsoa libraries.

Why this? Because if we force a 2.4 compiler to tsoa, this will impact the whole toolchain of the user.

NPM can be sometimes very wierd in terms on deps and devdeps - so let's carefully think about this.

I hate it if somebody breaks my build with a new NPM package ;) I'd like to avoid being "THAT" guy now by myself ;)))

egandro:
I hate it when I am right :(

Some other stupid NPM package forced me to upgrade to 2.4.1 - it took me 4 hours to fix some strang "bugs" with typings that - well are bugs - but I really want to deal with later..

PR #127, opened 2017-07-02
Changing authenticateMiddleware to return the auth function.

kingjerod:
I'm not sure why I didn't have this issue before when I was testing out security, but when using the `@Security` decorator, it is causing a 404 on that route. Through a lot of trial and error I narrowed it down to missing this return statement. I've tested it in Koa but am not familiar enough with Express/Hapi to test it there. 

Maybe this is some fluke on my end and this is not needed? I cannot get `@Security` to work otherwise.


lukeautry:
Thanks @kingjerod! Will release this in the next update.

PR #129, opened 2017-07-07
DecoratorPlugin

chaowlert:
Hi @lukeautry,

Your library is great, and it would be nicer if we can utilize from existing decorators.
Currently, I'm using [routing-controllers](https://github.com/pleerock/routing-controllers).
Therefore, I have idea to make this library extensible by plugin.

This pull request is to transform current codebase to accept plugin. And I already create plugin at 
https://github.com/chaowlert/tsoa-routing-controllers.

#### Basic usage

**install**
```
npm i -g tsoa
npm i -g tsoa-routing-controllers
```

**usage**
```
tsoa swagger --decoratorPlugin routing-controllers
```

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -2,6 +2,26 @@
     "version": "0.2.0",
     "configurations": [
         {
+            "type": "node",
+            "request": "launch",
+            "name": "Mocha Tests",
+            "program": "${workspaceRoot}/node_modules/mocha/bin/_mocha",
+            "args": [
+                "-u",
+                "tdd",
+                "--timeout",
+                "999999",
+                "--colors",
+                "${workspaceRoot}/tests/**/*.spec.ts",
+                "--compilers",
+                "ts:ts-node/register"
+            ],
+            "env": {
+                "NODE_ENV": "test"
+            },
+            "internalConsoleOptions": "openOnSessionStart"
+        },
+        {
             "name": "Generate",
             "type": "node",
             "request": "launch",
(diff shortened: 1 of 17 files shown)


joshuamking:
Is this going anywhere? Because this would be awesome to use!

BenjD90:
Very nice idea !

endor:
Do I understand correctly that the idea here is to use `routing-controllers`, but then be able to use `tsoa` alongside it to generate the swagger spec?

I think this PR is a good idea, but I guess there's some work to be done before it can be merged:
* First and foremost the merge conflicts would need to be solved. 
* I think it would be great to have some unit tests with a different decorator plugin than the default, for example the `routing-controller` one that @chaowlert has written. 
* The existing files use two spaces indentation, the new ones four.
* I think using `getPhrase` (typo) is a nice idea, but would be better in a separate PR, so there can be a separate decision on whether to merge it.

lukeautry:
There seems to be some popularity behind this idea.  The merge conflicts are also so old that this PR would require serious rework to make it in. My apologies for not addressing it sooner - I realize this response is coming at a comically late time.

I'm going to close this out, but if someone would like to take this effort back over, I think it would be reasonable to support something like this.

WoH:
I'm not sure if this is the right place, but it seems like a good spot to ask before creating an issue.
The [project referenced](https://github.com/typestack/routing-controllers) seems to do some of the things tsoa also ships with and probably some additional stuff (I'm unfortunately not familiar with it).
Is there maybe an interest to explore if it makes sense to use the re-export routing-controllers' decorators within tsoa?

I've seen some issues that may potentially be impacted here, for example:
- #353 
- #351 
- #342 
- #365

lukeautry:
Yeah, I think that it would be interesting to do an integration with routing-controllers specifically. It has a much broader design goal, so if you could just plug-and-play with tsoa, that would be nice. I'd almost be more inclined to just add a "mode" where the decorator semantics change over to using routing-controller semantics, but this would be a brittle approach.

dgreene1:
Yea, I'm mixed on the idea of bringing in Routing Controllers library. At a minimal glance, the semantics are different: `@Put("/users/:id")` vs our `@Put("/users/{id}")`. But I'm interested in whatever we can do to make more people want to use tsoa. So I like where this conversation is going. :)

Issue #130, opened 2017-07-07
Controllers have state


nenadalm:
If I want to set response code, I have to extend tsoa controller which have method `setStatus`. This requires controller to be instantiated on each request and leads to bugs if somebody forgets to do that (e.g. by specifying wrong scope in dependency injection container.

Wouldn't it be better if status and possibly other values were returned by the action so that controllers could be stateless? You could get inspiration from ring-clojure https://github.com/ring-clojure/ring/wiki/Concepts#responses.

So the response could look like:
```typescript
interface Response<T> {
    status?: Number;
    body: T;
    headers?: {[key: string]: string};
}
```
?

egandro:
You can have this with Dependency Injection / IoC.

In the Inversify example we have:

```
let provideSingleton = function(
  identifier: string | symbol | interfaces.Newable<any> | interfaces.Abstract<any>
) {
    return fluentProvider(identifier)
      .inSingletonScope()
      .done();
};
```

Add this:

```
import { Controller } from 'tsoa';
import { decorate, injectable } from 'inversify';
decorate(injectable(), Controller);

let provideTransient = function (
  identifier: string | symbol | interfaces.Newable<any> | interfaces.Abstract<any>
) {
  return fluentProvider(identifier)
    .inTransientScope()
    .done();
};

```

And decorate your controller this way:

```
@Route('foo')
@provideTransient (FooController)
export class FooController extends Controller {
  ...
}
```

Every Controller get's created per call. Please note - because of current limitations of tsoa you have to decorate all controllers.

nenadalm:
@egandro sure, but my proposal is to make controllers stateless.

Currently it works something like:
```typescript
class Controller extends TsoaController {
    public action() {
        this.setStatus(200); // storing status in magical place

        return {
            name: 'John Doe',
            count: 5,
        };
    }
}

const controller = new Controller();
const body = controller->action();
const status = controller.getStatus(); // retrieving status from magical place
response.status(status).send(JSON.stringify(body));
```

I propose to make it work like:
```typescript
class Controller extends TsoaController {
    public action() {
        return {
            status: 200,
            body: {
                name: 'John Doe',
                count: 5,
            },
        };
    }
}

const controller = new Controller();
const response = controller->action();
response.status(response.status).send(JSON.stringify(response.body));
```

Advantage is that action returns required data for response (instead of retrieving part of them and storing other part in some magical place). So as for me this is more clear + it's less error prone (see my previous  post).

[...]

PR #132, opened 2017-07-10
Avoid stateful controllers

nenadalm:
fixes https://github.com/lukeautry/tsoa/issues/130


isman-usoh:
Fix: 0534cbdbe2931efb58b0ee1d1780632bb3869b76

Example: https://github.com/lukeautry/tsoa/blob/0534cbdbe2931efb58b0ee1d1780632bb3869b76/tests/fixtures/controllers/testController.ts#L21

nenadalm:
@isman-usoh the controller you sent as example extends the Controller from tsoa. What I wanted is controller without base class without state (as title suggests). So I don't know how is it fix. But since I won't use this library for future projects it doesn't matter.

Reason I won't use it is simply that there is limit to what you can generate from types (e.g. you can't figure out that some property has to be `> 0`). So since I validate everything with joi (https://github.com/hapijs/joi) anyway, I'll investigate how to write apps so that I can use that to generate swagger (https://www.npmjs.com/package/hapi-swagger looks promising).

Issue #94, opened 2017-05-05
setting status codes is not working within Promise


ryo-nbx:
When defining error and success codes for API functions I realized a problem when using Promises. When setting a status code within a promise this code is not returned by the function. 

For example this controller function:
`@Tags('Car')`
`@SuccessResponse('200', 'successful')`
`    @Response('404', 'not found')`
`    @Response('407', 'internal error, could not process request')`
`    @Get('{id}')`
`    public async getCarById (@Path('id') id: string ) : Promise<Car> {`
`        return service.getCar(id).then( (car): any => {`
`            if (!car) {`
`                // not found`
`                this.setStatus(404);`
`            }`
`            return car;`
`        }).catch(function (error) {`
`            console.log(error);`
`            // internal error`
`            this.setStatus(407);`
`            return null;`
`        });`
`    }`

This function calls a method at a database service and gets a Promise. After evaluating the result of this promise a response code might be set. Calling this function with a missing ID should cause a 404 error. But in reality it returns 204. The console prompts a 404 object but this is not handled by the generated routes.

I tracked down the problem to the generated routes.ts file and its promiseHandler function. 
`function promiseHandler(promise: any, statusCode: any, response: any, next: any) {`

The method takes a promise and status codes from the controller. This tells me that custom status codes are only used when they are set before the promiseHandler function call. If you set a code within a controller function (and not within a promised function) it will be used. If you use a Promise and set a code within it, this setStatus() call is executed after the promiseHandler() call and so the status code will be ignored.

I looked for a solution and there seems to be a simple one but it requires to change TSOA template code.

I think this should work: 

tsoa/src/routeGeneration/templates/express.ts

`...`
`export function RegisterRoutes(app: any) {`
`    {{#each controllers}}`
`    {{#each actions}}`
`...`
`const promise = controller.{{name}}.apply(controller, validatedArgs);`
`let tController = undefined;`

`            if (controller instanceof Controller) {`
`                tController = (controller as Controller);`
`            }`
`            promiseHandler(promise, tController, response, next);`
`...`
`    function promiseHandler(promise: any, controller: Controller, response: any, next: any) {`
`        return promise`
`            .then((data: any) => {`
`                let statusCode = undefined;`
`if (controller) {`
`statusCode = controller.getStatus();`
`}`
`                if (data) {`
`                    response.json(data);`
`                    response.status(statusCode || 200);`
`                } else {`
`                    response.status(statusCode || 204);`
`                    response.end();`
`                }`
[...]

PR #136, opened 2017-07-14
BUGFIX: Issue #94 : getting the status into the promise then callback

PendingChanges:
As the controllers functions are all async you never get the status set into the function.
To fix that, you can get the status code into the then callback.
You cannot put "await" on the function call because of the apply.


PendingChanges:
As nenadalm said, no need to keep this pull request

PR #140, opened 2017-07-19
support implicit type

chaowlert:
This change will support

1. implicit parameter type
2. implicit return type


lukeautry:
If this really works, you're my hero. Will take a look at this when I do the roundup for the next release.

Issue #144, opened 2017-08-02
Provide proper exit code when error occurs


dvishniakov:
Currently functions _swaggerSpecGenerator_ and _routeGenerator_ from cli.js eat error codes. There is a console output but it's not very useful without workarounds and output parsing.
This bug prevents usage of this package in different pipelines, like "lint && tsc && tsoa swagger && release"
The simplest and fastest solution is to add **process.exit(1);** after console messages

isman-usoh:
Fix: 65a2cf9650b2a0cc24d476dce95505996e00c63e

PR #145, opened 2017-08-02
Return proper exit code on error

dvishniakov:
https://github.com/lukeautry/tsoa/issues/144


isman-usoh:
@dvishniakov  Thank you.

PR #152, opened 2017-08-20
@Hidden decorator to entirely hide an method from documentation

strongtigerman:
Allows user to exclude some private methods from swagger documenations by using  @Hidden decorator


lukeautry:
Great work @strongtigerman.

lukeautry:
@strongtigerman This is released under 1.3.4.

PR #153, opened 2017-08-20
fix @ Response<T> doesn't generate example

strongtigerman:
fix @Response<T> doesn't generate example


krueglt:
Does this also fix the SuccessResponse? If SuccessResponse is explicitly given, the @Example seems to get ignored.

PR #154, opened 2017-08-28
Feature: return file as response

vkagamlyk:
This feature consists of 2 parts:
1. @Produces decorator to set content type of response. Supported only 1 value.
Usage example: @Produces('text/html; charset=utf-8')
2. Return file as response.
Usage example:
const result = new FileResult();
result.data = 'Readable stream or any type supported by response.send()';


egandro:
Oh again here. This is also on my Todo List.

@vkagamlyk is it ok for you to take your code and fix it for the current release?

dgreene1:
I think @vkagamlyk is no longer interested in this and someone else picked it up in https://github.com/lukeautry/tsoa/pull/228 so we’ll track it there

PR #156, opened 2017-08-31
Fix for typo in route templates

eug48:
This recently introduced typo was detected by TypeScript's compiler in my project (strictNullChecks on). It looks like it would cause response status codes to be incorrect.

Thanks for this project!


lukeautry:
Ha, good find. Thanks!

Issue #155, opened 2017-08-30
Uncertainty in route paths for express template (and possibly for others)


amozh:
[This](https://github.com/lukeautry/tsoa/blob/master/src/routeGeneration/templates/express.ts#L40) line of code 
```app.{{method}}('{{../../basePath}}/{{../path}}{{path}}'``` 
in _src/routeGeneration/templates/express.ts_ template adds hardcoded `/` between base path and controller path.
Additionally [this](https://github.com/lukeautry/tsoa/blob/master/src/metadataGeneration/methodGenerator.ts#L102) line of code
```this.path = decoratorArgument ? `/${decoratorArgument.text}` : '';```
in _src/metadataGeneration/methodGenerator.ts_ adds hardcoded `/` as prefix to every action path.

So if i have base path `/api/v2` and 
```
@Route()
export class LoginController {
    @Post('login')
    public async login(...): Promise<...> { ... }
}
```
generated route will be
```
app.post('/api/v2//login',
        function(request: any, response: any, next: any) {
            ...
        });
```
As you can see there is a double `/` in the path because of hardcoded symbols mentioned above. 
Note that Swagger doc gets generated fine though.
This makes `/api/v2//login` endpoint working instead of `/api/v2/login`, call to latter returns 404.
Some servers are ignoring that and probably Express can ignore as well, but it would be great if TSOA can generate normal routes which are fully working on any server.

Here is the possible way to solve the problem:
1) Remove hardcoded `/` in code mentioned above.
2) Normalise all route paths to avoid redundant or missing `/` signs, possible way to do this shown below in the modified part of code from _src/routeGeneration/routeGenerator.ts_
```
/**
     * Removes all '/', '\', and spaces at the beginning and at the end of the path
     * Adds prefix and suffix if supplied
     */
    function normalisePath(pathStr: string, withPrefix?: string, withSuffix?: string): string {
      if(!pathStr || typeof pathStr !== 'string') {
        return pathStr;
      }
      let normalised = pathStr.replace(/^[/\\ ]+|[/\\ ]+$/g, '');
      normalised = withPrefix ? withPrefix + normalised : normalised;
      normalised = withSuffix ? normalised + withSuffix : normalised;
      return normalised
    }

    return routesTemplate({
      authenticationModule,
      basePath: this.options.basePath === '/' ? '' : normalisePath(this.options.basePath as string, '/'),
      canImportByAlias,
      controllers: this.metadata.controllers.map(controller => {
        return {
[...]

PR #161, opened 2017-09-05
Base, Controller and Method paths normalisation

amozh:
Fix for bug described in this issue https://github.com/lukeautry/tsoa/issues/155

### DIFF
--- a/src/metadataGeneration/controllerGenerator.ts
+++ b/src/metadataGeneration/controllerGenerator.ts
@@ -1,4 +1,5 @@
 import * as ts from 'typescript';
+import { normalisePath } from '../utils/pathUtils';
 import { getDecorators } from './../utils/decoratorUtils';
 import { GenerateMetadataError } from './exceptions';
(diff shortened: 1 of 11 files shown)


ajmath:
I think this PR broke folks who are using the root path.

For example, if I have left basePath off or set it to `/`, it will generate paths in my routes file like: `path: '//auth/okta/validate',` which will cause the http server (hapi in my case) to fail to launch.

The major difference in my case with this change is that basePath changes from `''` to `'/'` which causes the initial double slash.

After looking at the code, I'm not sure where there are so many parameters to the new function.  The second two are never used outside of the tests.

amozh:
@ajmath Yeah, my bad. Missed this check:
src/cli.ts:83
```
config.basePath = config.basePath || '/';
```
Will fix it asap.

amozh:
@ajmath here is the PR with fix
https://github.com/lukeautry/tsoa/pull/173

ajmath:
Thanks.

PR #168, opened 2017-09-21
Minor bugfixes

connected-jlovelock:
Two separate issues:

1. Original decorator parsing assumed all decorators were supplied by tsoa, which made parsing fail if you were using any other decorators (e.g. from other libs). Added a sanity check here to guard against trying to access fields on a null/undefined object.
1. Interface definitions for derived types/classes that overrode their parents' signatures were displayed incorrectly: the code would override the specific values with their more general inherited counterparts, rather than the other way around. This is now fixed.

### DIFF
--- a/src/metadataGeneration/controllerGenerator.ts
+++ b/src/metadataGeneration/controllerGenerator.ts
@@ -47,7 +47,7 @@ export class ControllerGenerator {
       .map(d => d.expression as ts.CallExpression)
       .filter(expression => {
         const subExpression = expression.expression as ts.Identifier;
-        return subExpression.text === decoratorName;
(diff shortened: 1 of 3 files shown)


src/metadataGeneration/controllerGenerator.ts
```diff
@@ -47,7 +47,7 @@ export class ControllerGenerator {
       .map(d => d.expression as ts.CallExpression)
       .filter(expression => {
         const subExpression = expression.expression as ts.Identifier;
-        return subExpression.text === decoratorName;
+        return (subExpression && subExpression.text) === decoratorName; // <-- fix
```
lukeautry: remove comment

PR #173, opened 2017-10-26
Bugfix for server error if base path is empty or `/`

amozh:
Fix for the issue caused by [this](https://github.com/lukeautry/tsoa/pull/161) PR described in  [comment](https://github.com/lukeautry/tsoa/pull/161#issuecomment-338747127) by @ajmath.

The problem was in trying to normalise endpoint path by peaces (base, controller and method).

Looks like a lot of changed files but most of them are _auto-generated routes_ and _tests_.

Actual changes:
1) Moved all paths normalisation from metadata generators to route and spec generators.
https://github.com/lukeautry/tsoa/pull/173/files#diff-d107cbd1a26a9e3dbdf1db48ecdeac37R76
https://github.com/lukeautry/tsoa/pull/173/files#diff-d0ebfd6b9cb2fdea327d4f71b33a282eR81

2) Instead of normalising paths only by peaces (base, controller and method), now additionally normalising them after normalised peaces are concatenated to a single string.
https://github.com/lukeautry/tsoa/pull/173/files#diff-d107cbd1a26a9e3dbdf1db48ecdeac37R93

3) Instead of rendering path from peaces in template (`{{../../basePath}}{{../path}}{{path}}`) now  using the new `fullPath` property of `action` object, which contains normalised full endpoint path, so in template it looks like (`{{fullPath}}`)
https://github.com/lukeautry/tsoa/pull/173/files#diff-d107cbd1a26a9e3dbdf1db48ecdeac37R98
https://github.com/lukeautry/tsoa/pull/173/files#diff-ec8b224baa9060a807593128914f5adfR40

4) Added controller with empty path and tests for it.
https://github.com/lukeautry/tsoa/pull/173/files#diff-15c48bf30d8d99500ba77a5707ee4643R8

Additionally i can add tests for empty base path, but not sure whether it is needed.

### DIFF
--- a/src/metadataGeneration/controllerGenerator.ts
+++ b/src/metadataGeneration/controllerGenerator.ts
@@ -1,6 +1,5 @@
 import * as ts from 'typescript';
 import { getDecorators } from './../utils/decoratorUtils';
-import { normalisePath } from './../utils/pathUtils';
 import { GenerateMetadataError } from './exceptions';
 import { MethodGenerator } from './methodGenerator';
 import { Tsoa } from './tsoa';
@@ -34,7 +33,7 @@ export class ControllerGenerator {
       location: sourceFile.fileName,
       methods: this.buildMethods(),
       name: this.node.name.text,
-      path: normalisePath(this.path as string, '/'),
+      path: this.path || '',
     };
   }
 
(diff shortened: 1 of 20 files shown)


algumugla:
Do we know when this bug fix will be released ?
Thank you

brsmith9000:
Any update on if this will be merged?

KyleGalvin:
Until this has been merged, I can't use the generated routes.ts in my project.

The class:
@Route("auth")
export class AuthController {
...
  @Post("signup")
  public async signup...

Gets generated to:

export function RegisterRoutes(app: any) {
    app.post('//auth/signup',

The double slash at the start of the route is causing my client to 404 all requests.

Manually removing the slashes in the generated code before runtime solves the issue.

lukeautry:
Merged, published in 2.1.1

Issue #187, opened 2018-01-21
Extensibility


taicho:
@lukeautry First and foremost I want to say I love the idea behind this framework / library. That being said, it would be nice to see some form of extensibility added beyond the templates. I'm really torn - I want to use this library on a substantial project but I'm concerned that time will be consumed trying to accommodate some inadequacies present.

Obviously, the simple answer could be "just fork it" and make the changes necessary...although looking at the stack of PRs it's obvious that you aren't able to devote as much time to this project as much as you'd like (which I totally understand).

Nevertheless, I have a couple humble suggestions as to what would make this library more viable and help gain adoption in addition to reducing maintenance:

- Add support for an @Custom decorator e.g. @Custom(name: String, value : any);
- Add support for the ability to specify a handler in tsoa.json associated with a name provided in an @Custom attribute that points to js/ts file containing the inputs/output required by tsoa for the swagger-gen and swagger-routes stage...

Anyways...many ways obviously to slice this cake and this may not be the best one but the point is that it would go along way towards making this library a staple in Node-TypeScript-Swagger development and projects if the emphasis was shifted towards extensibility. If this doesn't occur I may attempt to do the same but it seems like a really great opportunity to push your library further if that's your desire...

Cheers

karpoff:
agree about custom decorators
for example I'm using authorization middleware and I want to pass user object to controller handle
now I can use
```
    @Get('some')
    public async getSome(@Request() request: express.Request): Promise<any> {
        return myHandler(request.user);
    }
```

but I want to use
```
    @Get('some')
    public async getSome(@User() user: IMyUserInterface): Promise<any> {
        return myHandler(user);
    }
```
and then change route template to pass request.user into controller method
it is more user-friendly

lukeautry:
@taicho Thanks for the note. As you've pointed out, there's quite a stack of issues and pull requests around this library. However, it is not abandonware.

The good news is that I now personally depend on tsoa in production and therefore have more skin in the game to improve this library than ever. I will have more time to devote towards improving this library in the near future, and I think there are several steps we can take with regard to extensibility that makes this lib a true "staple".

Thavaprakash:
Hi @lukeautry, Can you integrate Webpack(ts-loader) instead of `tsc` in this project as webpack has many features.

lukeautry:
I doubt we'd want to use Webpack on this since it's a purely server-side library. What kind of webpack features would you be likely to use?

dgreene1:
[...]

PR #193, opened 2018-02-11
Add custom parameters defined by a provided function on how to get it…

oakesja:
… from the request

This addresses https://github.com/lukeautry/tsoa/issues/187 and extensibility without forcing people to change how routes are generated. 

The `@CustomParameter` parameter decorator takes in a function that defines how to get your custom parameter from the request. The `@CustomParameters` method decorator is placed on a controller method has use custom parameters. It is what actually manipulates the arguments by doing the mapping of the request to the given custom parameter.

Let me know if there are any additional tests or changes to the readme needed.

### DIFF
--- a/src/decorators/custom.ts
+++ b/src/decorators/custom.ts
@@ -1,0 +1,32 @@
+const METADATA_KEY = `tsoa_custom_parameter`;
+
+export function CustomParameter(getParam: (req: any) => any) {
+  return (target: object, key: string | symbol, index: number) =>  {
+    if (!target[METADATA_KEY]) {
+      target[METADATA_KEY] = {};
+    }
+
+    const details = {
+      getParam,
+      index,
+    };
+    if (target[METADATA_KEY][key]) {
+      target[METADATA_KEY][key].push(details);
+    } else {
+      target[METADATA_KEY][key] = [details];
+    }
+  };
+}
+
+export function CustomParameters() {
+  return (target: object, key: string | symbol, descriptor: any) =>  {
+    const originalMethod = descriptor.value;
+    descriptor.value = function(...args: any[]) {
+      target[METADATA_KEY][key].forEach(({ getParam, index }) => {
+        args[index] = getParam(args[index]);
+      });
+      return originalMethod.apply(this, args);
+    };
+    return descriptor;
+  };
+}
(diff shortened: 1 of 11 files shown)


AmazingTurtle:
Looks nice to me. Can you please rebase your change, @oakesja ?

oakesja:
@AmazingTurtle just rebased

AmazingTurtle:
@lukeautry would be glad if you had a Look on this. Otherwise you can grant me some permissions as contributor, I‘d appreciate to review and test all those PR‘s.

KyleGalvin:
Looking forward to being able to define endpoint tag arrays in my swagger.json

dgreene1:
@oakesja this seems like a really great idea. I have merge powers, so I can merge this right after https://github.com/lukeautry/tsoa/pull/385 closes since that has a lot of changes in it. But I have some questions. 

1. We should add a section to the readme that explains how to use this
2. I didn't see any code in here that updates the swagger yaml or that verifies the output. I'd like to see that in there before we merge this
3. I think that `@CustomParameter` needs to take a type parameter because otherwise tsoa won't know what `$ref` to assign the parameter when generating the documents. So `CustomParameter(getParam: (req: any) => any)` should be `CustomParameter<TInput, TOutput>(getParam: (req: TInput) => TOutput)`. TInput would be the type that's used to create the swagger `$ref`.
  * as a side effect of this, we'll have to verify that the typescript type for the route parameter is actually correct

So for instance, if add this decorator, we have to prevent this circumstance because the swagger model would tell the user that they need to send in a number only to have it parsed again
```ts
    @Get('SomeResource')
    @CustomParameters()
    public async customParameter( @CustomParameter(req => parseInt(req.numbericalId)) numericalId: number) {
      return { method };
    }
```
This could be prevented by explicitly defining the input type:
```ts
    @Get('SomeResource')
    @CustomParameters()
    public async customParameter( @CustomParameter<string>(req => parseInt(req.numbericalId)) numericalId: number) {
      return { method };
    }
```

4. I don't see why the route level `@CustomParameters` is necessary. Is there a way that it can be avoided? I feel like the route generation code should be able to identify that one of the parameters had the `@CustomParameter` decorators.

PR #196, opened 2018-02-14
Resetting controller status when it's singleton

AmazingTurtle:
Previously when having IoC and setting a response status it is remembered to the next request

### DIFF
--- a/src/routeGeneration/templates/express.ts
+++ b/src/routeGeneration/templates/express.ts
@@ -57,6 +57,9 @@ export function RegisterRoutes(app: any) {
 
             {{#if ../../iocModule}}
             const controller = iocContainer.get<{{../name}}>({{../name}});
+            if (typeof controller['setStatus'] === 'function') {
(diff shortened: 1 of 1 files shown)


AmazingTurtle:
Related: https://github.com/lukeautry/tsoa/issues/199

lukeautry:
Thanks!

andreasrueedlinger:
@AmazingTurtle @lukeautry why is the bugfix only in express template? And since typescript 2.7 this fix breaks the typescript transpilation with `error TS7017: Element implicitly has an 'any' type because type '<yourControllerType>' has no index signature.` if you have `"noImplicitAny": true`. I would suggest adding the fix in all the provided templates, but with an any cast:
```ts
if (typeof (<any>controller)['setStatus'] === 'function') {
  (<any>controller).setStatus(undefined);
}
```

Should I open an issue?

AmazingTurtle:
@andreasrueedlinger sorry, I forgot about that! I will take a look this evening.

PR #200, opened 2018-02-18
typescript-ioc docs as an alternative to inversify.js

KyleGalvin:
I wasn't a fan of inversify.js. The two-step setup and binding mechanism seemed cumbersome and counter-intuitive.
I managed to get tsoa set up with another library called typescript-ioc that has a much smaller code footprint

Here are some documentation updates for others that might be interested in the same

more examples available in https://github.com/KyleGalvin/backendBoilerplate
noteworthy example - database connection singleton injection:
https://github.com/KyleGalvin/backendBoilerplate/blob/master/src/models/typeorm.ts


lukeautry:
Thanks!

PR #217, opened 2018-04-23
Add RequestProp parameter decorator

nvonbulow:
There's the existing `Request` decorator to pass the whole request object into the controller method. However, with the way I'm testing (using the controller method directly), it's a pain to create the request object manually when all I want is the 'user' property. Using this decorator, I can directly pass in the user object, like
```ts
@Get('/{id}')
@Security('jwt')
@Tags('user')
public async getUser(@Path() id: string, @RequestProp() user: User) { /* verify permissions here */ }
```

And I'd just be able to call it in my test like shown below instead of manually instantiating an Express (or other framework) request
```ts
async function testGetUser() {
  const user = { ... }
  await expect(getUser(user.id, user)).resolves.toEqual(expectedResponse);
}
```

### DIFF
--- a/src/decorators/parameter.ts
+++ b/src/decorators/parameter.ts
@@ -32,6 +32,13 @@ export function RequestProp(name?: string): any {
 }
 
 /**
+ * Inject authenticated user (if any)
+ */
+export function User(): any {
+  return () => { return; };
+}
+
+/**
  * Inject value from Path
  *
  * @param {string} [name] The name of the path parameter
--- a/src/decorators/parameter.ts
+++ b/src/decorators/parameter.ts
@@ -23,6 +23,15 @@ export function Request(): any {
 }
 
 /**
+ * Inject value from http request
+ *
+ * @param {string} [name] The name of the request property
+ */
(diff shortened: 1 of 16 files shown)


endor:
The `User` parameter decorator seems project-specific to me. Why is it a good idea to add it to tsoa? 

I'm also not completely sold on the idea of `RequestProp` yet. If you need to instantiate a request for a test, wouldn't that just be abstracted into one method and that's it? Anyone else any opinions?

dgreene1:
@nvonbulow I think that object mocks and the like are up to the consumer. I can pretty effectively mock User via Jest. To be honest, I would recommend using supertest or some kind of API test running library which would execute the security middleware and would also simulate the request. So hopefully this shares some techniques and also why I think this PR is a bit unnecessary. I’m going to close it for housekeeping purposes (because the health of this library depends on new users’ perception of how well we keep up on issues). If someone else find this decorator useful, they can comment here and we’ll consider re-opening. Thank you for being a part of the tsoa community. Please keep the ideas coming since we find them valuable. :)

PR #222, opened 2018-05-22
Add "schemes" option into configuration

qqnc:
The protocol ( https | http )cannot be in the host, when importing to swagger hub with Swagger 2.0.  It needs a schemes option at the same level of "host".

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -38,7 +38,7 @@ export interface SwaggerConfig {
   entryFile: string;
 
   /**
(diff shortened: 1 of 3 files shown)


jamestharpe:
I did some quick investigating and found a workaround. You can add the following to your `tsoa.json` under `swagger` to set schemes:

```
"specMerging": "recursive",
"spec": {
  "schemes": ["http", "https"]
}
```

PR #228, opened 2018-05-30
Feature/produces file

ryo-nbx:
This is only a merge resolve of the PR https://github.com/lukeautry/tsoa/pull/154

So the original contributor of the code is vkagamlyk

Since the original PR has merge conflicts since almost a year I merged it to speed up the request.


ryo-nbx:
* added support for static file download, use FileResult.path for static files and FileResult.data for generated file
* you can offer a target file name with FileResult.filename but it will only be usable for static files
* if no path and no file data is provided status code 404 will be returned

ryo-nbx:
last update for now...

* found a problem with usage of _instanceof FileResult_ and anonymous object generation
  * added some convinience functions to FileResult to improve Handling
```
@SuccessResponse('200', 'successful')
@Produces('application/zip, application/octet-stream')
@Get('download/backup')
public async downloadBackup (): Promise<FileResult> {
    const store = service.getBackupStorage()
    await service.exportSystemBackup(store.fullPath)

    // does NOT WORK with instanceof
    // return { path: store.fullPath }

    // does work
    // const result = new FileResult()
    // result.path = store.fullPath
    // return result

    // does work
    // return new FileResult({ path: store.fullPath })

    // does work
    return FileResult.newInstance({ path: store.fullPath })
  }
```

* use sendFile() instead of Response.download() to achieve better compatibility with Swagger console
  * path must be absolute or specify root to res.sendFile

jamestharpe:
Is something needed to help get this pull request accepted? I'd really like to have this feature.

qqnc:
Need this feature. +1 Thanks

fhewitt:
In conflict again... :( This is a reasonable change, and a highly needed feature.

A part of me want to take the branch and resolve the conflicts, but it's a waste of time if the PR just stay there until new conflicts emerge. :(

ryo-nbx:
Fixed!

ryo-nbx:
https://blog.nubix.de/index.php/2018/09/21/file-download-and-upload-with-tsoa-hack/

I put an example online demonstrating how to achieve upload and download files with the current version of TSOA. Downloading is a little bit dirty since file responses are not handled yet by TSOA.

I hope the example might help some user here.

src/routeGeneration/templates/express.ts
```diff
@@ -135,7 +142,23 @@ export function RegisterRoutes(app: any) {
                 }
 
                 if (data || data === false) { // === false allows boolean result
-                    response.status(statusCode || 200).json(data);
+                    if (data instanceof FileResult) {
+                        if (data.path) {
+                          response.status(statusCode | 200);
```
dgreene1: There are a couple of `|`’s here that should be `||`’s

tests/fixtures/express/routes.ts
```diff
@@ -1,5 +1,7 @@
 /* tslint:disable */
-import { Controller, ValidateParam, FieldErrors, ValidateError, TsoaRoute } from '../../../src';
```
dgreene1: The routes.ts files are supposed to be gitignored, so whoever picks up this PR will need to run `git rm` on each of the routes.ts flies in the koa, express, etc folders.

src/decorators/produces.ts
```diff
@@ -0,0 +1,3 @@
+export function Produces(mediaType: string): any {
```
dgreene1: Shouldn’t string be a string literal of all of the possible media types that are allowed?

src/routeGeneration/templates/express.ts
```diff
@@ -1,8 +1,10 @@
 /* tslint:disable */
+import { Readable } from 'stream';
+
 {{#if canImportByAlias}}
```
dgreene1: If we handle this for express then we must handle it for koa and Hapi

drothmaler:
Thanks a lot for the review... @ryo-nbx is currently on vacation, so it will take some time, until he is back and will be able to to the necessary fixes.

src/decorators/produces.ts
zlumer: There are approx. [2500 different media types](https://www.iana.org/assignments/media-types/media-types.xhtml). Doesn't seem to be practical to include them all as a union type and then maintain the list. And providing only a subset is definitely worse than just allowing any string, leaving the control to the user.

dgreene1: Good point. I’ll mark this comment as resolved.

dgreene1:
@drothmaler / @ryo-nbx would you be interested in reopening this and making the requested changes?

nmai:
@ryo-nbx 🙏

PR #229, opened 2018-05-31
Fix typescript at 2.7.x for npm (non-yarn) users

ronjouch:
npm users started getting TS 2.9.1 on 2018-05-31, which
break tsoa builds with errors
`TypeError: Cannot read property 'getTypeAtLocation' of undefined`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,4685 @@
+{
+  "name": "tsoa",
+  "version": "2.1.4",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@types/accepts": {
+      "version": "1.3.5",
+      "resolved": "https://registry.npmjs.org/@types/accepts/-/accepts-1.3.5.tgz",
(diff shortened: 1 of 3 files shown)


ronjouch:
@lukeautry yeah! I was being conservative. But 2.8.4 has been working fine for my use case for weeks, so (to me) it's 👍. Updating the PR with a new package-lock and yarn.lock.

ronjouch:
@lukeautry done in dde11c1

ronjouch:
@lukeautry please wait before merging this.

- I'll do a proper regenerate of the `yarn.lock`. I had manually bumped typescript with the intent of touching as little as possible but, in hindsight, it's a bad idea as it ignores new/changed deps of `typescript@~2.8.4` and bugfixes.
- I'll also add an npm `package-lock.json`; feels weird to use locking for yarn but not npm.

ronjouch:
@lukeautry done in 3446d9d

Issue #225, opened 2018-05-25
HATEOAS support / get route by controller method


jamestharpe:
I'm working on a project that requires [HATEOAS](https://en.wikipedia.org/wiki/HATEOAS) support. To accomplish this, I need to be able to get a route based on a controller method.

For example:

```javascript
const pathToUser = routeTo(
  UsersController.prototype.getUser, 
  { usedId: 1 }); // returns '/baseUrl/users/1'
```

I've got this implemented and working in my project already and I was wondering if you'd be interesting in a contribution to TSOA to make it part of the framework.

Alternatively, I could publish it as a separate library that would replace the tsoa routing decorators.

dgreene1:
Closing due to inactivity in the PR: https://github.com/lukeautry/tsoa/pull/234#issuecomment-520676547

PR #234, opened 2018-06-09
Adds routeTo method for HATEOAS support

jamestharpe:
Addresses #225 . 

Sorry, I wasn't sure how to organize tests for this. Happy to update it w/ some guidance.

This makes it easy to add HATEOAS style links to response models.

Example: 

```typescript
@Get('{parentId}')
async getParent(parentId: string): Promise<Parent> {
  const result = {
    id: parentId,
    name: "Daddy",
    links: [
      { 
        rel: 'children',
        href: routeTo(ChildrenController.prototype.getChildren, {
            parentId: parentId
          })
      }
    ]
  }
}
```

### DIFF
--- a/src/decorators/methods.ts
+++ b/src/decorators/methods.ts
@@ -1,19 +1,56 @@
+export const keys = {
+  baseRoute: '__base_route',
+  controllerClass: '__controller_class',
+  pathExpression: '__pathExpression',
+};
+
+export function routeTo(handler, args?: object) {
+  const baseRoute = handler[keys.controllerClass][keys.baseRoute];
+  let pathValue: string = handler[keys.pathExpression] || '';
+  if (args && pathValue) {
+    const argProps = Object.getOwnPropertyNames(args);
+    for (const argProp of argProps) {
+      pathValue = pathValue.replace('{' + argProp + '}', args[argProp]);
+    }
+  }
+  return (
+    '/' + baseRoute + '/' + encodeURI(pathValue)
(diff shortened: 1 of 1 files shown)


endor:
Maybe you could add a file to `tests/unit/swagger/common` and add a few tests cases for the existing `GetTestController` with valid and invalid uses of `routeTo`.

src/decorators/methods.ts
```diff
@@ -1,19 +1,56 @@
...
+    for (const argProp of argProps) {
+      pathValue = pathValue.replace('{' + argProp + '}', args[argProp]);
+    }
+  }
+  return (
+    '/' + baseRoute + '/' + encodeURI(pathValue)
+  );
+}
+
+// const saveRoute =
```
endor: is this intentional?

jamestharpe:
Thanks for the feedback, I'll make some updates!

src/decorators/methods.ts
jamestharpe: Nope! I'll update, thanks!

dgreene1:
@jamestharpe if you want to update this with tests, we’d love to introduce hateos support. There are a lot of recent PRs that provide examples. Additionally we’d need to see some tests that verify that the Schema output has the proper links information as described in openAPI: https://swagger.io/docs/specification/links/

In the meantime I’m going to close the PR since it’s rather old. If you’d like to reopen it, please feel free to. :)

PR #246, opened 2018-07-09
Allow setting status and headers in controllers that do not extend Controller

endor:
The reason for this change is that I have difficulty extending my controllers from `Controller` when using `inversify` as then I would need to somehow be able to inject into the `Controller`-constructor. After this change it doesn't matter if controllers extend `Controller`, it only checks for the necessary methods. Does that make sense?

ping @AmazingTurtle @jamestharpe @abmohan

### DIFF
--- a/src/routeGeneration/templates/express.ts
+++ b/src/routeGeneration/templates/express.ts
@@ -98,18 +98,21 @@ export function RegisterRoutes(app: any) {
     }
     {{/if}}
 
+    function isController(object: any): object is Controller {
+        return 'getHeaders' in object && 'getStatus' in object && 'setStatus' in object;
+    }
+
     function promiseHandler(controllerObj: any, promise: any, response: any, next: any) {
         return Promise.resolve(promise)
             .then((data: any) => {
                 let statusCode;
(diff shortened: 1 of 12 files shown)


src/routeGeneration/templates/express.ts
```diff
@@ -98,18 +98,21 @@ export function RegisterRoutes(app: any) {
...
+    }
+
     function promiseHandler(controllerObj: any, promise: any, response: any, next: any) {
         return Promise.resolve(promise)
             .then((data: any) => {
                 let statusCode;
-                if (controllerObj instanceof Controller) {
-                    const controller = controllerObj as Controller
-                    const headers = controller.getHeaders();
+                if (isController(controllerObj)) {
```
lukeautry: I think it might be clearer to just inline these checks (e.g. see if controllerObj.getHeaders exists and is a function, and then call it) rather than extracting the function

lukeautry:
This looks good except for the style thing I pointed out.

src/routeGeneration/templates/express.ts
lukeautry: Actually, this is fine.

PR #249, opened 2018-07-27
Spec generator for OpenAPI 3.0.0

endor:
This is a first version of a spec generator for OpenAPI 3.0.0. Please check it out and let me know if this makes sense of if it should be done some other way. As you can see from the TODO at the top, not everything is done yet, but it should be able to produce most of the things that the current spec generator can and I guess users can always switch back if it doesn't yet support something they need. My thinking is that the sooner this gets merged the easier it'll be for other people to extend.

Pinging the other people that want to help maintain tsoa:
@AmazingTurtle @jamestharpe @0v3rst33r @abmohan

Here are examples of v2 vs. v3:
https://mermade.org.uk/examples/swagger.json
https://mermade.org.uk/examples/openapi.json

And here are some more useful links:
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md
https://swagger.io/blog/news/whats-new-in-openapi-3-0/
https://blog.readme.io/an-example-filled-guide-to-swagger-3-2/

### DIFF
--- a/src/metadataGeneration/tsoa.ts
+++ b/src/metadataGeneration/tsoa.ts
@@ -25,7 +25,6 @@ export namespace Tsoa {
     summary?: string;
     isHidden: boolean;
     operationId?: string;
-
   }
 
   export interface Parameter {
--- a/src/module/generate-swagger-spec.ts
+++ b/src/module/generate-swagger-spec.ts
@@ -3,7 +3,9 @@ import * as YAML from 'yamljs';
 import { SwaggerConfig } from '../config';
 import { MetadataGenerator } from '../metadataGeneration/metadataGenerator';
 import { Tsoa } from '../metadataGeneration/tsoa';
-import { SpecGenerator } from '../swagger/specGenerator';
+import { SpecGenerator2 } from '../swagger/specGenerator2';
+import { SpecGenerator3 } from '../swagger/specGenerator3';
+import { Swagger } from '../swagger/swagger';
 import { fsExists, fsMkDir, fsWriteFile } from '../utils/fs';
 
 export const generateSwaggerSpec = async (
@@ -22,7 +24,13 @@ export const generateSwaggerSpec = async (
       ignorePaths,
     ).Generate();
   }
-  const spec = new SpecGenerator(metadata, config).GetSpec();
+
+  let spec: Swagger.Spec;
+  if (config.version && config.version === '3.0.0') {
+    spec = new SpecGenerator3(metadata, config).GetSpec();
+  } else {
+    spec = new SpecGenerator2(metadata, config).GetSpec();
+  }
 
   const exists = await fsExists(config.outputDirectory);
   if (!exists) {
--- a/src/module/generate-swagger-spec.ts
+++ b/src/module/generate-swagger-spec.ts
@@ -26,7 +26,7 @@ export const generateSwaggerSpec = async (
   }
 
   let spec: Swagger.Spec;
-  if (config.version && config.version === '3.0.0') {
+  if (config.version && config.version.split('.')[0] === '3') {
     spec = new SpecGenerator3(metadata, config).GetSpec();
   } else {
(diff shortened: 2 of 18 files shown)


src/cli.ts
```diff
@@ -154,7 +156,12 @@ function swaggerSpecGenerator(args) {
     const compilerOptions = validateCompilerOptions(config.compilerOptions);
     const swaggerConfig = validateSwaggerConfig(config.swagger);
     const metadata = new MetadataGenerator(swaggerConfig.entryFile, compilerOptions, config.ignore).Generate();
-    const spec = new SpecGenerator(metadata, config.swagger).GetSpec();
+    let spec: Swagger.Spec;
+    if (config.swagger.version && config.swagger.version === '3.0.0') {
```
fhewitt: If someone try to set the version to 3.0.1 (current version), I suspect he will expect to get SpecGenerator3, not SpecGenerator2...

endor: I made it more generic now, so it'll use the `SpecGenerator3`, but I'm not sure about the warning. I'm not sure how we should handle those subversions in general. The code could get quite messy if we wanted to support them all. We could try to always support the latest. Since this is a first implementation it doesn't yet, but it also doesn't come with all the features yet. These could be next steps.

lukeautry:
Been meaning to look at this - hopefully I can get to it soon. I like the approach of maintaining support for both, despite the added complexity.

lukeautry:
@endor These are some kind of nasty merge conflicts - if you can figure out the merge I'd be happy to get this merged in. It would be great to say we support Swagger 3.0.

src/cli.ts
AmazingTurtle: I suggest using the latest version always and writing out the generated spec version. I wouldn't mind supporting 2.x, rather deprecate it in a major version bump

endor:
Conflicts fixed. I'm not sure we can say we support OpenAPI 3.0.0 after this, but it's at least a big step closer. Then people can start implementing the missing features and I guess if more people start using it to generate 3.0.0 specs we'll also notice potential bugs (:

endor:
Fixed again. @lukeautry should we merge this?

lukeautry:
@endor Sorry for the delay on this - if you can fix the merge conflicts, please merge right away and I'll get it into a release. I think this will solve several issues at once, and even if it's not 100% supported, we have a good jumping off point.

leolorenzoluis:
@endor Would you be able to merge the conflicts please? 💯

dgreene1:
@endor would you be willing to resolve the conflicts again?

endor:
I will try to get to it on Monday or Tuesday.

endor:
Since this has been going through 3 rebases, I would love for someone to check this before I or someone else merges it.

0v3rst33r:
I would gladly have a look at it @endor (will probably only get to it tomorrow evening if that is ok).

I want to get into the development effort here, and this seems like as good an opportunity as any 😄

lukeautry:
Thanks @0v3rst33r. I don't want to merge anything else until this goes in since it has been open for so long, so if someone familiar with OpenAPI 3.0 can verify the functionality here, we can get it merged.

dgreene1:
@lukeautry what if we just merged this as is?

I think this would be safe enough to do since the code defaults to version 2, so that would be non-breaking for current `tsoa` users. Additionally, those who are excited enough to try out `tsoa` with version 3, would be able to submit issues to us (which would get us the valuable feedback we want). Essentially we could merge this as beta functionality. It seems like the most Agile thing to do. And plus we would then have the PR merged and we could stop worrying about rebasing (I know that would make @endor happy, I bet 😃).

dgreene1:
@endor, @lukeautry any thoughts on the “give it a shot”/ Agile approach I’m recommending above? https://github.com/lukeautry/tsoa/pull/249#issuecomment-496583438

endor:
From me it's obviously a go ;)

PR #253, opened 2018-08-10
Also evaluate Examples when there is a SuccessResponse; fixes #245

### LINKED ISSUE #245: Using Example & SuccessResponse Decorators
Is it possible to use both of them ? I want to add a description and an example for my success response on my endpoint but the swagger generated does not include my example.

NB: each decorator works fine separately

endor:


jpramondon:
If I could approve this one, I'd do :D
We faced the exact same issue and fixed it the same way when we realized the function grasping the example was never called.

endor:
@jpramondon Since you basically made the same fix, I'll merge this soonish. I'll still check your PR for the other change later today.

PR #256, opened 2018-08-13
Fix swagger.json examples

### LINKED ISSUE #255: Missing examples in swagger.conf files
Hi there !

We've been using TSOA for quite a bit of time now (more than a year, that is) and we've been asked to improve our swagger.json file with some extra documentation. So we thought we could make good use of the existing _@Response_, _@SuccessResponse_ and _@Example_ decorators to do such improvement.

We came to 2 major issues combining these decorators that you will find described below. Please note that these issues have been addressed successfuly in a local 2.1.8-b version of TSOA (built out of a 2.1.8) and the changes have been reported in a 2.1.9 version currently submitted as a pull request named "Fix swagger.json examples".

Now for the tiny problems : 

- *@Response*: it sports a nice feature where a developer may pass the decorator an example as an object. The problem is the example is never written in the generated swagger.json file (due to some code that never gets executed, see the PR).

- *@SuccessResponse*: from the code, one can learn that this decorator should work well when combined with @Example, linking the provided example to the success response's description in the generated swagger.json file. The problem is the example can never be found there (due to a function that is not called, again see the PR for full details).

Thank you for reading and happy review !

jpramondon:
A quick change so that issue #255 on examples is fixed
Close #255

### DIFF
--- a/src/metadataGeneration/methodGenerator.ts
+++ b/src/metadataGeneration/methodGenerator.ts
@@ -120,7 +120,7 @@ export class MethodGenerator {
       if (expression.arguments.length > 1 && (expression.arguments[1] as any).text) {
         description = (expression.arguments[1] as any).text;
       }
-      if (expression.arguments.length > 2 && (expression.arguments[2] as any).text) {
+      if (expression.arguments.length > 2 && expression.arguments[2]) {
         const argument = expression.arguments[2] as any;
         examples = this.getExamplesValue(argument);
       }
@@ -155,7 +155,7 @@ export class MethodGenerator {
 
     let description = '';
     let name = '200';
-    const examples = undefined;
+    const examples = this.getMethodSuccessExamples();
 
     if (expression.arguments.length > 0 && (expression.arguments[0] as any).text) {
       name = (expression.arguments[0] as any).text;
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "tsoa",
-  "version": "2.1.5",
+  "version": "2.1.8",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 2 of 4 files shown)


jpramondon:
Related to #253

jpramondon:
Related to #208

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "tsoa",
   "description": "Build swagger-compliant REST APIs using TypeScript and Node",
-  "version": "2.1.8",
+  "version": "2.1.9",
```
endor: the releases are made in separate commits.

src/metadataGeneration/methodGenerator.ts
```diff
@@ -120,7 +120,7 @@ export class MethodGenerator {
       if (expression.arguments.length > 1 && (expression.arguments[1] as any).text) {
         description = (expression.arguments[1] as any).text;
       }
-      if (expression.arguments.length > 2 && (expression.arguments[2] as any).text) {
+      if (expression.arguments.length > 2 && expression.arguments[2]) {
```
endor: what exactly does this fix?

jpramondon: This part of the code is where the example is picked from the _@Response_ decorator.
For example:
```
@Response(400, "If something goes wrong in the request", {errorCode:1209, message:"you sent a bad parameter"})
```

I believe the problem with the previous code is that it was treated like a character string out of the _text_ attribute of an expression object, which it is not at this point. It really looks more like an _any_ typed structure that has of course not the said _text_ attribute, hence the removal of the check.

Removing this tiny extra condition magically makes the example to appear in the swagger documentation again.

endor:
@jpramondon Ah, still have to ask two things.

1. Could you combine the separate commits into one? It doesn't really make sense to have for example "Reverting to 2.1.8" in the history. 
2. Could you write a unit test for this response decorator example? For example in the `methodController.ts` there are already `Response` decorators in use. It's probably enough to extend `metadata.spec.ts`, if it fails before the fix and succeeds after.

jpramondon:
Job done!
- Commits have been squashed into one
- Unit tests have been added, inspired by #208

endor:
Hm, is github just displaying it wrong? Looks like there are 5 commits?

jpramondon:
I did a `rebase -i`, then committed again, which gave commit _aeb7763_
Should have I done otherwise ?

jpramondon:
I'm closing this one as I'm lost in a merge hell :/
Despite efforts to squash my commits, I always end up with a merge popping. So I'll be creating a new branch and pull request today with only A do-it-all commit.

PR #259, opened 2018-08-21
Allow combining two security methods with AND instead of OR

endor:
I need this rather urgently for my current project, so if you have time @lukeautry, please check it out. @jpramondon Welcome to review as well.

### DIFF
--- a/README.MD
+++ b/README.MD
@@ -178,9 +178,9 @@ export class UsersController {
     }
 }
 ```
-Note that the parameter `request` does not appear in your swagger definition file. 
+Note that the parameter `request` does not appear in your swagger definition file.
 Likewise you can use the decorator `@Inject` to mark a parameter as being injected manually and should be omitted in swagger generation.
-In this case you should write your own custom template where you inject the needed objects/values in the method-call. 
+In this case you should write your own custom template where you inject the needed objects/values in the method-call.
 
 ### Dependency injection / IOC
 
@@ -307,7 +307,7 @@ public async getResponse(): Promise<TestModel> {
 }
 ```
 
-### Authentication 
+### Authentication
 
 Authentication is done using a middleware handler along with `@Security('name', ['scopes'])` decorator in your controller.
 
@@ -342,7 +342,7 @@ First, define the security definitions for swagger, and also configure where the
 ```
 
 
-In the middleware, export the function based on which library (Express, Koa, Hapi) you are using. You only create 1 function to handle all authenticate types. The `securityName` and `scopes` come from the annotation you put above your controller function. 
+In the middleware, export the function based on which library (Express, Koa, Hapi) you are using. You only create 1 function to handle all authenticate types. The `securityName` and `scopes` come from the annotation you put above your controller function.
 `./authentication.ts`
 ```ts
 import * as express from 'express';
@@ -364,7 +364,7 @@ export function expressAuthentication(request: express.Request, securityName: st
             return Promise.reject({});
         }
     }
-    
+
     if (securityName === 'jwt') {
         const token = request.body.token || request.query.token || request.headers['x-access-token'];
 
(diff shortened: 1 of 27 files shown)


src/metadataGeneration/tsoa.ts
```diff
@@ -43,8 +43,7 @@ export namespace Tsoa {
   }
 
   export interface Security {
-    name: string;
-    scopes?: string[];
+    [key: string]: string[];
```
jpramondon: Too bad we're losing the _Security_ object true definition. Are we ?

src/routeGeneration/templates/express.ts
```diff
@@ -76,23 +76,42 @@ export function RegisterRoutes(app: any) {
...
                     }
-                })
-                .catch((error: any) => {
-                    responded++;
-                    if (responded == security.length && !success) {
-                        response.status(401);
-                        next(error)
+
+                    Promise.all(promises)
+                        .then((values) => { succeed(values[0]); })
```
jpramondon: Is it possible that a more explicit name than _values_ is used there ? Or does _expressAuthentication_ returns something not that much defined ?

src/routeGeneration/templates/express.ts
```diff
@@ -76,23 +76,42 @@ export function RegisterRoutes(app: any) {
...
             let responded = 0;
             let success = false;
+
+            const succeed = function(user: any) {
+                if (!success) {
+                    success = true;
+                    responded++;
+                    request['user'] = user;
+                    next();
+                }
```
jpramondon: Could _responded_ and the _next_ function call be outside of the _if_ ?

README.MD
```diff
@@ -531,6 +531,32 @@ The according swagger definition can be merge-overwritten inside `tsoa.json`. He
...
+```ts
+@Security({
+  tsoa_auth: ['write:pets', 'read:pets'],
+  api_key: [],
+})
+@Get('OauthAndAPIkey')
+public async GetWithAndSecurity(@Request() request: express.Request): Promise<any> {
+}
+```
+
```
jpramondon: Documentation becomes highly required as the difference is really subtle between _AND_ and _OR_ combinations. One could really get confused without it. Good point.

jpramondon:
Not being an expert on Hapi or Koa makes it difficult to review these parts with relevant comments

src/metadataGeneration/tsoa.ts
endor: Sorry, I don't know what that means.

src/routeGeneration/templates/express.ts
endor: Ah, I removed the comment there. I think the idea - and I can only guess because I didn't write it - was that we're not calling next multiple times here.

src/routeGeneration/templates/express.ts
endor: It is actually open what `expressAuthentication` returns. Could potentially also be a token object for instance. But in practice it's probably most often a user and since that's also used in `succeed`, I'll fix it here.

endor:
The same is true for me. I've basically just copied the code from express.

src/metadataGeneration/tsoa.ts
fhewitt: I think the idea of the comment is that a security was "a name and a scope list" now it's "a hashmap of string array." Therefore, on the autocomplete and code comprehension, we've lost something.

endor: I guess the alternative would be to have a return type in the `methodGenerator:233` of `[[Tsoa.Security]]`, where the inner array denotes a combination of AND and the outer array denotes a combination of OR. I thought the way I did it would be easier to grasp, but yes, some information is lost. Do you guys think I should do it the other way? At least `Validators` uses a similar pattern, but I admit, that it's clearer there what the parts mean.

jpramondon: So that's a matter of choosing between code readability and ease of use ? I'm betting on "ease of use". I know it might not go in the way of maintainability, but if it's easier to grasp for an end user ...

endor: How does this make a difference for an end user? They're just using the decorators. They won't even see this internal representation.

PR #267, opened 2018-08-31
change decorator return type to Function for no-unsafe-any tslint rule.

### LINKED ISSUE #266: no-unsafe-any on decorators
Looking through the code, all the decorators (@Get, @Route, @Post, etc...) seem to return a `Function` yet the typescript definition files has them returning `any`. This, of course, breaks the `no-unsafe-any` tslint rule. Any way we can change these definitions from:

```
export declare function Get(value?: string): any;
export declare function Post(value?: string): any;
export declare function Put(value?: string): any;
export declare function Patch(value?: string): any;
export declare function Delete(value?: string): any;
```

to 

```
export declare function Get(value?: string): Function;
export declare function Post(value?: string): Function;
export declare function Put(value?: string): Function;
export declare function Patch(value?: string): Function;
export declare function Delete(value?: string): Function;
```
?

(for all the decorators, not just method.js)

I'll create a PR for this if it is something you want.

pburrows:
fixes #266

Turned decorator return types to Function.

Also fixed some unrelated tslint errors on initial pull (minor things like alphabetizing and double-quotes vs single-quotes).

### DIFF
--- a/src/interfaces/decorator-return.ts
+++ b/src/interfaces/decorator-return.ts
@@ -1,0 +1,3 @@
+export type MethodDecoratorReturn = (target: any, propertyKey: string, descriptor: PropertyDescriptor) => void;
+export type ClassDecoratorReturn = (constructor: any) => void;
+export type ParameterDecoratorReturn = (target: object, propertyKey: string | symbol, parameterIndex: number) => void;
--- a/tslint.json
+++ b/tslint.json
@@ -4,7 +4,6 @@
   ],
   "rules": {
     "arrow-parens": false,
-    "ban-types": false,
     "trailing-comma": [
       true,
       {
--- a/tslint.json
+++ b/tslint.json
@@ -4,6 +4,7 @@
   ],
   "rules": {
     "arrow-parens": false,
+    "ban-types": false,
     "trailing-comma": [
       true,
       {
--- a/src/decorators/example.ts
+++ b/src/decorators/example.ts
@@ -1,3 +1,3 @@
-export function Example<T>(exampleModel: T): any {
+export function Example<T>(exampleModel: T): Function {
   return () => { return; };
 }
--- a/src/decorators/example.ts
+++ b/src/decorators/example.ts
@@ -1,3 +1,5 @@
-export function Example<T>(exampleModel: T): Function {
+import { MethodDecoratorReturn } from '../interfaces/decorator-return';
+
+export function Example<T>(exampleModel: T): MethodDecoratorReturn {
   return () => { return; };
 }
--- a/src/decorators/methods.ts
+++ b/src/decorators/methods.ts
@@ -1,19 +1,21 @@
-export function Get(value?: string): Function {
+import { MethodDecoratorReturn } from '../interfaces/decorator-return';
+
+export function Get(value?: string): MethodDecoratorReturn {
   return () => { return; };
 }
 
-export function Post(value?: string): Function {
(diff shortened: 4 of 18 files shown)


tslint.json
```diff
@@ -4,6 +4,7 @@
   ],
   "rules": {
     "arrow-parens": false,
+    "ban-types": false,
```
endor: If we made the return type more specific, we wouldn't need to change this here, right?

endor:
Why add "dist"?

tslint.json
pburrows: `ban-types` by default will ban return types of `Function`. Since it wasn't used anywhere else, I turned off the rule. 

If you don't like turning off that tslint rule for the tsoa project, another option would be to change the return types from Function to something like DecoratorReturn, and then in another file, define decorator return like this:

```
export type DecoratorReturn = () => void;
```

pburrows:
Ugh. I added /dist as a second commit so I could use it in a test app through npm. I didn't realize it would update this PR! My bad, I'll remove it.

endor:
Out of curiosity.. the `no-unsafe-any` doesn't seem to be part of the recommended rules for tslint. Why are you using it?

tslint.json
endor: Yeah, I think that would be nicer.

pburrows:
I use the [microsoft tslint extensions](https://github.com/Microsoft/tslint-microsoft-contrib/blob/master/recommended_ruleset.js), which adds a lot of additional extensions for common security mistakes, and common 'gotchas' that larger teams of developers can often run into. Their [recommended ruleset](https://github.com/Microsoft/tslint-microsoft-contrib/blob/master/templates/recommended_ruleset.js.snippet) has no-unsafe-any turned on.

As a practical matter, it stops developers from using the `any` salt-shaker, and sprinkling it all over the place. Makes people start thinking through the actual types that get returned. 

Essentially, if you HAVE to use `any` somewhere, it should be a defeat, and come with a comment as to why it was used (as well as a tslint comment to disable the any).

tslint.json
pburrows: ok, I'll add that. I'll put the DecoratorReturn definition in the src/interfaces/ folder.

src/interfaces/decorator-return.ts
```diff
@@ -0,0 +1,3 @@
+export type MethodDecoratorReturn = (target: any, propertyKey: string, descriptor: PropertyDescriptor) => void;
```
endor: I don't yet understand where these signatures come from. Could you explain that?

endor:
Please squash the commits into one.

pburrows:
doesn't squash happen automatically when you merge and choose "squash and merge"?

MatthiasKunnen:
You could try: `git rebase -i 53be8b116442b76bb1b75207ca5efb92391224bf` and replace `pick` with `s` which stands for squash. 53be8b116442b76bb1b75207ca5efb92391224bf is the first commit you made on this branch. If the SHA has changed locally for you, you'll have to change this.

endor:
Strange. @jpramondon also recently had problems with this. You are right though, that there's a `Squash and merge` option. If @MatthiasKunnen's suggestion doesn't work, I can try that next.

jpramondon:
It ended up working, but it came to the price of having to delete the project, fresh check it out again, move to the PR branch and `git rebase -i HEAD~x` it.

pburrows:
ok, squashing seems to have done something... Hope it worked!

PR #269, opened 2018-09-05
fix install package from github directly

Michsior14:
The `prepublish` doesn't run when installing the package directly from Github. The solution is to use `prepare` hook. [NPM issue](https://github.com/npm/npm/issues/3055)


endor:
I'm confused. When I read here: https://docs.npmjs.com/misc/scripts, it says:

> prepublish: Run BEFORE the package is packed and published, as well as on local npm install without any arguments. (See below)
> prepare: Run both BEFORE the package is packed and published, and on local npm install without any arguments (See below). This is run AFTER prepublish, but BEFORE prepublishOnly.

Where is the difference between the two, except that one runs after the other? I guess the documentation hasn't been updated?

PR #270, opened 2018-09-06
[#271] Updates template and tests for hapi v17

slapersonne:
#271 
This pull requests aims to use tsoa with the breaking changes of hapi v.17, including usage of [h response toolkit](https://hapijs.com/api/17.0.0#response-toolkit), and replacement of callbacks by promises or async/await handling.
It also includes some fixes about the typing of tests' security controller.


jpramondon:
I support this PR ! :D
Long story short, Simon (@slapersonne) and the rest of our team are moving from Express to Hapi as an experiment on a new project. We discovered the last v17 release wasn't working with a 2.1.8 tsoa.
So Simon did the port work and we are now giving his work back to the community.

src/routeGeneration/templates/hapi.ts
```diff
@@ -41,15 +41,16 @@ export function RegisterRoutes(server: any) {
...
                 pre: [
                     {
-                      method: authenticateMiddleware({{json security}})
+                      method: authenticateMiddleware({{json security}}),
+                      assign: "user"
                     }
                 ],
                 {{/if}}
-                handler: (request: any, reply: any) => {
+                handler: (request: any, h: any) => {
```
endor: what does `h` stand for?

jpramondon: I asked myself the same one and the only match I could find is Hapi ... just a guess.
@slapersonne ?

PR #272, opened 2018-09-07
[#176] Add a description for tags.

### LINKED ISSUE #176: global tags configuration
Hi there, 

we need sonthing like this - to define the tags - inside the root of the swagger.json

grafik

Thx!

Greetings
crazyx13th

jstienne:
- Add tags for swagger config
- Add documentation for tags in the README.MD

This PR resolves the #176 issue. It permits to generate this part inside the root of the swagger.json : 
```
"tags": [
    {
        "name": "User",
        "description": "Operations about users",
        "externalDocs": {
            "description": "Find out more about users",
            "url": "http://swagger.io"
        }
    }
]
```

Close #176

### DIFF
--- a/README.MD
+++ b/README.MD
@@ -527,6 +527,31 @@ The according swagger definition can be merge-overwritten inside `tsoa.json`. He
 }
 ```
 
+### Tags
+
+If you have a project that needs a description and/or external docs for tags, you can configure the internal generators to use the correct tags definitions and external docs by providing a tags property to swagger property in tsoa.json.
+
+```json
+{
+  "swagger": {
+    "tags":  [
+      {
+        "name": "User",
+        "description": "Operations about users",
+        "externalDocs": {
+          "description": "Find out more about users",
+          "url": "http://swagger.io"
+        }
+      }
+    ],
+    ...
+  },
+  "routes": {
+    ...
+  }
+}
+```
+
 ### Use awesome Swagger tools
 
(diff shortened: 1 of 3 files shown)


jpramondon:
Indeed, we might have a design issue there. We'll discuss about the possible use of the conf file instead in the coming few days.
Thanks for your feedback.

jpramondon:
Justine changed her code to follow the better "by configuration" approach that @endor suggested.
Documentation is still due though, but should follow suit.

endor:
Would be nice to get the squashed commits.

README.MD
```diff
@@ -527,6 +527,54 @@ The according swagger definition can be merge-overwritten inside `tsoa.json`. He
...
+    ],
+    ...
+  },
+  "routes": {
+    ...
+  }
+}
+```
+
+Tags are defined with the `@Tags('tag1', 'tag2', ...)` decorator in the controllers and/or in the methods like in the following examples.
```
endor: This part of the documentation should go to the `Decorators` section.

endor: I think we should remove the part about the decorator being available also for controllers. It's true, but I don't yet understand what the outcome is. I guess there isn't one? Like.. it does nothing?

README.MD
```diff
@@ -527,6 +527,54 @@ The according swagger definition can be merge-overwritten inside `tsoa.json`. He
 }
 ```
 
+### Tags
+
+If you have a project that needs a description and/or external docs for tags, you can configure the internal generators to use the correct tags definitions and external docs by providing a tags property to swagger property in tsoa.json.
```
endor: 👍

Issue #188, opened 2018-01-29
operationId Question


taicho:
I'm making some adjustments locally to use tsoa in conjunction with swagger-tools. I was just curious, what was the reasoning behind not just making the operationId that is generated for each route in the swagger file not just match the method name(s) in the controller(s) exactly rather than automatically uppercase the first character as it does currently?

taicho:
To be clear I have a workaround for my purposes but still, I was curious as to the reasoning behind this logic.

egandro:
I also didn't like the idea that in 2.x the Name of the Controller is no longer present in the OperiationId.

We should make this more flexible.

lukeautry:
I think there probably should be an @OperationId decorator that overrides any 'inferred' operationId

redlock:
I've created a pull request that adds OperationId() decorator #273

lukeautry:
Fixed with #277

StingyJack:
I dont mean to necro this post but I think the point was that if there are two controllers like this

``` ts
@route('routeA')
export class ControllerA {
  @post("activate")
  public async activate() {
etc...


@route('routeB')
export class ControllerB {
  @post("activate")
  public async activate() {
etc...

```
... that tsoa should not generate an invalid OAS file that has duplicate operationID's of "activate", but rather should generate operationId's like "routeA_activate" and "routeB_activate" without having to specify any additional decorators.

WoH:
You can do that via `config.operationIdTemplate`.

PR #273, opened 2018-09-08
[#188] Added an OperationId decorator that overrides inferred id

redlock:


src/metadataGeneration/methodGenerator.ts
```diff
@@ -200,6 +201,21 @@ export class MethodGenerator {
...
+      return undefined;
+    }
+    if (opDecorators.length > 1) {
+      throw new GenerateMetadataError(`Only one OperationId decorator allowed in '${this.getCurrentLocation}' method.`);
+    }
+
+    const decorator = opDecorators[0];
+    const expression = decorator.parent as ts.CallExpression;
+    const ops = expression.arguments.map((a: any) => a.text as string);
+    return ops[0]
```
endor: `;` missing

src/swagger/specGenerator.ts
```diff
@@ -100,7 +100,11 @@ export class SpecGenerator {
     const pathMethod: Swagger.Operation = pathObject[method.method] = this.buildOperation(controllerName, method);
     pathMethod.description = method.description;
     pathMethod.summary = method.summary;
-    pathMethod.tags = method.tags;
+    
+    pathMethod.tags = [controllerName].concat(method.tags || [])
```
endor: What does this do?

src/swagger/specGenerator.ts
```diff
@@ -100,7 +100,11 @@ export class SpecGenerator {
     const pathMethod: Swagger.Operation = pathObject[method.method] = this.buildOperation(controllerName, method);
     pathMethod.description = method.description;
     pathMethod.summary = method.summary;
-    pathMethod.tags = method.tags;
+    
+    pathMethod.tags = [controllerName].concat(method.tags || []) 
+    
+    // Use operationId tag otherwise fallback to generated. Warning: This doesn't check uniqueness.
```
endor: Should uniqueness be checked?

endor:
It would be good to:

* Add unit tests to show how this works (one example where it's overridden and one where it's not)
* Add documentation to the README in the decorators section
* Squash the commits

src/swagger/specGenerator.ts
redlock: This creates a default tag based on the controller name. I'm using it to group methods by controller name. I will remove it from the pull request as it is another modification that isn't related to operationId.

src/swagger/specGenerator.ts
endor: Okay, let's go without it for now. Then if people later want it, we can still add it in.

src/swagger/specGenerator.ts
```diff
@@ -100,7 +100,11 @@ export class SpecGenerator {
     const pathMethod: Swagger.Operation = pathObject[method.method] = this.buildOperation(controllerName, method);
     pathMethod.description = method.description;
     pathMethod.summary = method.summary;
-    pathMethod.tags = method.tags;
+    
+    pathMethod.tags = method.tags // [controllerName].concat(method.tags || [])
```
endor: `;` missing. And I think we can remove the comment.

redlock:
So I've handled all the points except the unit test. I'm just not sure where to place it, and I'm not familiar with the test suite/configuration. I'd appreciate if someone more familiar with the tests can add it. Thanks.

Issue #218, opened 2018-04-25
[Koa] global error handler not called


RafaelKr:
Hi,

I want to integrate bugsnag into my project.
For that you have to implement the global Koa error handler: `app.on('error', bugsnag.koaHandler)`
but the handler is never called.

https://github.com/koajs/koa/wiki/Error-Handling

Is there any other way to implement that?

For my local project I found a workaround: https://github.com/RafaelKr/tsoa/commit/5cdb2768db374525466d085032e1276951a40023
But now one of the Koa tests outputs an error:
```
{ BadRequestError: error thrown
    at Object.throw (/home/rafael/coding/Office/Scatch/code/server/tsoa/node_modules/koa/lib/context.js:93:11)
    at /home/rafael/coding/Office/Scatch/code/server/tsoa/tests/fixtures/koa/routes.ts:2021:22
    at <anonymous>
    at process._tickDomainCallback (internal/process/next_tick.js:228:7) message: 'error thrown' }
    ✓ returns error if thrown in controller
```
I tried to fix it with `app.silent = true` or `app.on('error', () => {})` in the file [/
tests/fixtures/koa/server.ts](https://github.com/lukeautry/tsoa/blob/977b535/tests/fixtures/koa/server.ts), but the error message is appearing nevertheless and the error handler is never called.

Edit: I think the error output stems from [/tests/fixtures/koa/server.ts#L33](https://github.com/RafaelKr/tsoa/blob/5cdb2768db374525466d085032e1276951a40023/tests/fixtures/koa/server.ts#L33), somehow I overread that while testing.

In my local app the error handler is called when I do the following:
```
import { Controller, Get, Route } from 'tsoa'

@Route('error')
export class ErrorTestController extends Controller {

  @Get()
  public async getError (): Promise<any> {
    throw {
      status: 500,
      message: 'test'
    }
  }

}
```

lukeautry:
@RafaelKr I think this is fixed as of 2.3.7. Let me know if it's still happening.

PR #276, opened 2018-09-12
[#218] Bubble up validation errors to the global Koa error handler.

aidenwallis:


endor:
Anybody who uses koa any opinions on this? @yolofy @cjam

cjam:
@endor it seems reasonable to me.  I wonder if you may want to put a throw into the authentication checks as well?  see ⬇️ 

https://github.com/lukeautry/tsoa/pull/276/files#diff-f2b44027f6bd1ac07d07eac3a27d7507R92

endor:
@aidenwallis do you still want to extend this with @cjam's suggestion?

aidenwallis:
Awesome, thank you!

x1yy:
this `context.throw(..` at `registerRoutes` makes validation fields error response fail.

this is the response
before:
```
{
	"fields": {
		"proposal.password": {
			"message": "'password' is required"
		}
	},
	"message": "",
	"status": 400,
	"name": "ValidateError"
}
```
after:
```
Anerroroccurredduringtherequest
```

cjam:
Hmm. It almost seems that the error isn't being exposed, which might happen if `error.status` [here](https://github.com/aidenwallis/tsoa/blob/59d200e9c721a8a6d9cc799b64d5d2a9b26eb969/src/routeGeneration/templates/koa.ts#L55) is undefined as it would default the value to `500` in `ctx.throw` (based on https://github.com/koajs/koa/blob/master/docs/api/context.md#ctxthrowstatus-msg-properties).  I've had issues with extending the error object in the past in typescript for some reason.  Are you able to debug and inspect the error object in the `catch` at `registerRoutes` and confirm the status being passed into `ctx.throw`?

Issue #261, opened 2018-08-23
Path parameters in @Route decorator


mushketyk:
Is it possible to provide a path parameter in the `@Route` decorator? I am thinking about something like this:

```typescript
@Route('/folder/{folderId}/file')
class FilesController {

  @Get('{fileId}')
  public async GetFileById(folderId: string, fileId: string): Promise<FileResponse> {
    // Both parameters from the @Route and @Get are available here
    ...
  }
}
```

This would be very handy to avoid some code duplication and would make routes configuration a bit cleaner.

At the moment I get the following error:

```
Generate swagger error.
 Error: @Path('fileId') Can't match in URL: '{fileId}'.
```

endor:
This is not possible yet. The path from the route decorator is not passed on to the parameter validation. It seems like a sensible thing to do though. Would you be interested in implementing this?

AmazingTurtle:
You can use attributes on your parameters as well. This is how I did it and it works:

```
    @Get('getShareAttachment/{attachmentId}')
    public async getShareAttachment(@Request() request: express.Request,
                                    @Path() attachmentId: string,
                                    @Query() shareId: number) {
        ...

    }
```

dgreene1:
To avoid code bloat in tsoa, I nominate that we do not support this feature when it’s very easy for users to add the path parameter as follows:

```ts
@Route('/folder/')
class FilesController {

  @Get('{folderId}/file/{fileId}')
  public async GetFileById(folderId: string, fileId: string): Promise<FileResponse> {
    ...
  }
}
```

Personally, I also find this version to be more readable. That being said, if there’s enough interest we can re-open the issue.

angellandros:
This is more of a philosophical problem. Consider the following example:
`/api/v1/user/12/device/1/`
I have `user/{userId}` at the beginning of everything. Now, how do you expect the `@Route` of `DeviceController` to look like, `'user'` (like all other controllers) or `'user/{userId}/device'`?

dgreene1:
Yes, it’s a philosophical question that doesn’t have a clear answer.

For instance I have worked on many projects where routes to `user/{userId}` and `user` we’re both in the UserController. You seem to suggest that they should be different controller classes.

[...]

PR #280, opened 2018-09-14
Add support for path parameters in @Route decorator #261

kermitology:


endor:
Why two commits with the same title?

tests/fixtures/inversify/managedController.ts
```diff
@@ -5,15 +5,15 @@ import { TestModel } from '../testModel';
 import { ManagedService } from './managedService';
 
 @injectable()
-@Route('ManagedTest')
+@Route('{routeParam}/ManagedTest')
```
endor: why in the inversify controller?

src/metadataGeneration/parameterGenerator.ts
```diff
@@ -143,7 +144,24 @@ export class ParameterGenerator {
...
+          default: getInitializerValue(parameter.initializer, type),
+          description: this.getParameterDescription(parameter),
+          in: 'path',
+          name: pathName,
+          parameterName,
+          required: true,
+          type,
+          validators: getParameterValidators(this.parameter, parameterName),
+        };
+      } else {
```
endor: why does this need to be inside the routePath check if/else block? can't we handle this before the routePath check?

src/metadataGeneration/parameterGenerator.ts
```diff
@@ -143,7 +144,24 @@ export class ParameterGenerator {
       throw new GenerateMetadataError(`@Path('${parameterName}') Can't support '${type.dataType}' type.`);
     }
     if (!this.path.includes(`{${pathName}}`)) {
-      throw new GenerateMetadataError(`@Path('${parameterName}') Can't match in URL: '${this.path}'.`);
+      if (this.routePath !== undefined ) {
```
endor: there's a space too much

tests/fixtures/inversify/managedController.ts
kermitology: I just wanted a simple controller to test with. Could be expanded to other tests.

src/metadataGeneration/parameterGenerator.ts
kermitology: Sorry, my vscode does autoformatting.

src/metadataGeneration/methodGenerator.ts
```diff
@@ -13,6 +13,7 @@ export class MethodGenerator {
 
   constructor(
     private readonly node: ts.MethodDeclaration,
+    private readonly parentPath?: string,
```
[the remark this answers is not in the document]
kermitology: That would make more sense.

endor:
Now there's no unit test anymore? It would be good to have a unit test. If it's easiest in the inversify controller, then let's put it there. Otherwise this might also justify just creating a new small controller, similar to the `headController`.

kermitology:
Sorry, I haven't done a fork merge like this before, I didn't realize this was committing to the pull request. I'm working on a separate test.

tests/fixtures/inversify/routes.ts
```diff
@@ -110,6 +110,7 @@ export function RegisterRoutes(app: any) {
     authenticateMiddleware([{ "MySecurity": [] }]),
     function(request: any, response: any, next: any) {
       const args={
+        routeParam: { "in": "path", "name": "routeParam", "required": true, "dataType": "string" },
```
endor: this shouldn't be in here anymore?

tests/fixtures/routeParameter/routeParameterController.ts
```diff
@@ -0,0 +1,19 @@
+import { inject, injectable } from 'inversify';
```
endor: why use inversify in this test controller?

tests/integration/route-parameter.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+import * as request from 'supertest';
+import { iocContainer } from '../fixtures/routeParameter/ioc';
+import { RouteParameterService } from '../fixtures/routeParameter/routeParameterService';
+import { app } from '../fixtures/routeParameter/server';
+import { TestModel, TestSubModel } from '../fixtures/testModel';
+
+const basePath = '/v1';
+
+describe('RouteParameter Express Server', () => {
+  it('can handle get request with no path argument', () => {
```
endor: why not just add another controller to `fixtures/controllers`, but a whole new "environment" (`fixtures/routeParameter`)?

PR #282, opened 2018-09-17
Change error code in koa to be backwards-compatible

endor:
The unit tests were broken after #276. I made some changes to fix them. This should also resolve your issue @x1yy. Please check and let me know. Is this still okay for you @aidenwallis? Also ping @cjam.

### DIFF
--- a/src/routeGeneration/templates/koa.ts
+++ b/src/routeGeneration/templates/koa.ts
@@ -52,9 +52,8 @@ export function RegisterRoutes(router: any) {
             try {
               validatedArgs = getValidatedArgs(args, context);
             } catch (error) {
-              context.status = error.status || 500;
-              context.body = error;
-              context.throw(context.status, error.message, error);
+              context.status = error.status;
+              context.throw(error.status, JSON.stringify({ fields: error.fields }));
               return next();
             }
 
@@ -88,10 +87,9 @@ export function RegisterRoutes(router: any) {
           const fail = function(error: any) {
               responded++;
               if (responded == security.length && !success) {
-                context.status = error.status || 401;
-                context.body = error;
-                context.throw(context.status, error.message, error);
-                next();
+                  context.status = error.status || 401;
+                  context.throw(context.status, error.message, error);
+                  next();
               }
           }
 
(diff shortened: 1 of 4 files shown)


src/routeGeneration/templates/koa.ts
```diff
@@ -88,10 +86,8 @@ export function RegisterRoutes(router: any) {
           const fail = function(error: any) {
               responded++;
               if (responded == security.length && !success) {
-                context.status = error.status || 401;
```
cjam: Since this failure is due to all of the registered authentication handlers failing, I wonder if it makes sense to keep the default `context.status = error.status || 401`.  Otherwise the authentication middleware will have to make sure to set the status in the error that they throw, since I think Koa will automatically treat those errors as internal.

src/routeGeneration/templates/koa.ts
```diff
@@ -146,9 +142,7 @@ export function RegisterRoutes(router: any) {
             next();
         })
         .catch((error: any) => {
-            context.status = error.status || 500;
-            context.body = error;
-            context.throw(context.status, error.message, error);
+            context.throw(error.status, JSON.stringify({message: error.message}));
```
cjam: I think in some cases (like validation errors), a lot of information about the issue is contained within the error object itself.  I wonder if it makes sense to continue sending the entire error object, as some people may have been exploiting the fact that the error would be within the body of the response.

src/routeGeneration/templates/koa.ts
endor: Are you saying that instead of `context.throw(error.status, `, we should have `context.throw(error.status || 401, `? Is there any advantage to setting `context.status` before using `context.throw`?

cjam: I think so yea, otherwise, the authentication middleware provided will need to know that it must throw an error with a status set.  If one isn't provided, koa will default to 500 which will be misleading from the client perspective.  I think a 401 is a sensible default since, in this particular location, we've tried all of the registered security handlers and none of them succeeded, therefore, returning an unauthorized status seems reasonable.

To your question, from what I can see in the [docs here](https://github.com/koajs/koa/blob/master/docs/api/context.md#ctxthrowstatus-msg-properties) I wouldn't have expected there to be any difference but I also found [this issue](https://github.com/koajs/koa/issues/803) which leads me to think that setting the `context.status` might, unfortunately, be necessary.

endor: should be fixed.

PR #283, opened 2018-09-20
Fix security decorator to also work on controller

endor:


endor:
@jorshali Could you please check if this fixes it for you?

jorshali:
Thanks for looking into this.  Unfortunately, this doesn't seem to resolve the issue.  I'm still getting the following error on the generated routes.ts:
```
Type '{ "name": string; }' is not assignable to type 'Security'.
    Property '"name"' is incompatible with index signature.
      Type 'string' is not assignable to type 'string[]'.
```
In addition, I noticed with the latest I'm getting a compile issue that seems to be related to the merged pull request:  https://github.com/lukeautry/tsoa/pull/267

I was able to fix the issue by reverting the change, but the following compile error is now being returned for the `@Security` annotation:
```
Unable to resolve signature of class decorator when called as an expression.
```
It seems to be related to the new `MethodDecoratorReturn` return type.  Once I change it back to a `Function` return type, the compilation issue is resolved.

endor:
Yes, as the return type is a method decorator, we can't use this decorator on a class anymore. If we make it general `Function`, we'd need to add `ban-types: false` in the tslinting, which would apply to other types than `Function` as well, so not ideal either. I'm not sure what the best approach here is. But it probably requires a separate PR. 

So from your perspective this works and is mergable?

endor:
ping @benjamin-carl

endor:
Okay, and here's a PR, that should fix the decorator return type issue, even though we have to introduce `ban-types: false`: #284.

jorshali:
The latest in master works as expected.  Thanks!

PR #286, opened 2018-09-21
Added missing HTTP method HEAD

bbuhler:


bbuhler:
I rebased my branch, updated the return type and added some unit tests and fixtures.

Note: I did not added fixtures for hapi, because it throws an error during compile saying that head method is not supported.

> tsoa/node_modules/hapi/lib/route.js:41
        Hoek.assert(method !== 'head', 'Method name not allowed:', display);
             ^
AssertionError: Method name not allowed: head /v1/HeadTest

PR #291, opened 2018-09-24
Removed handlebars-helpers (- 200 dependencies)

atomheartother:
Hello,

Handlebars-helpers is a package used in a single file, in a single line of code:
```js
handlebarsHelpers.comparison({
  handlebars,
});
```
This function, in the original source ([here](https://github.com/helpers/handlebars-helpers/blob/master/lib/comparison.js), at `comparison.default`), just checks that its parameters aren't null. It returns an empty string if they're null, so as far as I can tell, since you didn't even check `handlebarsHelpers.comparison`'s return value, this entire call is fairly useless.

Removing `handlebars-helpers` causes a fresh install of `tsoa` to install around 367 packages instead of 554, meaning by removing this one function call we're removing around 190 dependencies from this project. This is good for everyone involved, and there are no drawbacks that I can think of, unless i've misunderstood this function call. Even if I have, the entirety of `handlebars-helpers` is extremely simple functions which we could reimplement in `tsoa` directly and which would, again, make the installation experience much better for users.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,6 @@
   "license": "MIT",
   "dependencies": {
     "handlebars": "^4.0.11",
-    "handlebars-helpers": "^0.9.8",
     "lodash.indexof": "^4.0.5",
     "lodash.map": "^4.6.0",
     "merge": "^1.2.0",
--- a/src/routeGeneration/routeGenerator.ts
+++ b/src/routeGeneration/routeGenerator.ts
@@ -1,6 +1,5 @@
 import * as fs from 'fs';
 import * as handlebars from 'handlebars';
-import * as handlebarsHelpers from 'handlebars-helpers';
 import * as path from 'path';
 import * as tsfmt from 'typescript-formatter';
 import { Tsoa } from '../metadataGeneration/tsoa';
@@ -57,10 +56,6 @@ export class RouteGenerator {
   private buildContent(middlewareTemplate: string, pathTransformer: (path: string) => string) {
     handlebars.registerHelper('json', (context: any) => {
       return JSON.stringify(context);
-    });
-
-    handlebarsHelpers.comparison({
-      handlebars,
     });
 
     const routesTemplate = handlebars.compile(middlewareTemplate, { noEscape: true });


lukeautry:
@cadriel It looks like we added in handlebars-helpers in this PR: https://github.com/lukeautry/tsoa/pull/81/files

What would be the impact of removing it? I don't think the built-in templates use comparisons, but perhaps custom templates would. If that's the case though, the client code can always add whatever helpers they want outside of tsoa. I think it probably makes sense to merge this, but it may break some downstream code that was expecting that helper to be applied.

atomheartother:
> What would be the impact of removing it?
I'm not familiar with your codebase so I can't say with 100% certainty but this PR shows 319 tests passing, 0 failing.

>  it may break some downstream code that was expecting that helper to be applied.
You're right, but we can re-implement this function in `tsoa` and still remove all those useless dependencies. The code of `handlebarsHelpers.comparison` is:

```js
helpers.default = function() {
  for (var i = 0; i < arguments.length - 1; i++) {
    if (arguments[i] != null) return arguments[i];
  }
  return '';
};
```
Unless I have *completely* misunderstood how you guys are using this package, you can just rewrite this function, call it from this spot instead of the package and there should be absolutely no difference with how `tsoa` functioned before. It'll just install faster. Though again I may just completely misunderstand this function call.

cadriel:
I *think* the original intent here was to expose a bunch of handlebar helpers to the custom route templates that users may need.

I believe there is no way for consumers of tsoa to define their own helpers for the route templates - as that's all built in to the `routeGenerator.ts` file.

If you were to remove it with no other changes - you'd limit peoples ability to create complex route templates - which defeats the purpose of exposing that level of customizability.

I think a better approach might be to look into ways for consumers to expose their own helpers. That way you could remove the dep to `handlebars-helpers` whilst still exposing control to consumers of `tsoa`.

endor:
Do I understand correctly that 

```
handlebarsHelpers.comparison({
  handlebars,
});
```

introduces all of the helpers in `comparison`, like `and`, `contains`, etc.? If so, just removing this code does indeed seem to limit users' ability to customize their templates.

We could pass helpers to `buildContent` and register them their manually. They could for example be read in `routeGenerator` from a separate `routesConfig.middlewareTemplateHelpers` file.

cadriel:
Correct yeah. I don't think the intent was to have them *all* in `comparison` though - but that's where they ended up. :)

But yeah, reading them in from some user controllable file sounds good.

endor:
@atomheartother Do you still want to change this PR to allow passing custom helpers?

PR #293, opened 2018-09-29
Added @User decorator

### LINKED ISSUE #274: Suggestion @User decorator
Currently, to get the user in the controllers, you need to add `@Request request` and then extract the user using `request.user`. I propose to add a `@User user` that would do this step for you. 
If this is something that would be accepted I could make a PR sometime this month.

MatthiasKunnen:
This resolves #274.

### DIFF
--- a/README.MD
+++ b/README.MD
@@ -428,22 +428,34 @@ export function koaAuthentication(request: Request, securityName: string, scopes
 }
 ```
 `./controllers/securityController.ts`
-```ts
-import { Get, Route, Security, Response } from 'tsoa';
+```typescript
+import { Get, Route, Security, Response, User } from 'tsoa';
 
 @Route('secure')
 export class SecureController {
     @Response<ErrorResponseModel>('Unexpected error')
     @Security('api_token')
     @Get("UserInfo")
-    public async userInfo(@Request() request: any): Promise<UserResponseModel> {
-        return Promise.resolve(request.user);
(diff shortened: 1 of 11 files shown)


README.MD
```diff
@@ -428,26 +428,38 @@ export function koaAuthentication(request: Request, securityName: string, scopes
...
+import { Get, Route, Security, Response, User } from 'tsoa';
 
 @Route('secure')
 export class SecureController {
     @Response<ErrorResponseModel>('Unexpected error')
     @Security('api_token')
     @Get("UserInfo")
-    public async userInfo(@Request() request: any): Promise<UserResponseModel> {
-        return Promise.resolve(request.user);
+    public async userInfo(@User() user: UserResponseModel | undefined): Promise<UserResponseModel> {
```
endor: Hm, I can't follow. Why is it `UserResponseModel | undefined`?

endor:
@lukeautry So, what do you think?

README.MD
MatthiasKunnen: Yeah, I must admit, it's confusing. I added undefined because I was thinking about my own use case in which I return undefined in the authentication function when authentication is optional. I'll update it.

dgreene1:
It looks like the owner / key maintainer already responded with this thoughts on this: https://github.com/lukeautry/tsoa/issues/274#issuecomment-428735396

I'm closing for housekeeping purposes. In order for tsoa to get increased adoption, it's important that we have a smaller number of open issues and PRs.

Issue #296, opened 2018-10-02
Null values are replaced by undefined in parameter validator


slapersonne:
This is due to param validator using `property.default` if param is `undefined` or `null`, event if ther is no `property.default`, resulting in null values replaced by undefined.

This is an issue for our team because we use `null` values for deletion of attributes in our PATCH endpoints.

Would it be worth considering to return `property.default` only if there is a default value defined in the `PropertySchema`, and return the value otherwise ?

Cerberuser:
I'd ask whether the nullability should be the default behavior, or should we perform check for union types to find out. For now perhaps default is better, since it's far easier and can be performed by the PRs now in process (#295, #297).

slapersonne:
Closed by #297

PR #297, opened 2018-10-02
[#296] Paremeter validation : check if there is a property.default before returning it

slapersonne:
#296 
This change aims to prevent replacing `null` values by `undefined` in parameter validation, by checking there is a default value defined in the `PropertySchema` of the parameter to validate before returning it.

### DIFF
--- a/src/routeGeneration/templateHelpers.ts
+++ b/src/routeGeneration/templateHelpers.ts
@@ -24,8 +24,10 @@ export function ValidateParam(property: TsoaRoute.PropertySchema, value: any, ge
         value,
       };
       return;
-    } else {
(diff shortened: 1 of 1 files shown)


endor:
Could you please add a unit test show-casing this change?

endor:
Is this part of https://github.com/lukeautry/tsoa/pull/295/files?

slapersonne:
It was not aimed to, but you are right, it seems #295 includes this one, so I guess this one can be closed if we #295 is merged with a tiny change (check default value is defined before returning it whatever `property.dataType` is).

slapersonne:
This has been resolved with #297

PR #300, opened 2018-10-09
explicitly set any types for context, next in koa route generation

### LINKED ISSUE #292: Koa + noImplicitAny
I'm not confident I'm not missing anything, but in my project the compiler is choking on the generated koa `routes.ts`, complaining that the `context` and `next` middleware arguments implicitly have type `any`.

I see two solutions, both of which may be bad/wrong:

1. Be explicit about the type of `router` argument here: https://github.com/lukeautry/tsoa/blob/master/src/routeGeneration/templates/koa.ts#L37 - but I assume it's explicitly any in case you're using some router other than koa-router.

2. Be explicit about `context` and `next`, even if it's just explicit `any`.

If someone can point me in the right direction, I'd be happy to put together a PR.

osdiab:
Quick fix to definitively fix #292. Explicitly set context and next types to `any`. Tested by editing the templates in `node_modules/tsoa/dist` with the same change on a stub project of mine, doesn't seem to complain.

### DIFF
--- a/src/routeGeneration/templates/koa.ts
+++ b/src/routeGeneration/templates/koa.ts
@@ -41,7 +41,7 @@ export function RegisterRoutes(router: any) {
             {{#if security.length}}
             authenticateMiddleware({{json security}}),
             {{/if}}
-            async (context, next) => {
+            async (context: any, next: any) => {
             const args = {
                 {{#each parameters}}
                     {{@key}}: {{{json this}}},


osdiab:
Checking in on getting this merged @lukeautry . Thanks for the great library!

dgreene1:
Since this PR needed to be rebased, I just did the tiny PR for ya. So this is closed in favor of https://github.com/lukeautry/tsoa/pull/412

PR #310, opened 2018-11-13
the idea is to allow ignore some of the parameters o the controleer

dkuida:


src/metadataGeneration/methodGenerator.ts
```diff
@@ -1,12 +1,12 @@
 import * as ts from 'typescript';
-import { getDecorators } from './../utils/decoratorUtils';
-import { getJSDocComment, getJSDocDescription, isExistJSDocTag } from './../utils/jsDocUtils';
-import { GenerateMetadataError } from './exceptions';
-import { MetadataGenerator } from './metadataGenerator';
-import { ParameterGenerator } from './parameterGenerator';
-import { getInitializerValue, resolveType } from './resolveType';
-import { getSecurities } from './security';
-import { Tsoa } from './tsoa';
+import {getDecorators} from './../utils/decoratorUtils';
```
endor: Why this change?

src/metadataGeneration/methodGenerator.ts
```diff
@@ -57,7 +56,7 @@ export class MethodGenerator {
   }
 
   private buildParameters() {
-    const parameters = this.node.parameters.map((p) => {
+    const map = this.node.parameters.map((p) => {
```
endor: It might be nicer to just add the `filter` to the end of the `map`, because the intermediate result isn't really a `map`.

src/metadataGeneration/parameterGenerator.ts
```diff
@@ -160,10 +162,14 @@ export class ParameterGenerator {
 
   private getParameterDescription(node: ts.ParameterDeclaration) {
     const symbol = MetadataGenerator.current.typeChecker.getSymbolAtLocation(node.name);
-    if (!symbol) { return undefined; }
```
endor: Why these changes?

src/metadataGeneration/tsoa.ts
```diff
@@ -31,7 +31,7 @@ export namespace Tsoa {
   export interface Parameter {
     parameterName: string;
     description?: string;
-    in: 'query' | 'header' | 'path' | 'formData' | 'body' | 'body-prop' | 'request';
+    in: 'query' | 'header' | 'path' | 'formData' | 'body' | 'body-prop' | 'request' | 'ignored';
```
endor: Should this be `ignore`?

endor:
In general, what is the idea behind having a parameter in the API that's hidden from the specification? It would be good to add an explanation to the decorator part of the README and also add some unit tests showing how such a parameter doesn't show up in the swagger spec.

dkuida:
will be done today-ish - please do not close the PR yet, sorry for the delay

src/metadataGeneration/methodGenerator.ts
dkuida: mistake

src/metadataGeneration/methodGenerator.ts
dkuida: done

lucas-rudd:
I know this PR is rather old. But, I do not see a README for this new functionality. It looks like you're adding a new `ignore` (or, `ignored`) decorator.

@endor , I opened an issue wanting behavior like this which was closed #377

We're using AWS lambda as our Serverless back-end. The handler takes an event with a body, queryParams, pathParams, etc. However, it also can take a Context object, as well as a Callback, both of which get passed to the handler by AWS, but not by the consumer of the API.

In the issue I opened, I thought it would be beneficial to have some sort of @Ignore annotation to ignore the Context and Callback types so that tsoa did not try to resolve them. Otherwise, a lot of additional refactoring would need to be done to get tsoa to work properly with this style of defining AWS lambda function (as well as other Cloud functions such as Google Cloud Functions and Azure Functions, both of which pass in a similar context/callback type which does not need to be exposed in a Swagger doc).

package.json
```diff
@@ -69,10 +68,13 @@
     "@types/yamljs": "^0.2.30",
     "@types/yargs": "^8.0.3",
     "body-parser": "^1.18.3",
+    "moleculer": "^0.13.7",
```
dgreene1: I think that if you could submit this PR without bringing in `moleculer`, `moleculer-web`, and `moleculer-decorates` then I'd be more likely to approve it. But I don't understand why we're bringing in so many additional dependencies.

tests/fixtures/moleculer/get.service.ts
```diff
@@ -0,0 +1,44 @@
...
+  public async getModel(ctx: AuthContext<ChatsActionParams>) {
+    const { withUser } = ctx.params;
+    const fromUser = ctx.meta.user.id;
+    return this._getModel(withUser, fromUser);
+  }
+
+  @Get('/getModel/{withUser}')
+  @Security('ApiKeyAuth')
+  @Method
+  private _getModel(@Path() withUser: string, @Ignore() fromUser: string): Promise<TestClassModel> {
```
dgreene1: I don't see `fromUser` in the path of the `@Get` decorator. So it's ignoring a path variable that isn't on the path. So this test is not set up correctly.

tsconfig.json
```diff
@@ -4,10 +4,11 @@
         "sourceMap": true,
         "noImplicitAny": false,
         "module": "commonjs",
-        "target": "es5",
+        "target": "es2017",
```
dgreene1: I don't see why it's necessary to change the tsconfig target.

src/metadataGeneration/parameterGenerator.ts
```diff
@@ -31,6 +31,8 @@ export class ParameterGenerator {
         return this.getQueryParameter(this.parameter);
       case 'Path':
         return this.getPathParameter(this.parameter);
+      case 'Ignore':
+        return null;
```
dgreene1: I don't think we should change the return type of `Generate` to `Tsoa.Parameter | null`. I think that if you want to ignore a property we have one of two options we can go with:

1) if the goal of the decorator to ignore the validation, then we should rename the decorator to be `IgnoreValidation` and then we should enforce that the `type.dataType` is equal to `any` since that's what the user is basically saying to us (that they don't want any type correctness).
2) if the goal of the decorator is to just not have to send the parameter, then it really begs the question of why does the route even have the query parameter? Anyway, if that's the goal, then we should still return a `Tsoa.Parameter` but it would return this:
```ts
case 'Ignore':
  return {
    //...
    type: {
      dataType: 'any',
    },
    validators: {},
    required: false
}
```

dgreene1:
There are quite a few requested changes. Specifically there should be a discussion about what the intention for the new decorator is. See this comment thread: https://github.com/lukeautry/tsoa/pull/310#discussion_r313068523

Issue #314, opened 2018-11-23
Model validation with optional properties and additionalProperies broken


dpoetzsch:
The following type signature of a request body fails validation at runtime if the optional property `a` is not provided:
```typescript
interface CreationRequestBody {
    a?: string;

    // other keys are allowed as well
    [key: string]: any;
}
```
If I request this endpoint with an empty body I get the following output:
```
{ error: 'ValidateError',
  fields:
   { 'body..a'
      { message: 'No matching model found in additionalProperties to validate a',
        value: 'a' } } }
```

endor:
As far as I know tsoa already allows sending more properties than explicitly defined in that interface. Why add the `[key: string]: any;`?

dpoetzsch:
I did not know about this behavior, thanks for the tip.

Well then, I guess it is more useful for documentation purposes, by explicitly stating that other keys are allowed as well.

tpaulshippy:
We just encountered this issue this morning. Thanks for getting it fixed. Looking forward to the release.

PR #315, opened 2018-11-23
Fix for #314

dpoetzsch:
This fixes issue #314.

The fix was to not check already validated properties against `additionalProperties` again. This assumes that `additionalProperties` is used exclusively for properties defined as `[key: type]: type`.

If `additionalProperties` was meant to set up additional restrictions on *all* properties please let me know, then the fix would be the allow `undefined` as value for type `any`. For my use case this would work as well, but it would break models like the following (which might be unintuitive to users):
```typescript
interface ExampleModel {
    a?: string;
    [key: string]: number;
}
```

### DIFF
--- a/tests/unit/swagger/templateHelpers.spec.ts
+++ b/tests/unit/swagger/templateHelpers.spec.ts
@@ -1,69 +1,88 @@
 import { expect } from 'chai';
 import 'mocha';
-import * as templateHelpers from './../../../src/routeGeneration/templateHelpers';
+import { ValidationService } from './../../../src/routeGeneration/templateHelpers';
 
-describe('templateHelpers', () => {
-
-    afterEach(() => {
-        // clean up global models
-        Object.keys(templateHelpers.models).forEach(k => {
-            delete templateHelpers.models[k];
-        });
-    });
+describe('ValidationService', () => {
 
     describe('Model validate', () => {
         it('should validate a model with declared properties', () => {
-            templateHelpers.models.ExampleModel = {
-                properties: {
-                    a: { dataType: 'string', required: true },
-                },
-            };
+            const v = new ValidationService({
+                ExampleModel: {
+                    properties: {
+                        a: { dataType: 'string', required: true },
+                    },
+                }
+            });
             const error = {};
-            const result = templateHelpers.validateModel('', { a: 's' }, 'ExampleModel', error);
+            const result = v.validateModel('', { a: 's' }, 'ExampleModel', error);
             expect(Object.keys(error)).to.be.empty;
             expect(result).to.eql({ a: 's' });
         });
 
         it('should not require optional properties', () => {
-            templateHelpers.models.ExampleModel = {
-                properties: {
-                    a: { dataType: 'string' },
-                },
-            };
+            const v = new ValidationService({
+                ExampleModel: {
+                    properties: {
+                        a: { dataType: 'string' },
+                    },
+                }
+            });
             const error = {};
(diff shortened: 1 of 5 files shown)


tests/unit/swagger/templateHelpers.spec.ts
```diff
@@ -4,6 +4,62 @@ import * as templateHelpers from './../../../src/routeGeneration/templateHelpers
 
 describe('templateHelpers', () => {
 
+    afterEach(() => {
+        // clean up global models
+        Object.keys(templateHelpers.models).forEach(k => {
```
endor: why not simple reset it to `{}`?

tests/unit/swagger/templateHelpers.spec.ts
```diff
@@ -4,6 +4,62 @@ import * as templateHelpers from './../../../src/routeGeneration/templateHelpers
 
 describe('templateHelpers', () => {
 
+    afterEach(() => {
```
endor: hm.. if we were to call `ValidateParam` in the unit tests instead of directly calling `validateModel`, wouldn't that save is this `afterEach` and also the exports above?

endor:
Can you explain what you mean when you say that `additionalProperties` could be meant to set up additional restrictions on all properties? The `ExampleModel` is illegal TypeScript, right?

dpoetzsch:
Actually your are right, it is illegal typescript, I did not see that, thanks.

What I meant is probably made more clear by the following test case:

```typescript
it('should validate all model properties against constraint defined in additional properties', () => {
    templateHelpers.models.ExampleModel = {
        additionalProperties: {
            dataType: 'integer',
            validators: { minimum: { value: 10 } }
        },
        properties: {
            a: { dataType: 'integer' },
        },
    };
    const error = {};
    templateHelpers.validateModel('', { a: 9 }, 'ExampleModel', error);
    expect(Object.keys(error)).not.to.be.empty;
});
```
Is this the expected outcome? If yes, my fix is not correct, yet.

tests/unit/swagger/templateHelpers.spec.ts
dpoetzsch: This is not possible in typescript because exported globals are read-only from outside the module.

I agree this is a bit hacky.

I think a cleaner approach would be to avoid the global and wrap all the validation functions into a service class that takes the model as constructor argument. But I did not want to change so much in your code.

tests/unit/swagger/templateHelpers.spec.ts
dpoetzsch: `models` is accessed by `validateArray` as well, which is directly called from the tests. This is where such a dependency might occur.

I will just refactor this, removing the shared state, no problem.

endor:
Ah, good question. I think it's good to be explicit, so I'd say the above test case is not expected to work.

tests/unit/swagger/templateHelpers.spec.ts
endor: Ah, you're right. I'm not sure those "sub"-validators should be exported. But thanks for fixing the shared state.

dpoetzsch: Yeah, I'm always unsure as well. But I personally think in this case it makes sense to make them available so they can be tested individually because validation is quite complex for each case..

dpoetzsch:
I removed the global variable now and added the test above but with inverted expectation (expecting no error now). For backwards compatibility I also added a ValidateParam helper function that wraps instance creation.

 All tests work now.

PR #317, opened 2018-11-27
Better handling of array query parameters

### LINKED ISSUE #316: collectionFormat is not specified for array query parameters in the generated swagger
This is because the check on parameter type in `specGenerator`.`buildParameter` for arrays is wrong (`parameter.type === 'array'` instead of `parameterType.type === 'array'`)

slapersonne:
This aims to better handle the array query parameters, with a fix for `collectionFormat` property generation in swagger.
CollectionFormat could then later be specified with a tag, but I kept the default value for this fix.

Close #316

### DIFF
--- a/src/metadataGeneration/parameterGenerator.ts
+++ b/src/metadataGeneration/parameterGenerator.ts
@@ -110,27 +110,35 @@ export class ParameterGenerator {
     const parameterName = (parameter.name as ts.Identifier).text;
     const type = this.getValidatedType(parameter, false);
 
-    if (type.dataType === 'array') {
-      const arrayType = type as Tsoa.ArrayType;
-
-      if (!this.supportPathDataType(arrayType.elementType)) {
-        throw new GenerateMetadataError(`@Query('${parameterName}') Can't support array '${arrayType.elementType.dataType}' type.`);
-      }
-    } else {
-      if (!this.supportPathDataType(type)) {
-        throw new GenerateMetadataError(`@Query('${parameterName}') Can't support '${type.dataType}' type.`);
-      }
-    }
-
-    return {
+    const commonProperties = {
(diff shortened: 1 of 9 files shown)


src/metadataGeneration/tsoa.ts
```diff
@@ -39,6 +39,11 @@ export namespace Tsoa {
     validators: Validators;
   }
 
+  export interface ArrayParameter extends Parameter {
```
endor: Why introduce this?

src/metadataGeneration/parameterGenerator.ts
```diff
@@ -110,27 +110,35 @@ export class ParameterGenerator {
     const parameterName = (parameter.name as ts.Identifier).text;
     const type = this.getValidatedType(parameter, false);
 
+    const commonProperties = {
+      default: getInitializerValue(parameter.initializer, type),
+      description: this.getParameterDescription(parameter),
+      in: 'query' as 'query',
```
endor: What does this do?

slapersonne: This because `in` property is typed as `'query' | 'header' | 'path' | 'formData' | 'body' | 'body-prop' | 'request'`. With just `in: 'query'` for the definition, it is typed as `string` by default, so there is an incompatible type error when returning the parameter.
An alternative would be to use `as Tsoa.Parameter` for the return.

src/metadataGeneration/tsoa.ts
slapersonne: To avoid defining the `collectionFormat` in `Parameter`, because it is specific to array parameters. There was already an `ArrayType` for array-specific types so I thought it would be fair to also have an `ArrayParameter`.

endor:
Thanks. This looks good.

PR #332, opened 2019-02-10
Numeric enums

### LINKED ISSUE #331: Numeric Enums Still Don't Work
Does this fix actually work? I'm still getting this issue. Currently on `tsoa#2.3.8`

Edit: Looks like it never got in even those this PR got merged. What happened?

_Originally posted by @drdgvhbh in https://github.com/lukeautry/tsoa/pull/106#issuecomment-462087394_

drdgvhbh:
Fixes #331


tests/unit/swagger/templateHelpers.spec.ts
```diff
@@ -172,34 +172,60 @@ describe('ValidationService', () => {
     });
 
     describe('Enum validate', () => {
+        const name = 'name';
+        type Enumeration = string[] | number[];
+        interface FieldErrors { [key: string]: any; }
+        let validateEnum: (value: any, enumeration: Enumeration, fieldErrors?: FieldErrors) => any;
+
+        beforeEach(() => {
+          const validationService = new ValidationService({});
```
endor: Personally I don't really like having logic in the tests. I'd rather have a bit of repetition but the meaning obvious.

src/swagger/specGenerator.ts
```diff
@@ -69,10 +69,13 @@ export class SpecGenerator {
 
       // Enum definition
       if (referenceType.enums) {
+        const isNumericEnumArray = !(referenceType.enums as any[]).some(isNaN);
+        const type = isNumericEnumArray ? 'integer' : 'string';
```
endor: Shouldn't this be `number` instead of `integer`?

src/metadataGeneration/typeResolver.ts
```diff
@@ -186,7 +186,12 @@ export class TypeResolver {
 
     const enumDeclaration = enumNodes[0] as ts.EnumDeclaration;
 
+    const typeChecker = this.current.typeChecker;
     function getEnumValue(member: any) {
+      const constantValue = typeChecker.getConstantValue(member);
```
endor: Hm, doesn't this return `string | number | undefined`? So shouldn't we check for `undefined` below?

src/metadataGeneration/typeResolver.ts
```diff
@@ -199,7 +204,11 @@ export class TypeResolver {
 
     if (extractEnum) {
       const enums = enumDeclaration.members.map((member: any, index) => {
-        return getEnumValue(member) || String(index);
+        const enumValue = getEnumValue(member);
+        if (enumValue !== 0 && !enumValue) {
```
endor: Can't we check for `undefined` here?

tests/unit/swagger/templateHelpers.spec.ts
```diff
@@ -172,34 +172,60 @@ describe('ValidationService', () => {
...
+
+        it('should not discriminate between numeric and string enum values', () => {
+          function run(
+            val: string | number,
+            enumeration: string[] | number[],
+          ) {
+            const result = validateEnum(val, enumeration);
+            expect(result).to.equal(val);
+          }
+          run('1', [0, 1]);
```
endor: I have a bit of trouble understanding this. Is the idea that in your app you could have a numeric enum defined, but incoming values in query parameters can be strings and so that should be valid?

src/metadataGeneration/typeResolver.ts
dgreene1: He’s saying that it can’t be null, undefined or an empty string but that it can be 0. If he just checked for not undefined then he would lose the empty string check and also 0 would be in appropriately flagged.

tests/unit/swagger/templateHelpers.spec.ts
dgreene1: @drdgvhbh this PR ultimately looks great. I do agree with @endor here that the intent of your test isn’t very clear. I just mean that it would be wonderful if you would add a comment into the test explaining to future contributors why it’s okay that a stringified number satisfies a numeric enum check. Because I’m a bit uncertain if it is okay. So any clarification that you can add to the code would be helpful.

tests/unit/swagger/templateHelpers.spec.ts
```diff
@@ -172,34 +172,60 @@ describe('ValidationService', () => {
...
-            const name = 'name';
             const value = 'SAY';
-            const result = new ValidationService({}).validateEnum(name, value, error, ['HELLO', 'HI'] as any);
+            const result = validateEnum(value, ['HELLO', 'HI'], error);
             expect(result).to.equal(undefined);
-            expect(error[name].message).to.equal(`should be one of the following; ['HELLO', 'HI']`);
+            expect(error[name].message).to.equal(
+              `should be one of the following; ['HELLO', 'HI']`);
+        });
+
```
dgreene1: @drdgvhbh We need a unit test that verifies that the following would be an error:

Given the following enum, the validation service should return an error if the value provided was a non-numeric string like “foo”.

```ts
export enum MyIntEnum {
  one = 1
  two = 2
}
```

dgreene1:
@drdgvhbh PR looks mostly there. If you could just make the two unit tests corrections and merge master back into the branch then we could accept this PR. Thank you. :)

dgreene1:
@drdgvhbh any interest wrapping this PR up?

> PR looks mostly there. If you could just make the two unit tests corrections and merge master back into the branch then we could accept this PR. Thank you. :)

drdgvhbh:
I'll have a look this week ;)

dgreene1:
@drdgvhbh why did you think it was okay to remove our CI tests (https://github.com/lukeautry/tsoa/pull/332/commits/fab74522f1e4f12dc134d92ca56163930f0b150e) by deleting the github workflows?

If you add that back in then we can consider merging this PR.

src/swagger/specGenerator.ts
```diff
@@ -11,6 +11,44 @@ export abstract class SpecGenerator {
     return this.getSwaggerType(type);
   }
 
+  private buildDefinitions() {
```
dgreene1: I believe that you meant to make this change to the child (`specGenerator2.ts`) and not to the parent (`specGenerator.ts`).

So please remove `specGenerator.ts`.`buildDefinition` and make those changes to the two children:
* `specGenerator2.ts`.`buildDefinition`
* `specGenerator3.ts`.`buildComponents`
  * Side note: judging by the fact nothing in `specGenerator3` calls `buildDefinitions` that tells me that the PR as it currently stands does not actually allow numerical enums for OpenAPI 3.0. Can you please add tests for the swagger definitions?
    * test file for OpenAPI 2.0 (i.e. Swagger) definitions: `tests\unit\swagger\definitionsGeneration\definitions.spec.ts`
    * test file for OpenAPI 3.0 definitions: `tests\unit\swagger\schemaDetails3.spec.ts`

drdgvhbh:
> @drdgvhbh why did you think it was okay to remove our CI tests ([fab7452](https://github.com/lukeautry/tsoa/commit/fab74522f1e4f12dc134d92ca56163930f0b150e)) by deleting the github workflows?
> 
> If you add that back in then we can consider merging this PR.

I havent finished sorry the PR yet sorry. It wasn't letting me push unless i deleted because of "stale issues". I'll add it back in later

dgreene1:
I reopened the original issue for you. But that shouldn’t have affected it. Also, github will let you push even if the actions say that it fails. The actions only prevent it from being able to merged into master.

Anyway, take your time and thank you for working on this. :)

tests/unit/swagger/templateHelpers.spec.ts
drdgvhbh: I'm not sure if its okay either. It does preserve existing functionality though.

Currently an enum `[1, 2, 3]` becomes `['1', '2', '3']` in swagger. The change would make it the same as `[1, 2, 3]`, but you wouldn't screw over the people were inputting `'1'`, `'2'`, `'3'` in their http requests. 

I'd say turn it into a separate issue that gets put in the the next minor release.

drdgvhbh:
Should be good now. The github thing was apparently a Gitkraken issue :P.

tests/unit/swagger/templateHelpers.spec.ts
```diff
@@ -264,6 +264,76 @@ describe('ValidationService', () => {
...
+      const name = 'name';
+      const value = 'SAY';
+      const error: any = {};
+      const enumeration: Enumeration = ['HELLO', 'HI'];
+      const result = new ValidationService({}).validateEnum(name, value, error, enumeration);
+      expect(result).to.equal(undefined);
+      expect(error[name].message).to.equal(`should be one of the following; ['HELLO', 'HI']`);
+    });
+
+    it('accepts a string value of a numeric enum', () => {
```
dgreene1: Can you please make this a negative test case?

In other words, can you make it so that submitting `1` is a failure if the enum contains `”1”`?

I’m asking for this because `”1”.toLowerCase()` is a function but `1.toLowerCase()` is not. So if we let an string replace a number (and vica-versa) then we put tsoa users at risk of a runtime error.

tests/unit/swagger/templateHelpers.spec.ts
```diff
@@ -264,6 +264,76 @@ describe('ValidationService', () => {
...
+      const name = 'name';
+      const value = '1';
+      const error: any = {};
+      const enumeration: Enumeration = [0, 1];
+      const result = new ValidationService({}).validateEnum(name, value, error, enumeration);
+      expect(result).to.equal(value);
+    });
+
+    it('accepts a numeric value of a string-numeric enum', () => {
```
dgreene1: Same for this one as above (https://github.com/lukeautry/tsoa/pull/332/files/fa05ae3c9aeefa741c2643909b5b07655b0fe9ed#discussion_r327400448). This case should be a negative test case. We can’t allow a number to swap as a string.

drdgvhbh:
Not sure if I understand fully, but I think this is what you meant?

dgreene1:
Yup, commit https://github.com/lukeautry/tsoa/commit/42fa49e8fbb2b33fb711d50fbb00f9ae75d5f13c added the tests I was looking for. :)

PR #335, opened 2019-02-21
koa error handling

technicallyfeasible:
If one of the controller methods throws an error then this error was caught and swallowed by the authentication "fail" handler.
The promise handler now defaults to a status of 500 in case the original error did not have a status property.

Not sure whether this change breaks any current behavior that people rely on. Originally the router would instead set a status of 404 "Not found" if the controller method threw an error, now any koa handlers higher up in the chain will capture errors so normal error handling can kick in.


technicallyfeasible:
Bump.
This is an issue that I currently have to fix manually in our production code. Would it be possible for someone to look at this and merge it please?

Or if there is a reason this is not merged yet please let me know and I'll try to improve it.

lukeautry:
@technicallyfeasible Released in `3.9.81`.

PR #345, opened 2019-04-11
Controller inheritance

aleangel1212:
Allow for controllers to extend each other via TypeScript `extends`, inheriting all decorated methods from ancestors in the inheritance chain.

**controllers/baseController.ts**
```
export class BaseController extends SuperBaseController{
  @Get('Get')
  public async getMethod(): Promise<Model> {
    return
  }
}
```

**controllers/userController.ts**
```
export class UserController extends BaseController {}
```

UserController will have the `Get` method

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "tsoa",
-  "version": "2.3.7",
+  "version": "2.3.81",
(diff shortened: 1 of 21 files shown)


asadhaider13:
Any updates on this?

PR #347, opened 2019-04-16
Custom Attributes

### LINKED ISSUE #294: Adding special attributes to swagger documentation
Hey there

We need to add a custom (not documented) attribute to the swagger file that gets generated. Is this at all possible now? It wasn't quite sure if this was possible through jsdoc, and if so, how.
Makes sense if it's not possible, but it's kind of a deal breaker in terms of us adopting.

Cheers
And I appreciate all the hard work you've put into this so far.

aleangel1212:
Fixes #294 

Allows controller methods to define 'CustomAttribute' decorators that define additional attributes to be added to the swagger file.

i.e.
```typescript
@Route('User')
export class UserController {
  @CustomAttribute('att1', 'val1')
  @CustomAttribute('att2', { key1: 'att2val1', key2: 'att2val2' })
  @CustomAttribute('att3', ['att3val1', 'att3val2', 'att3val3'])
  @CustomAttribute('att4', [{ key1: 'att4val1' }, { key2: 'att4val2' }])
  @Get()
  public async get(): Promise<User> {
    return {} as User
  }
}
```

Generates:
```yaml
/User:
        get:
            operationId: Get
            produces:
                - application/json
            responses:
                '200':
                    description: Ok
                    schema:
                        $ref: '#/definitions/User'
            security: []
            parameters: []
            att1: val1
            att2:
                key1: att2val1
                key2: att2val2
            att3:
                - att3val1
                - att3val2
                - att3val3
            att4:
                -
                    key1: att4val1
                -
                    key2: att4val2
```


endor:
To me this looks good. What do you think @lukeautry @dgreene1 @RckMrkr?

RckMrkr:
I don't feel close enough to the code to comment sufficiently on that part, but from the usage perspective, it looks just as I imagined it.

endor:
I'll merge this after the OpenAPI 3.0.0 branch has been merged.

endor:
@aleangel1212 Can you solve the conflicts still?

dgreene1:
@aleangel1212 I'm confused by the swagger.yaml that this outputs because "att1" is not a valid parameter for swagger. If your run the output through https://editor.swagger.io/ you'll find that it's not valid. I don't think we should allow anything to be produced that isn't a valid swagger document. Perhaps @RckMrkr @pavlrd can elaborate on their original use case?

src/metadataGeneration/customAttribute.ts
```diff
@@ -0,0 +1,63 @@
...
+    if (decoratorKeyArg.kind !== ts.SyntaxKind.StringLiteral) {
+      throw new Error('First argument of Custom Attribute must be the attribute key as a string');
+    }
+
+    const attributeKey = getArgumentValue(decoratorKeyArg);
+
+    if (!decoratorValueArg) {
+      throw new Error(`Custom Attribute '${attributeKey}' must contain a value`);
+    }
+
```
dgreene1: I'd like to see a check like the following:
```ts
if (!attributeKey.lastIndexOf("x-", 0) === 0) {
  throw new Error('Custom attributes must begin with "x-" to be valid. Please see the following link for more information: https://swagger.io/docs/specification/openapi-extensions/');
}
```

I'd also need to see a unit test that verifies this error is thrown in order to accept this PR.

tindleaj:
Added a check for the proper 'x-' prefix, it'll now throw if it is invalid OAS2. Also added unit testing around that.

dgreene1:
I have a little more reviewing to do, but can someone add to the readme with some explanation of example use cases?

Additionally, I’m unclear if the model validation should be using these custom properties for anything?

That’s why I’d like to know more about the real life use cases for these so we can better understand the implications to the larger codebase.

src/swagger/specGenerator.ts
```diff
@@ -6,6 +6,253 @@ import { Swagger } from './swagger';
 export class SpecGenerator {
   constructor(protected readonly metadata: Tsoa.Metadata, protected readonly config: SwaggerConfig) { }
 
+  public GetSpec() {
+    let spec: Swagger.Spec = {
```
dgreene1: @WoH would you be willing to review this PR? I think you’re most familiar with the spec generator code at this point.

dgreene1: @aleangel1212 I noticed that a lot of this code is a copy and paste of the specGenerator 2 code. I think there has to be a DRYer way to do this. And also, shouldn’t the custom attributes get different validations depending on if the consumer has selected Swagger 2 vs OpenAPI 3.0? I ask because I believe  (and @endor could confirm) that the SpecGenerator2 and SpecGenerator3 files exist for that very reason.

WoH: This almost positive this code was introduced due to a faulty merge. I'd take a closer look at the whole PR if there is anyone ready to pick it up.

src/metadataGeneration/customAttribute.ts
```diff
@@ -0,0 +1,67 @@
+import * as ts from 'typescript';
+import { getInitializerValue } from './initializer-value';
+import { Tsoa } from './tsoa';
+
+const getArgumentValue = (argument: ts.Expression) => {
```
WoH: We could improve the structure here. If you don't switch on the kind but list 3 if blocks, each one will ensure type safety (by taking advantage of the type guards of the compiler API). 

I.e.:
```ts
if (ts.isStringLiteral(argument)) {
  return argument.text; // this is valid without casting now
}
```

src/metadataGeneration/customAttribute.ts
```diff
@@ -0,0 +1,67 @@
...
+  switch (argument.kind as ts.SyntaxKind) {
+    case ts.SyntaxKind.StringLiteral: {
+      const argumentValue = (argument as any).text;
+
+      return argumentValue;
+    }
+
+    case ts.SyntaxKind.ObjectLiteralExpression: {
```
WoH: This block contains a big amount of assumptions disguised by the `any` cast.
Property is `ObjectLiteralElementLike` which can be `PropertyAssignment | ShorthandPropertyAssignment | SpreadAssignment | MethodDeclaration | AccessorDeclaration`

This piece of code is only handling PropertyAssignment, so it should throw on any thing else.

src/metadataGeneration/customAttribute.ts
```diff
@@ -0,0 +1,67 @@
...
+      const argumentValueProperties = (argument as any).properties;
+
+      const argumentValue = {};
+
+      for (const property of argumentValueProperties) {
+        const name = property.name.text;
```
WoH: `property.name` may be `undefined`, should be checked.
If it is defined, it can be `Identifier | StringLiteral | NumericLiteral | ComputedPropertyName`.
By using getText(), all of these can be handled correctly, not just `Identifier`

src/metadataGeneration/customAttribute.ts
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    default: {
+      throw new Error('Custom Attribute values must be of type: string, object, or array');
+    }
+  }
+};
+
+export function getCustomAttributes(decorators: ts.Identifier[]): Tsoa.CustomAttribute[] {
+  const customAttributes: Tsoa.CustomAttribute[] = decorators.map((customAttributeDecorator: ts.Identifier) => {
```
WoH:
```suggestion
  const customAttributes: Tsoa.CustomAttribute[] = decorators.map(customAttributeDecorator => {
```

src/metadataGeneration/customAttribute.ts
```diff
@@ -0,0 +1,67 @@
...
+    const expression = customAttributeDecorator.parent as ts.CallExpression;
```
WoH: Questionable assumption.

src/swagger/swagger.ts
```diff
@@ -179,6 +179,8 @@ export namespace Swagger {
     schemes?: Protocol[];
     deprecated?: boolean;
     security?: Security[];
+    // Used to apply custom attributes to paths
+    [key: string]: any;
```
WoH:
```suggestion
    [key: string]: unknown;
```

src/swagger/specGenerator.ts
```diff
@@ -6,6 +6,253 @@ import { Swagger } from './swagger';
...
+    const bodyPropParameter = this.buildBodyPropParameter(controllerName, method);
+    if (bodyPropParameter) {
+      pathMethod.parameters.push(bodyPropParameter);
+    }
+    if (pathMethod.parameters.filter((p: Swagger.BaseParameter) => p.in === 'body').length > 1) {
+      throw new Error('Only one body parameter allowed per controller method.');
+    }
+
+    // Apply custom attributes
+    method.customAttributes.forEach((customAttr) => pathMethod[customAttr.key] = customAttr.value);
```
WoH: This line should be the one moved, everything else added in this file should be removed.

tests/fixtures/controllers/invalidCustomMethodAttributeController.ts
```diff
@@ -0,0 +1,48 @@
+import {
+  Controller,
+  CustomAttribute,
+  CustomMethodAttribute,
```
WoH: Same issues as in the `customMethodAttributeController.ts`

tests/unit/swagger/schemaDetails.spec.ts
```diff
@@ -39,3 +39,30 @@ describe('Schema details generation', () => {
 
   it('should set API license if provided', () => expect(licenseName).to.equal(getDefaultOptions().license));
 });
+
+describe('Custom Attribute schema generation', () => {
+  const metadata = new MetadataGenerator('./tests/fixtures/controllers/methodController').Generate();
+  const spec = new SpecGenerator(metadata, getDefaultOptions()).GetSpec();
```
WoH:
```suggestion
  const spec = new SpecGenerator2(metadata, getDefaultOptions()).GetSpec();
```
Also, add tests for SG3 please.

src/metadataGeneration/customAttribute.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    const [decoratorKeyArg, decoratorValueArg] = expression.arguments;
```
WoH: If you have the suggested check there, this is safe to do without any casting: 
```suggestion
    const [decoratorKeyArg, decoratorValueArg] = customAttributeDecorator.parent.arguments;
```

src/metadataGeneration/customAttribute.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    if (decoratorKeyArg.kind !== ts.SyntaxKind.StringLiteral) {
+      throw new Error('First argument of Custom Attribute must be the attribute key as a string');
+    }
+
+    const attributeKey = getArgumentValue(decoratorKeyArg);
```
WoH: I think it would be easier to read is this would just check for 
```ts
if (!ts.isStringLiteral(decoratorKeyArg)) {
  throw new Error('The first argument of @CustomAttribute must be the attribute key as a string');
}
```
and subsequently just accessed `decoratorKeyArg.text`.
The call to getArgumentValue is unnecessary for the `decoratorKeyArg`.

dgreene1:
@aleangel1212 can you make the requested changes?

zlumer:
Any chance this will be merged? I guess there's a workaround to skip CLI and inject custom attributes manually as a build step, but it's pretty suboptimal.

dgreene1:
@zlumer it seems that the original author of the PR @aleangel1212 has stopped working on this. If you would like to submit new commits to the branch that fix the requested changes to the PR, we would approve it. But there are a couple of small but important requested changes. Most importantly that custom attributes must start with an “x-“ because the spec dictates that.

PR #348, opened 2019-04-16
Controller Inheritance and Generic Resolution

aleangel1212:
Allow for controllers to extend each other via TypeScript `extends`, inheriting all decorated methods from ancestors in the inheritance chain.

**controllers/superBaseController.ts**
```typescript
export class SuperBaseController<T>{
  @Post()
  public async postMethod(): Promise<T> { return }
}
```

**controllers/baseController.ts**
```typescript
export class BaseController<T> extends SuperBaseController<T>{
  @Get()
  public async getMethod(): Promise<T> { return }
}
```

**controllers/userController.ts**
```typescript
export class UserController extends BaseController<Model> {}
```

UserController will have the both the 'Post' and 'Get' method

### DIFF
--- a/src/metadataGeneration/tsoa.ts
+++ b/src/metadataGeneration/tsoa.ts
@@ -5,6 +5,7 @@ export namespace Tsoa {
   }
 
   export interface Controller {
+    inheritanceList: string[];
     location: string;
     methods: Method[];
     name: string;
--- a/src/metadataGeneration/tsoa.ts
+++ b/src/metadataGeneration/tsoa.ts
@@ -5,7 +5,6 @@ export namespace Tsoa {
   }
 
   export interface Controller {
-    inheritanceList: string[];
     location: string;
     methods: Method[];
     name: string;
@@ -97,4 +96,6 @@ export namespace Tsoa {
   export interface ReferenceTypeMap {
     [refName: string]: Tsoa.ReferenceType;
   }
+
+  export type GenericTypeMap = Map<string, Map<string, string>>
 }
--- a/tests/fixtures/controllers/duplicateMethodController.ts
+++ b/tests/fixtures/controllers/duplicateMethodController.ts
@@ -1,14 +1,15 @@
 import {
   Get,
   Patch,
+  Put,
   Route,
 } from '../../../src';
 import { ModelService } from '../services/modelService';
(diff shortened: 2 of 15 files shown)


src/metadataGeneration/tsoa.ts
```diff
@@ -96,4 +96,6 @@ export namespace Tsoa {
   export interface ReferenceTypeMap {
     [refName: string]: Tsoa.ReferenceType;
   }
+
+  export type GenericTypeMap = Map<string, Map<string, string>>
```
endor: Could it be more readable if we introduced type alias like `BaseClassName` or `ResolvedModel`?

tests/fixtures/controllers/duplicateMethodController.ts
```diff
@@ -0,0 +1,27 @@
...
+  Get,
+  Patch,
+  Put,
+  Route,
+} from '../../../src';
+import { ModelService } from '../services/modelService';
+import { TestModel } from '../testModel';
+import { BaseController } from './baseController';
+
+@Route('DuplicateMethodTest')
```
endor: I have some trouble following why methods here are duplicate but not for the `InheritanceMethodController` below. I also haven't yet understood if it's allowed to override a method defined in a base class or not. Maybe some documentation could help.

HoldYourWaffle:
https://github.com/lukeautry/tsoa/pull/348/commits/df8ef765bedfba5ec36bc9e0c7757075ab96228f is no longer needed due to https://github.com/lukeautry/tsoa/commit/15eb96f4d2358d19e380818b2b1ee664d4c6740b

moshewe:
I'd like to revive this PR... How do we go about it?

src/metadataGeneration/tsoa.ts
moshewe: `ResolvedModel` fits much better, we don't want to imply at any point that a certain class in the chain of extensions is the "base".

tests/fixtures/controllers/duplicateMethodController.ts
moshewe: The purpose of this PR seems to be to reflect typescript class inheritance into OpenAPI spec, so I think you're supposed to be allowed to override methods.

PR #359, opened 2019-06-07
How to correctly generate open api 3 document 358

0v3rst33r:
### API and Spec versions in separate configs

A few things to note:

- The `specVersion` is a number format, so only single (`Major`) spec versions supported.  This might not be ideal?
- Had to expose the `validateSwaggerConfig` method in the `cli.ts` file in order to add unit tests for the configs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 dist
 node_modules
 routes.ts
+.idea/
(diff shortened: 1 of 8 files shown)


0v3rst33r:
@lukeautry thanks for this merge.  Are you going to do a `2.4.1` release (and also publish to `npm`) for this?

PR #361, opened 2019-06-10
Add routesFileName option to change the filename of generated router files

### LINKED ISSUE #360: Add option to change output filename
The `routes.ts` output filename is currently hard coded:
https://github.com/lukeautry/tsoa/blob/7c9dbd0798340c04721dfdca3f2fd31d8be3aca8/src/routeGeneration/routeGenerator.ts#L26

I'd love to have an option to change this filename. I was going to make a PR for this, but the fact that such a trivial thing hasn't been implemented yet makes me wonder if there's an actual reason for it.

HoldYourWaffle:
Fixes #360.
I'm not 100% happy with this implementation but I'm not sure there's a better way that preserves backwards compatibility without a redundant `routesFile` option.
I don't have any experience with this codebase however so I could very well have missed a very obvious yet clear solution 😉


HoldYourWaffle:
Seems like this somehow breaks the tests, but not consistently. No idea how that happens (especially the inconsistency) but I'll hopefully be able to look into it tomorrow.

HoldYourWaffle:
Fixed

lukeautry:
Would it make sense to write a test for some part of this?

HoldYourWaffle:
I was thinking about that too, I think it does. I'm not sure what kind of test I could write for this though.

src/routeGeneration/routeGenerator.ts
```diff
@@ -122,8 +122,13 @@ export class RouteGenerator {
   }
 
   private getRelativeImportPath(fileLocation: string) {
-    fileLocation = fileLocation.replace('.ts', '');
-    return `./${path.relative(this.options.routesDir, fileLocation).replace(/\\/g, '/')}`;
+    const outputDirectory =
+      path.extname(this.options.routesDir) === '' ? // routes dir is a directory
```
dgreene1: @HoldYourWaffle, I agree that this needs unit tests. If you make the `buildContent` function a public function, you could write a test that checks if the handlebars output string contains the correct authenticationModule. Like did it find the path?
```
{{#if authenticationModule}}
import { koaAuthentication } from '{{authenticationModule}}';
{{/if}}
```

src/routeGeneration/routeGenerator.ts
```diff
@@ -122,8 +122,13 @@ export class RouteGenerator {
...
+      this.options.routesDir :
```
dgreene1: I know it's a little late to discuss options, but @HoldYourWaffle I think it would be better if `routesDir` had to be a dir and we added a new `routeFilePath` config option that had to be a `.ts` file. I find it confusing as a user to see that `routeDir` could be a file. To put it another way, I think apples should stay apples, and oranges should stay oranges.

src/routeGeneration/routeGenerator.ts
endor: It should also be possible to write a test that does something similar to `generate-routes.ts`.. so instantiate the `RouteGenerator` with specific options (`routesDir` as file and `routesDir` as dir) and then check the resulting file. This way the function could stay private and we would additionally test the public interface.

src/routeGeneration/routeGenerator.ts
HoldYourWaffle: I agree, this was a bit of a hacked-together and rushed implementation (as mentioned by my original post).

If you feel like adding a (past-me would say "redundant") `routeFile`-specific option I'd be happy to implement that. I think there are two paths we could take for that:
1. We add a separate `routeFileName` option that changes the `routes.ts` part (my preferred option).
2. We add a `routeFilePath` option as a sort-of "sibling" (English is not my first language) to `routesDir` that serves the same purpose but without the `routes.ts` suffix. What would happen if both `routesDir` and `routeFilePath` are specified though?

dgreene1: I suppose I’d be fine either option. If you selected option 2, you would just add validation that `routesDir` and `routesFilePath` were not both present. The validation would throw:

```
if(config.routesDir && config.routesFilePath) {
  throw new Error(`You can only provide routesDir or routesFilePath in your config.`);
}
```

HoldYourWaffle: I implemented option 1 since it's a lot simpeler and intuitive to use (at least in my opinion)

HoldYourWaffle:
I'm not sure I implemented the 'default value' of the new option correctly, it feels weird to manually add it to every test invocation.

I still have to add tests, I'll see if I can put something together tomorrow.

src/config.ts
```diff
@@ -132,6 +132,11 @@ export interface RoutesConfig {
    */
   routesDir: string;
 
+  /**
+   * Routes filename; the filename of the generated route file ('routes.ts' by default)
+   */
+  routesFileName: string;
```
dgreene1: This should be `routesFileName?: string;` since it's optional.

src/cli.ts
```diff
@@ -86,6 +86,10 @@ const validateRoutesConfig = async (config: RoutesConfig): Promise<RoutesConfig>
   }
   if (!config.routesDir) { throw new Error('Missing routesDir: Configuration must contain a routes file output directory.'); }
 
+  if (!config.routesFileName) {
```
dgreene1: remove this. See the below comment.

src/routeGeneration/routeGenerator.ts
```diff
@@ -31,7 +31,7 @@ export class RouteGenerator {
   constructor(private readonly metadata: Tsoa.Metadata, private readonly options: RoutesConfig, private readonly minimalSwaggerConfig: SwaggerConfigRelatedToRoutes) { }
 
   public async GenerateRoutes(middlewareTemplate: string, pathTransformer: (path: string) => string) {
-    const fileName = `${this.options.routesDir}/routes.ts`;
+    const fileName = `${this.options.routesDir}/${this.options.routesFileName}`;
```
dgreene1: Don't set the default in `cli.ts` because there are users that use tsoa programmatically that would not get the default value. Please move it here.

So basically:
```ts
const routesFileNameDefault = `routes.ts`;
const routesFileName = this.options.routesFileName ? this.options.routesFileName : routesFileNameDefault;
const fileName = `${this.options.routesDir}/${routesFileName}`;
```

tests/prepare.ts
```diff
@@ -105,6 +111,7 @@ const log = async <T>(label: string, fn: () => Promise<T>) => {
       middleware: 'express',
       middlewareTemplate: './tests/fixtures/custom/custom-tsoa-template.ts.hbs',
       routesDir: './tests/fixtures/custom',
+      routesFileName: 'routes.ts',
```
dgreene1: @HoldYourWaffle here's your opportunity to test this feature. If you set this one to `customRoutes.ts` and if the integration tests don't fail, then your feature works.

src/routeGeneration/routeGenerator.ts
```diff
@@ -31,7 +31,7 @@ export class RouteGenerator {
   constructor(private readonly metadata: Tsoa.Metadata, private readonly options: RoutesConfig, private readonly minimalSwaggerConfig: SwaggerConfigRelatedToRoutes) { }
 
   public async GenerateRoutes(middlewareTemplate: string, pathTransformer: (path: string) => string) {
-    const fileName = `${this.options.routesDir}/routes.ts`;
```
dgreene1: Please add validation here that `this.options.routesDir` is actually a directory. I think the only way that we can do that is check:
* that it doesn't end in `.ts` or `.js` or .json` etc.
  * If it does end in an extension, we should throw an error with the message of `routesDir must be a directory but it was given the path of a file. Perhaps you were looking to use the routesFileName config option?`

src/routeGeneration/routeGenerator.ts
HoldYourWaffle: I knew I was doing something wrong. It just felt wrong every step of the way... I'll get to changing it.

src/config.ts
HoldYourWaffle: I did this for typing reasons under the assumptions that all incoming settings would get their default value in `cli.ts` (which was very dumb, since I'm using TSOA programmatically too...)

tests/prepare.ts
HoldYourWaffle: Would changing just this line be enough?

src/routeGeneration/routeGenerator.ts
HoldYourWaffle: Makes sense.
Wouldn't it make more sense to just [directly check](https://stackoverflow.com/a/15630832/10434371) if a path is a directory?

HoldYourWaffle:
I've addressed your concerns.

I opted for a direct `lstat`-test instead of checking if the directory exists first because we probably want to throw an error if the output directory doesn't exist (right? I could be wrong, just give the word and I'll change it).

I'm very sorry for created so much merge commits (it's honestly hurting my eyes everytime I look at it), but rebasing this would be literal conflict hell so I'd rather avoid that if possible.

src/routeGeneration/routeGenerator.ts
```diff
@@ -31,7 +32,11 @@ export class RouteGenerator {
   constructor(private readonly metadata: Tsoa.Metadata, private readonly options: RoutesConfig, private readonly minimalSwaggerConfig: SwaggerConfigRelatedToRoutes) {}
 
   public async GenerateRoutes(middlewareTemplate: string, pathTransformer: (path: string) => string) {
-    const fileName = `${this.options.routesDir}/routes.ts`;
+    if (!fs.lstatSync(this.options.routesDir).isDirectory()) {
+      throw new Error(`routesDir should be a directory`);
+    }
+
+    const fileName = `${this.options.routesDir}/${this.options.routesFileName || 'routes.ts'}`;
```
dgreene1: I should have noticed this before but we need to check that routesFileName ends in .ts. Not if you use .endsWith please first check if the version of NodeJS that we set in package.json’s engine section supports it.

HoldYourWaffle: I'm not sure why such a check would be necessary, I agree that helpful "warnings" (such as the directory check) can be useful, but this seems like unnecessary hand-holding to me. Maybe I do actually want my generated file to end in something else than `.ts` 🤷‍♂ (can't think of a reason why right now but it's a possibility).
And if someone actually forgot the `.ts` extension it's a 5 second fix, so it doesn't *really* hurt DX in my opinion.
Unlesss there's an actual technical reason why it could cause issues of course, but I can't think of anything non-IDE related that would unconditionally break with a "wrong" file extension.

If you really want it though I'll be happy to add it.

> if you use .endsWith please first check if the version of NodeJS that we set in package.json’s engine section supports it.

An automated test environment could check for these sort of things by running the tests on multiple engines.

HoldYourWaffle: I'm sorry for my broken early-morning English, what I'm trying to say is: I feel like this (in my opinion) unnecessary hand-holding reduces flexibility (albeit for a small group) while not really improving DX. In fact it could be hurting DX for those who want their file to end in something other than `.ts`.

dgreene1: @HoldYourWaffle my primary concern as a volunteer who works on tsoa is to provide functionality that helps the majority of users and that prevents future issues coming in. If even 1 person writes an issue about “hey, why didn’t routesFileName work for me” then it has wasted my free time. Additionally, any erroneous issue created on tsoa’s github makes tsoa look like a less professional repository. So my goal is at all times to diminish the number of times that someone has to reach out to tsoa. The library should provide the support itself.

We need to prioritize the goals and features of the majority user over the fringe case users.

And to be 100% clear, the routes.ts file is in fact a TypeScript file. So if you don’t give it a .ts extension then the person’s project won’t build. Tsoa can not be responsible for that.

HoldYourWaffle: > If even 1 person writes an issue about “hey, why didn’t routesFileName work for me” then it has wasted my free time

That's an interesting viewpoint. I wholeheartedly agree that "sanity checks" are useful, this just seems a little overkill in my opinion. It's really not that difficult to figure out why a file doesn't get an extension if you didn't tell it to.
I have added the check now.

> We need to prioritize the goals and features of the majority user over the fringe case users.

Don't get me wrong, I definitely agree on this. I just think that it's important to *allow* "edge-case" configurations if it's reasonably possible to do so; mostly because I've had to fork multiple projects that didn't want to cooperate in the past (the origin of this PR) and that's just annoying. This isn't the fault of the library per se, but it's always nice to avoid such things.

> And to be 100% clear, the routes.ts file is in fact a TypeScript file. So if you don’t give it a .ts extension then the person’s project won’t build. Tsoa can not be responsible for that.

It wouldn't be TSOA's responsibility, because the user is the one that forgot the extension.

Writing this, maybe a warning would be better suited in this situation? Users that want to do non-standard stuff can still do so but the forgetful ones still get a nice little heads-up. If you want me to leave it as-is that's of course fine too, you're the maintainer after all 😉

dgreene1:
@HoldYourWaffle, I just merged this to discover that the unit tests are failing in CodeShip. Can you explain to me how you were able to commit with failing unit tests? Also, please submit a PR fixing it.

HoldYourWaffle:
That's weird, I never skipped any tests. Will investigate.

HoldYourWaffle:
I found the issue that *caused* the fail, but I can't figure out why the tests didn't run for me.
Either way I should've noticed that they didn't run, my apologies.

(This is the kind of situation an automatic test runner (like travis) would be useful for, to catch stupid mistakes from people like me)

HoldYourWaffle:
I found out why the tests didn't run for me, the GUI I sometimes use to stage/commit apparently doesn't run hooks. My apologies for the inconvenience.

PR #364, opened 2019-06-12
Rename default templates with hbs extension

HoldYourWaffle:
Using the `.ts` file extension for the template files is both semantically incorrect (making it harder to find the templates) and very inconvenient when editing because there are syntax errors literally **everywhere**.
Using `.hbs` as file extension allows ide's to give "proper" syntax highlighting, in the sense of not getting all those errors. You lose the TypeScript highlighting but due to the invalid handlebars syntax this didn't even really work in the first place.

In short: the typescript language "highlighting" makes it much harder to read & edit the templates because of all the errors. Using an `hbs` extension fixes this problem.

A comparison:
Comparison


lukeautry:
Great idea, this was super annoying.

PR #368, opened 2019-07-10
fix: disable eslint along with tslint in templates

scttcper:
some are using eslint instead of tslint, there shouldn't be any negative side effects of disabling both in the default templates

### DIFF
--- a/src/routeGeneration/templates/express.hbs
+++ b/src/routeGeneration/templates/express.hbs
@@ -1,4 +1,5 @@
 /* tslint:disable */
+/* eslint-disable */
 {{#if canImportByAlias}}
   import { Controller, ValidationService, FieldErrors, ValidateError, TsoaRoute } from 'tsoa';
 {{else}}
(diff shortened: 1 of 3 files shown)


endor:
Hm, I can't really follow why this makes sense. This project uses `tslint`, so it makes sense to explicitly disable that for the templates. If you use `tsoa` within your project, it's a dependency, and you probably don't want to lint your dependencies, right?

dgreene1:
@endor I think it’s because routes.ts lives within the consumers own repo so it does get linted. I could see why this would be a problem for people. I’d nominate to approve this PR, but I’ll allow @endor to respond first.

@scttcper, you’re okay with using eslint’s ignore / glob options in the meantime right?

endor:
Okay, that's a fair point. But I'm still not sure whether it wouldn't be better to tell the linter to just ignore the `routes.ts` file, since it's generated.

scttcper:
@endor @dgreene1 When moving people at our company from tslint to eslint we were looking for lowest friction. Routes.ts would be the only thing we need to ignore and tslint was already ignoring it. 

Totally fine using .eslintignore file

dgreene1:
Since [tslint has officially been deprecated in favor of eslint](https://medium.com/palantir/tslint-in-2019-1a144c2317a9), I think it makes a lot of sense to accept this PR.

PR #369, opened 2019-07-14
Allow explicitly typing the SuccessResponse

WoH:
This PR addresses a use case where one would want to return a type union `A | B` documented as `SuccessResponse<A>()` and `Response<B> `from a controller.
In these cases, the function body returns a type union of `A | B` to get full TS support.
In order to document this behavior using TSOA, it should be possible to override the default return type using the SuccessResponse decorator (i.e. `@SuccessResponse<A>(...)`).

Alternatives considered:
- You can't use `@Response<A>(200, ...)` since the default behavior of TSOA is to overwrite that with the default SuccessResponse [here](https://github.com/lukeautry/tsoa/blob/master/src/metadataGeneration/methodGenerator.ts#L43).
- Returning type `A` from the controller function and returning `B as any` feels bad

Syntax: `@SuccessResponse<OverrideType>(...)`

Please note that the test right now abuses a typecast as any. If there are suggestions on how to test the behavior more efficiently I'd gladly apply that suggestion.
Also, this does not address returning a type union within one response documented as oneOf by default.

### DIFF
--- a/tests/fixtures/controllers/methodController.ts
+++ b/tests/fixtures/controllers/methodController.ts
@@ -71,6 +71,13 @@ export class MethodController extends Controller {
         return Promise.resolve();
     }
 
+    @SuccessResponse<TestModel>('201', 'Created')
+    @Get('typedSuccessResponse')
+    public async typedSuccessResponse(): Promise<void> {
+        this.setStatus(201);
+        return Promise.resolve();
+    }
+
     @Security('api_key')
     @Get('ApiSecurity')
     public async apiSecurity(): Promise<TestModel> {
--- a/src/decorators/response.ts
+++ b/src/decorators/response.ts
@@ -1,4 +1,4 @@
-export function SuccessResponse(name: string | number, description?: string): Function {
+export function SuccessResponse<T>(name: string | number, description?: string): Function {
   return () => { return; };
 }
 
--- a/src/metadataGeneration/methodGenerator.ts
+++ b/src/metadataGeneration/methodGenerator.ts
@@ -173,7 +173,9 @@ export class MethodGenerator {
       description,
       examples,
       name,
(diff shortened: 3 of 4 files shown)


tests/fixtures/controllers/methodController.ts
```diff
@@ -71,6 +71,13 @@ export class MethodController extends Controller {
         return Promise.resolve();
     }
 
+    @SuccessResponse<TestModel>('201', 'Created')
+    @Get('typedSuccessResponse')
+    public async typedSuccessResponse(): Promise<void> {
+        this.setStatus(201);
+        return Promise.resolve();
```
dgreene1: This example clearly states why I think this is a dangerous approach. Your swagger model is going to say that you API returns an object with the shape `TestModel` but your code does not actually return a `TestModel` at all. To me this defeats the purpose of using TypeScript. I apologize, but I really can't accept this PR. However, now that I understand your use case better, I believe what you're looking for is the ability to clearly define the the shape of the error.

> As a user I'd want to know why my requests fail. As a developer, when I know I can't handle a request, I want to not just throw an error and exit but have a well defined and documented response through swagger (with example).
> Writing typescript, I'd want to retrieveModel(...conditions): Model | ModelNotFound.
> In swagger, both these things should be documented, the 200 with the model and the 404 ModelNotFound with specific reasons why the model was not found.

To accomplish `Model | ModelNotFound` the best way to accomplish that translation between the union type and the swagger model is to implement `$OneOf`. So I'm going to close this PR, and then if you or someone wants to pick up issue https://github.com/lukeautry/tsoa/issues/393, then it would be much easier to accomplish the use case you're asking for. But I can't approve a PR that would cause the swagger model to be out of sync with the code since that's the mission statement of tsoa.

PR #375, opened 2019-07-26
Corrects the oauth scope type + adds husky to ensure contribution quality

dgreene1:
@lukeautry / @endor I apologize for making a PR without an issue first. As a co-maintainer of tsoa, I should know better! :/ That being said, this change seems pretty innocuous since it's mostly just a type change.

**Context:**
If you're using tsoa programmatically, and you tried to send in scopes as their correct type, you will get a TypeScript error:
image

**Solution:**
Update the swagger type definitions to use the correct type for `scopes`. Check those out here:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/a59cd46314f728a29da753dabaa99e75fbea6d89/types/swagger-schema-official/index.d.ts#L216

**Bonus**
I added `husky` which forces a the unit tests and linter to be run before anyone can commit their code. I though that this extra bit of quality checking would be valuable since we allow a lot of contributions from the community and since we don't run the unit tests as a blocker before a PR is merge-able (or at least I don't think we have Travis running anything like that). I probably should have made that a separate PR so that these two issues (the swagger types & the husky hooks) could be discussed independently. My bad! @lukeautry, let me know if you want me to break them apart so that their atomic.

tl;dr: If I don't have husky running the unit tests for me, I often forget to run them. Ultimately having Travis set up to run the tests on each PR is probably a more full-proof approach.

### DIFF
--- a/src/swagger/swagger.ts
+++ b/src/swagger/swagger.ts
@@ -301,7 +301,7 @@ export namespace Swagger {
     description?: string;
     flow: 'password';
     tokenUrl: string;
-    scopes?: OAuthScope[];
+    scopes?: OAuthScope;
   }
 
   export interface OAuth2ApplicationSecurity {
@@ -309,7 +309,7 @@ export namespace Swagger {
     description?: string;
     flow: 'application';
     tokenUrl: string;
-    scopes?: OAuthScope[];
+    scopes?: OAuthScope;
   }
 
   export interface OAuth2AccessCodeSecurity {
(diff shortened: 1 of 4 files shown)


dgreene1:
@endor / @lukeautry do you guys care if I merge this? It's just additional quality checking (via husky) and some type corrections.

src/swagger/swagger.ts
```diff
@@ -318,11 +318,11 @@ export namespace Swagger {
     flow: 'accessCode';
     tokenUrl: string;
     authorizationUrl: string;
-    scopes?: OAuthScope[];
```
endor: if this change doesn't break any tests.. does that mean we're not testing this code yet?

Issue #170, opened 2017-10-11
"headers" property for responses is not generated in swagger


ThomasKruegl:
Hi,

we are trying to use tsoa to generate a swagger specification for our API. We are running into several issues, here is one of them.
There are other issues here on github regarding the use of custom headers in responses. From what i have understood it is now possible to set custom headers, and this works fine for the generated routes.
But for the generated swagger, there is simply no way of getting the headers property created for responses right?

kind regards,
Thomas

dgreene1:
I believe the answer is to use the `@Request` parameter decorator. I’ve added more information to the readme on how to use it. Once you have that (and have set the type to be either `Express.Request` or `Koa.Request`), you have the full ability to set the header in the same that you would have with pure Express or Koa.

As for declaring it in Swagger... I actually don’t know (I just recently joined as a maintainer).

So I’ll mark this issue as [help wanted] so that maybe someone can either:
* tell us if there is a way to do this in tsoa
* or add that functionality

My apologies for the delay in answering. We have new, additional maintainers who will be adding more support in the future.

ThomasKruegl:
Currently i am working in a completely different environment, and don't have time to follow up on this. Just saying so you know i won't be able to test or do anything here.
But thanks for picking this up anyway :)

sundowndev:
Up, this is still relevant today. `this.setHeader` works for generated routes but doesn't write in the specs. So for now it's impossible to change `application/json` response content type in the docs.

```ts
const stat: fs.Stats = await fs.promises.stat(filePath);

this.setStatus(200);
this.setHeader('Content-Type', 'application/pdf');
this.setHeader('Content-Length', stat.size.toString());
```

![image](https://user-images.githubusercontent.com/16480203/94701712-40cd0b00-033d-11eb-853d-d687b9a5bf64.png)

WoH:
> Up, this is still relevant today. `this.setHeader` works for generated routes but doesn't write in the specs. So, from what I understand, for now it's impossible to change `application/json` response content type in the docs.
> 
> ```ts
> const stat: fs.Stats = await fs.promises.stat(filePath);
> 
> this.setStatus(200);
> this.setHeader('Content-Type', 'application/pdf');
> this.setHeader('Content-Length', stat.size.toString());
> ```
> 
> ![image](https://user-images.githubusercontent.com/16480203/94701712-40cd0b00-033d-11eb-853d-d687b9a5bf64.png)

[...]

PR #378, opened 2019-07-30
Add support for object keyword

lucas-rudd:
This PR adds support for interfaces that look like the following
```
export interface IApiResponse {
  statusCode: number;
  body?: string;
  headers?: object;
}
```

The current functionality results in an error being thrown:
`Error: Unknown type: ObjectKeyword`

While, the expected behavior is to generate a swagger definition with the `headers` value set to the following
```
headers:
   type: object
   x-nullable: true
```

### DIFF
--- a/src/metadataGeneration/typeResolver.ts
+++ b/src/metadataGeneration/typeResolver.ts
@@ -70,6 +70,10 @@ export class TypeResolver {
       return { dataType: 'any' } as Tsoa.Type;
     }
 
+    if(this.typeNode.kind === ts.SyntaxKind.ObjectKeyword) {
+      return { dataType: 'object' } as Tsoa.Type;
+    }
+
     if (this.typeNode.kind !== ts.SyntaxKind.TypeReference) {
       throw new GenerateMetadataError(`Unknown type: ${ts.SyntaxKind[this.typeNode.kind]}`);
     }
--- a/tests/fixtures/testModel.ts
+++ b/tests/fixtures/testModel.ts
@@ -37,6 +37,8 @@ export interface TestModel extends Model {
    */
   boolValue: boolean;
   boolArray: boolean[];
+  object: object;
+  objectArray: object[];
   enumValue?: EnumIndexValue;
   enumArray?: EnumIndexValue[];
   enumNumberValue?: EnumNumberValue;
--- a/tests/integration/express-server.spec.ts
+++ b/tests/integration/express-server.spec.ts
@@ -781,6 +781,8 @@ describe('Express Server', () => {
           testSubModel2: false,
         },
       },
+      object: { foo: 'bar' },
+      objectArray: [{ foo1: 'bar1'}, { foo2: 'bar2'}],
       numberArray: [1, 2],
       numberValue: 5,
       optionalString: 'test1234',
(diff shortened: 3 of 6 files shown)


dgreene1:
@lucas-rudd what issue number does this solve?

I’m closing the PR until we agree that it’s a problem that tsoa should solve. I.e. we need to know more about the use case. If you’re trying to find a way to add the headers for the response then this is not the correct way to do it. Headers aren’t supposed to be returned on the response.data property.

Additionally, the use of `Object` is strongly discouraged in TypeScript.

If I’m missing something (and I probably am) feel free to reply and maybe one of the other maintainers can fill in what I’ve missed. I think maybe an issue can be created to try to come up with a way to define the headers in the Swagger YAML. But again, I don’t think this would be the solution.

dgreene1:
Nevermind, I just realized that we already support `{ dataType: 'any' }` even though that is also discouraged. So I'm going to re-open the PR. Thank you for your contribution. I still think it would be great to open an issue first before submitting a PR.

lucas-rudd:
Sorry about that. I have an issue opened now, but it's already a bit after the fact 😅 

I guess I did not explain the use-case properly.

Just to give a little more context. I would really love to use this tool. However, the code-base I'm working in is already well established, and does not always follow best practices (e.g. the use of the `any` keyword, and the use of the `object` keyword for types, when an actual type could easily be created).

Refactoring such a large code-base to follow best practices is ideal; but it would be a pain to have write interfaces for all our cases of the `object` keyword _before_ utilizing this tool.

I imagine other people are in a similar boat.

I was able to npm link `tsoa` with this change to my code-base and generate swagger docs for several handlers, and it's a game-changer. We had previously been writing docs by hand. This creates much more detailed and accurate documentation that we will be able to very easily keep up-to-date (unlike our current docs).

dgreene1:
@lucas-rudd this old issue is asking how to set the header type in the swagger yaml. Did you say that you found a way to do that? Because from your description above it looks like you'd actually be setting `response.data.headers` instead of `response.headers`.

But if you have solved that, it would be great if you could pass that knowledge down to this PR below:
https://github.com/lukeautry/tsoa/issues/170

lucas-rudd:
The interface above was only an example of something that has an `object` type in the contract. I do not think this is in any way related to the issue mentioned above.

The swagger generated from this is something like the following
```
definitions:
    IApiResponse:
        properties:
            statusCode:
                type: number
                format: double
            body:
                type: string
                x-nullable: true
            headers:
                type: object
                x-nullable: true
        required:
            - statusCode
        type: object
paths:
    'path':
        post:
            operationId: Controller
            produces:
                - application/json
            responses:
                '201':
                    description: 'Certificate created'
                    schema:
                        $ref: '#/definitions/IApiResponse'
                ...
```

I _think_ that issue is regarding how to generate something more like this
```
definitions:
    IApiResponse:
        properties:
            statusCode:
                type: number
                format: double
            body:
                type: string
                x-nullable: true
        required:
            - statusCode
        type: object
paths:
    'path':
        post:
            operationId: Controller
            produces:
                - application/json
            responses:
                '201':
                    description: 'Certificate created'
                    headers:
                      Example-Header:
                        schema:
                          type: integer
                    schema:
                        $ref: '#/definitions/IApiResponse'
...
```

I did not find a way to do that.

dgreene1:
Thank you for clarifying. I merged the PR. I think everyone who uses `any` or `Object` knows that their not getting a ton of advantage from tsoa anyway. So if it helps people to adopt this library and build up from there... then I think that’s great! :)

PR #383, opened 2019-08-04
Rate limit decorator for tsoa

caringdeveloper:
# Why?
It would be nice to add rate limiting to certain endpoints in an API with a decorator so that the middleware for rate limiting is attached to the routes by the tsoa generator.

# What was done?
To support this idea I added a new @RateLimit() decorator which can be used to decorate routes that should be protected by a rate limiting strategy.

As a starting point this is currently implemented for express.js only (mainly because all my backends run on express.js) but the goal is to support this for hapi and koa as well.

# What still needs to be done?
* hapi support (help wanted)
* koa support (help wanted)
* support more options besides number of requests and search window

# How to use?
```javascript
@Get()
@SuccessResponse(200, "List of all ogs")
@Security("Bearer", ["user", "admin"])
@RateLimit(1, 60)
public async GetAllDogs(): Promise<DogDto[]> {
  return await this._getAnimals();
}
```

Best regards,
caringdeveloper

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
   "author": "Luke Autry <lukeautry@gmail.com> (http://www.lukeautry.com)",
   "license": "MIT",
   "dependencies": {
+    "express-rate-limit": "^5.0.0",
     "fs-extra": "^7.0.1",
     "handlebars": "^4.1.2",
     "lodash.indexof": "^4.0.5",
--- a/src/decorators/rateLimit.ts
+++ b/src/decorators/rateLimit.ts
@@ -1,0 +1,7 @@
+/**
+ * @param {requestCount} number Allowed request in time frame
+ * @param {searchWindow} number Time frame
+ */
+export function RateLimit(requestCount: number, searchWindow: number): Function {
+  return () => { return; };
+}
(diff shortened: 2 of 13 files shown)


dgreene1:
This idea is cool, but it doesn't have anything to do with the design goals of tsoa. Additionally, there was no issue created before the PR so we (i.e. the maintainers) didn't have an opportunity to weigh in on this sooner. So I apologize that you already did some work, but you really should create a Github issue first.

I was about to recommend that this be extracted into it's own library, but while reviewing the PR I realized that it already is it's own library (`express-rate-limit`).

So I'm going to close this PR since anyone who wants to rate limit their routes can do this with their own custom middleware. That's what `koa` and `express` are designed to do: allow for custom middleware to be easily added.

I apologize for closing a PR that has functionality that is important to you, but I think it's important that a library "Do One Thing and Do It Well."

That being said, please feel free to open an issue for something like "TSOA's ability to allow for 3rd party decorators." If enough tsoa users think it would be valuable to have a plugin system, then this would allow users to include your rate limiter plugin or whatever plugin they would like. I think this is the more extensible solution rather than including everyone's particular needs into tsoa. Think of it like how there are custom eslint rules for things that are not deemed appropriate for inclusion in the core eslint repository. I should know... one of [my favorite PRs that I've ever written was closed for the same reasons]:(https://github.com/palantir/tslint/pull/4318#pullrequestreview-178609382):
1) There was either no issue created first or the issue wasn't marked as "Accepting PRs"
2) The idea was better off as a standalone package

azgron:
Is there anything new with the external package? (like nestjshttps://github.com/nestjs/throttler)

Issue #336, opened 2019-02-26
Option to restrict extra properties in request body


jorshali:
TSOA allows sending more properties than explicitly defined in an interface.  For example, if my interface expects an Employee JSON as the body defined as:
```
export class Employee {
  public username: string;
  public firstName: string;
  public lastName: string;
}
```
A caller could send the following body and it would be acceptable:
```
{
  "username": "johndoe",
  "firstName": "John",
  "lastName": "Doe",
  "employeeId": 1234
}
```
I'm requesting an option to disallow this through the tsoa.json configuration.  For example,
```
{
  "routes": {
    ... ...
    "strictProperties": true
  }
}
```
In the example above, the request body with the `employeeId` included would fail as a bad request if the `strictProperties` option was set to true.

To avoid impacting existing APIs this would be false by default.  If any extra properties were provided, a validation error would be returned with a bad request response.

dgreene1:
@jorshali this feature was added via https://github.com/lukeautry/tsoa/pull/385 and released in last version. See the instructions in the updated readme: https://github.com/lukeautry/tsoa#how-to-ensure-no-additional-properties-come-in-at-runtime

Issue #264, opened 2018-08-28
Implement a "strict" requests checking mode 


mushketyk:
At the moment `tsoa` only validates if all fields in a request are present, but it does not prevent undeclared fields from being passed to a request handler.

For example, if we have an API like:

```typescript
export class CreateUserRequest {
  name: string

  surname: string
}

Route('users')
class UsersController {

  @Post()
  public async createUser(@Body() user: CreateUserRequest) {
    ...
  }
}
```

With this definition, `tsoa` will validate that both `name` and `surname` fields are present, but it allows sending a request with the following body:

```javascript
{
  "name": "John",
  "surname": "Doe",
  "id": "new-id" // Some malicious field passed in the request
}
```

`tsoa` allows to send a request like this and will pass additional fields to a request handler.

I think that `tsoa` should implement a "strict" mode (maybe even enabled by default) that should either:

* Filter out unexpected fields
* Return 400 if there are unexpected fields in a request

mushketyk:
@endor I might help with this. How do you prefer to handle the "strict" mode? Should it be enabled by default? Should it return 4xx? Should this be configurable?

fhewitt:
If should be configurable both on an global way through tsoa config (because it break backward compatibility if enabled by default, therefore need an easier global fix), and override by specific endpoints for fine control and progressive implementation through decorator `@Strict` or similar.

Yes, it should return a 400 Bad Request

endor:
Why would you want some of your endpoints to be strict and some not? That would seem like an inconsistent API. You're probably right about keeping backwards compatibility by turning it off by default so it behaves as before, but once you turn it on, shouldn't it really be turned on for all?

dgreene1:
> @endor From my point of view this should be default with an option to turn it off.

I agree. Yes, it would be a breaking change, but it's a change that would follow TypeScript's own design where you have to explicitly allow excess properties in defined interfaces. So as a newbie to `tsoa` it's a bit of a surprise to discover that the default is "not strict mode."

Btw, I'd also be interested in helping with this one. Has anyone started on a branch? (@mushketyk?) Or do we need more consensus before starting this work?

> @endor Why would you want some of your endpoints to be strict and some not? That would seem like an inconsistent API. You're probably right about keeping backwards compatibility by turning it off by default so it behaves as before, but once you turn it on, shouldn't it really be turned on for all?

[...]

PR #385, opened 2019-08-05
Issue 264 disallow excess properties

dgreene1:
This adds a new config option called `noImplicitAdditionalProperties` that solves https://github.com/lukeautry/tsoa/issues/264 and https://github.com/lukeautry/tsoa/issues/336.

In order to accomplish a large PR like this I wanted to make sure I had a ton of unit test coverage. So to make that easier I also added to the launch.json and tasks.json a way for us to debug the unit tests! 🎉

Please note, this looks like a lot of code changes, but it's probably 80% new unit tests.

**Note:** I did not turn this option on by default because:
(a) in swagger, additional properties are allowed by default
(b) if I did turn it on by default that would create breaking contracts for existing APIs. APIs will need to inform their users so they can take the time to stop sending extra properties. But this will be great for new APIs that are being written! 😃


README.MD
```diff
@@ -141,21 +141,23 @@ See [CLI documentation](#command-line-interface)
...
+    await generateSwaggerSpec(swaggerOptions);
 
-    await generateSwaggerSpec({
+    await generateRoutes({
         basePath: "/api",
         entryFile: "./api/server.ts",
-        outputDirectory: "./api/dist",
-    });
+        routesDir: "./api",
+    }, swaggerOptions);
```
dgreene1: I had to update the documentation sincee the new option `noImplicitAdditionalProperties` is required by both cli commands (`generateRoutes` and `generateSwaggerSpec`) due to it affecting both swagger creation and runtime validation.

package.json
```diff
@@ -29,7 +29,7 @@
     "preversion": "npm test",
     "postversion": "git push origin master && git push --follow-tags",
     "pretest": "cross-env NODE_ENV=tsoa_test ts-node ./tests/prepare.ts",
-    "test": "cross-env NODE_ENV=tsoa_test mocha **/*.spec.ts --exit --compilers ts:ts-node/register",
+    "test": "tsc -p ./tests/tsconfig.json --noEmit && cross-env NODE_ENV=tsoa_test mocha **/*.spec.ts --exit --require ts-node/register",
```
dgreene1: I discovered while adding new unit tests that the `./tests/` folder was not actually compiling. This went undetected because (a) we don't have husky running the tests (but we could if https://github.com/lukeautry/tsoa/pull/375 is merged) and (b) even if we did, ts-node kinda sorta bypasses the compilation. So I wanted to explicitly compile the tests directory to make sure that we don't have bugs creep in.

src/metadataGeneration/tsoa.ts
```diff
@@ -68,8 +68,18 @@ export namespace Tsoa {
...
 
+  export type TypeStringLiteral = 'string' | 'boolean' | 'double' | 'float' | 'integer' | 'long' | 'enum' | 'array' | 'datetime' | 'date' | 'binary' | 'buffer' | 'byte' | 'void' | 'object' | 'any' | 'refEnum' | 'refObject';
+
+  export type RefTypeLiteral = 'refObject' | 'refEnum';
+
+  export type PrimitiveTypeLiteral = Exclude<TypeStringLiteral, RefTypeLiteral | 'enum' | 'array' | 'void'>;
+
   export interface Type {
-    dataType: 'string' | 'double' | 'float' | 'integer' | 'long' | 'enum' | 'array' | 'datetime' | 'date' | 'buffer' | 'void' | 'object' | 'any' | 'refEnum' | 'refObject';
+    dataType: TypeStringLiteral;
```
dgreene1: I'm exporting the union of types so that I could do exhuastiveness checking. See the `assertNever` method I'm using https://github.com/lukeautry/tsoa/pull/385/files#diff-817ee96768e8e5542508a612054dfd39R1

src/swagger/specGenerator.ts
```diff
@@ -35,30 +36,111 @@ export class SpecGenerator {
     return methodName.charAt(0).toUpperCase() + methodName.substr(1);
   }
 
-  protected getSwaggerType(type: Tsoa.Type): Swagger.Schema {
-    const swaggerType = this.getSwaggerTypeForPrimitiveType(type);
-    if (swaggerType) {
-      return swaggerType;
+  public throwIfNotDataFormat(strToTest: string): Swagger.DataFormat {
+    const guiltyUntilInnocent = strToTest as Swagger.DataFormat;
+    if (
```
dgreene1: The cool thing about this change is that the compiler (via `assertNever`) will inform us on where changes are necessary if someone were to add a new type to the `Swagger.DataFormat` union.

tests/fixtures/controllers/getController.ts
```diff
@@ -40,6 +40,12 @@ export class GetTestController extends Controller {
     modelsArray: new Array<TestSubModel>(),
     numberArray: [1, 2, 3],
     numberValue: 1,
+    object: {
```
dgreene1: Btw, I had to change this because there was a typescript error in tests. That's why I added `tsc` to the `test` command in package.json, so errors caused by previous submissions would be caught sooner.

src/metadataGeneration/tsoa.ts
```diff
@@ -82,7 +92,7 @@ export namespace Tsoa {
     elementType: Type;
   }
 
-  export interface ReferenceType extends Type {
+  export interface ReferenceType extends Type, RefTypeMinimal {
     description?: string;
     dataType: 'refObject' | 'refEnum';
```
endor: why not have this as `RefTypeLiteral`?

src/module/generate-swagger-spec.ts
```diff
@@ -1,6 +1,7 @@
 import * as ts from 'typescript';
 import * as YAML from 'yamljs';
 import { SwaggerConfig } from '../config';
+export { SwaggerConfig } from '../config';
```
endor: aren't we importing the same thing above?

src/routeGeneration/routeGenerator.ts
```diff
@@ -3,10 +3,14 @@ import * as path from 'path';
 import * as tsfmt from 'typescript-formatter';
 import { Tsoa } from '../metadataGeneration/tsoa';
 import { fsReadFile, fsWriteFile } from '../utils/fs';
-import { RoutesConfig } from './../config';
+import { RoutesConfig, SwaggerConfig } from './../config';
 import { normalisePath } from './../utils/pathUtils';
 import { TsoaRoute } from './tsoa-route';
 
+interface ISwaggerConfigRelatedToRoutes {
```
endor: We're not using the convention to prepend an `I` to interfaces anywhere else, so I wouldn't add it here, just to keep things consistent

dgreene1: ✔️

src/module/generate-swagger-spec.ts
dgreene1: It looks similar but it's actually an export. I'm doing that so that users who are using tsoa programmatically are able to import the `SwaggerConfig` interface.

Otherwise users will have to resort to advanced types (`Parameters<T>`) like I did to be able to initialize a `SwaggerConfig` variable:
```ts
// I wish I could have written this as
// const swaggerGenArgs: SwaggerConfig = [
const swaggerGenArgs: Parameters<typeof tsoa.generateSwaggerSpec> = [
		{
			outputDirectory: staticFileServingDir,
			entryFile: './src/index.ts',
			host: 'localhost:3000',
			basePath: '/api/v1',
			yaml: true,
		},
	];

	const swaggerGenerationResult = await tsoa.generateSwaggerSpec(...swaggerGenArgs);
```

src/metadataGeneration/tsoa.ts
dgreene1: Are you asking if `RefTypeMinimal` should be called `RefTypeLiteral`? `RefTypeMinimal` isn't a string literal, it's an object. So I think calling it `RefTypeLiteral` would be confusing. 

```ts
  export interface RefTypeMinimal {
    dataType: RefTypeLiteral;
  }
```

But I am open to changing the name if you have other ideas for `RefTypeMinimal`.

dgreene1: Oh, I think I understand why `dataType` isn't `RefTypeLiteral`. I can make that change.

dgreene1:
@endor I made the corrections you asked for. I also realized while I was in there that my new tests for OAS 3 were false-positive, so I corrected that. Sorry to have you re-review, but hopefully you can just check out the last commit https://github.com/lukeautry/tsoa/pull/385/commits/7ef2bd9e94c7e888bd76359aa888f0cc3b22a206

src/swagger/specGenerator.ts
```diff
@@ -35,30 +36,111 @@ export class SpecGenerator {
...
+        return this.getSwaggerTypeForEnumType(type as Tsoa.EnumerateType);
+    } else {
+        return assertNever(type.dataType);
+    }
   }
 
   protected getSwaggerTypeForReferenceType(referenceType: Tsoa.ReferenceType): Swagger.BaseSchema {
-    return {};
+    return {
+        // Dont' set additionalProperties value here since it will be set within the modle of the $ref when it gets created
```
endor: typo in model

src/swagger/specGenerator.ts
```diff
@@ -35,30 +36,111 @@ export class SpecGenerator {
...
+  protected getSwaggerTypeForPrimitiveType(dataType: Tsoa.PrimitiveTypeLiteral): Swagger.Schema {
+
+    const defaultAdditionalPropertiesSetting = true;
+
+    if (dataType === 'object') {
+      if (process.env.NODE_ENV !== 'tsoa_test') {
+        // tslint:disable-next-line: no-console
+        console.warn(`The type Object is discouraged. Please consider using an interface such as:
+              export interface IStringToStringDictionary {
+                  [key: string]: string;
```
endor: very minor, but here you use two spaces and below four.

src/swagger/specGenerator.ts
```diff
@@ -35,30 +36,111 @@ export class SpecGenerator {
...
 
-  protected getSwaggerTypeForPrimitiveType(type: Tsoa.Type): Swagger.Schema | undefined {
-    const map = {
-      any: { type: 'object' },
+  protected getSwaggerTypeForPrimitiveType(dataType: Tsoa.PrimitiveTypeLiteral): Swagger.Schema {
+
+    const defaultAdditionalPropertiesSetting = true;
+
+    if (dataType === 'object') {
+      if (process.env.NODE_ENV !== 'tsoa_test') {
```
endor: why is using `object` not discouraged in tests?

src/swagger/specGenerator2.ts
```diff
@@ -62,7 +63,16 @@ export class SpecGenerator2 extends SpecGenerator {
         };
 
         if (referenceType.additionalProperties) {
-          definitions[referenceType.refName].additionalProperties = this.buildAdditionalProperties(referenceType.additionalProperties);
+            definitions[referenceType.refName].additionalProperties = this.buildAdditionalProperties(referenceType.additionalProperties);
```
endor: hm.. there are four spaces here vs. two spaces in the rest of the file?

src/swagger/specGenerator2.ts
```diff
@@ -62,7 +63,16 @@ export class SpecGenerator2 extends SpecGenerator {
...
+        } else {
+            // Since additionalProperties was not explicitly set in the TypeScript interface for this model
+            //      ...we need to make a decision
+            if (config.noImplicitAdditionalProperties) {
```
endor: couldn't this just be `definitions[referenceType.refName].additionalProperties = !config.noImplicitAdditionalProperties`?

endor: this makes me think.. why do we use two different names that we have to be careful with because they're opposites?

tests/integration/koa-server.spec.ts
```diff
@@ -71,7 +72,7 @@ describe('Koa Server', () => {
 
     return verifyPostRequest(basePath + '/PostTest', data, (err: any, res: any) => {
       const model = res.body as TestModel;
-      expect(model).to.deep.equal(model);
+      expect(model).to.deep.equal(data);
```
endor: nice catch (:

src/swagger/specGenerator.ts
dgreene1: I didn’t want the test output to get cluttered with these console logs. And since we don’t use a logger with levels (such a `pino` or `Bunyan`) this is the only want to configure the log to not show up on test runs. The beauty is that thanks to us using `tsoa_test` instead of the standard `test` this should still give the message to all of the consumers of the library.

endor: But what I mean is.. if we generally discourage it, why is it in the tests? Or am I misunderstanding something?

src/swagger/specGenerator2.ts
dgreene1: Very good point on the opposite name issue. I elected to choose a “negative” variable name because the config is optional.

The alternative would be more onerous. Like if we called it `allowAdditionalProperties` then the default would have to be true since that’s what Swagger does.

So our internal code would be:
```ts
schema.additionalProperties = config.allowAdditionalProperties !== undefined && config.allowAdditionalProperties === false ? false : true;
```

Actually as I’m writing this, it’s not that bad. Maybe I was just overthinking it? I’ll ask the people who first asked for this feature— What do you think @mushketyk and @jorshali?

src/swagger/specGenerator2.ts
dgreene1: I can fix it. I’d really like us to switch over to `prettier` since it would fix this for us maintainers. Would you be open to me migrating us to prettier after this PR merged? If I do it as part of this PR it will light up the entire repository with changes and then it won’t be possible to see what was substantively changed.

endor: Uh, sure. I've never used it. Isn't linting able to tell us about this anyway?

dgreene1: Prettier formats the code for you. So we could remove the three annoying longer rules (dangling comma, ending white space, two spaces) and then just have Prettier fix is when the commit happens.
https://github.com/prettier/prettier

src/swagger/specGenerator.ts
dgreene1: Its included in the tests because I merged this PR that a new tsoa user really wanted: https://github.com/lukeautry/tsoa/pull/378

Unfortunately after zip merged that PR, I realized that `object` doesn’t make a lot of sense. It makes even less sense when you don’t allow additional properties. I’d appreciate any guidance. What do you think we should do @endor?

Btw, here’s a great description of why object is strange in TypeScript. https://stackoverflow.com/a/18961904/706768

endor: I guess we'll have to live with it. It's not a big deal, just something we'll probably be wondering about after a while. It's a difficult line to draw between nudging people to use a better way while also allowing older codebases.

src/swagger/specGenerator2.ts
dgreene1: Looks like we're only waiting for people to agree on a name for the new option, otherwise this could get merged.

@endor I just pushed the remaining fixed.

dgreene1: Alright, since we haven't heard any of the tie breakers, I'm going to go with my gut and merge this. There are a couple of PRs in the pipeline that depend on the structure of this code, so I'm going to merge it.

Issue #384, opened 2019-08-05
Better documentation for when external interfaces are not found


dgreene1:
**Proposal:**
Add a more clear message when tsoa can not find the interface that you're using.

Specifically, I'd like to recommend we change this:
```
throw new GenerateMetadataError(`No matching model found for referenced type ${typeName}.`);
```
to be:
```
throw new GenerateMetadataError(`No matching model found for referenced type ${typeName}. If ${typeName} comes from a dependency, please create an interface in your own code that has the same structure. Tsoa can not utilize interfaces from external dependencies.`);
```

**Context:**
It seems like a common challenge of adopting `tsoa` is the idea that the interfaces that you expose to your consumers should not change. Therefore `tsoa` does not allow you to use interfaces that are defined in dependencies. This has caused problems for a few users ( https://github.com/lukeautry/tsoa/issues/340, https://github.com/lukeautry/tsoa/issues/377, https://github.com/lukeautry/tsoa/issues/334, and https://github.com/lukeautry/tsoa/issues/205)

PR #386, opened 2019-08-05
added a better error message and a link to a full explanation

dgreene1:
This fixes https://github.com/lukeautry/tsoa/issues/384

### DIFF
--- a/ExternalInterfacesExplanation.MD
+++ b/ExternalInterfacesExplanation.MD
@@ -1,0 +1,166 @@
+# Why am I getting this error?
+
+`No matching model found for referenced type`
+
+If you get that error, the most likely cause is that you are asking TSOA to create a swagger model from an interface that lives in an external dependency.
+
+## Why is that a problem?
+
+* Technical problem
+  * TSOA can't currently get interfaces from node_modules. For performance reasons, TSOA should not be crawling all of the files in node_modules.
+* Architectural/Quality problem
+  * The consumers of the API you are writing do not want to have that API contract change on them. If your swagger/open-api documents/contracts were to be generated from external interfaces they would risk being changed any time that you update a library.
+
+# Solution:
+
+Create an interface that lives inside your own code base (i.e. it does live in `./node_modules`) but that has the same structure.
+
+## Detailed Solution
+
+Here's the code that's getting the `No matching model found` error:
+```ts
+import * as externalDependency from 'some-external-dependency';
+
+@Route('Users')
+export class UsersController {
+
+    /**
+     * Create a user
+     * @param request This is a user creation request description
+     */
+    @Post()
+    public async Create(@Body() user: externalDependency.IUser): Promise<void> {
+        return externalDependency.doStuff(user);
+    }
+
+}
+```
+And that external dependency has the following code:
+```ts
+// node_modules/someExternalDependency/index.d.ts
+
+export interface IUser {
+    name: string;
+}
+
+export function doStuff(user: IUser) {
+    // ...
+}
+```
+
+Here's how you solve it:
+```ts
+import * as externalDependency from 'some-external-dependency';
+
+// a local copy:
+interface IUserAbstraction {
+    name: string
+}
+
+@Route('Users')
+export class UsersController {
+
+    /**
+     * Create a user
+     * @param request This is a user creation request description
+     */
+    @Post()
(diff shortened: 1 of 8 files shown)


ExternalInterfacesExplanation.MD
```diff
@@ -0,0 +1,166 @@
...
+// node_modules/humanLogger/index.ts
+export interface IHuman {
+    name: string;
+}
+
+export function onlyAcceptsAHuman(human: IHuman){
+    console.log(`I got a human with the name ${human.name}`
+}
+```
+And here's your local code that has it's own interface that has a different name but the same structure. The fact that the following code compiles shows that the two interfaces can be substituted for each other due to structural subtyping:
```
endor: its own

ExternalInterfacesExplanation.MD
```diff
@@ -0,0 +1,166 @@
...
+     */
+    @Post()
+    public async Create(@Body() user: IUserAbstraction): Promise<void> {
+        return externalDependency.doStuff(user);
+    }
+
+}
+```
+
+## But wait, but my code needs get data that the library can handle... will that work?
```
endor: I don't understand that sentence.. what does it mean?

ExternalInterfacesExplanation.MD
```diff
@@ -0,0 +1,166 @@
...
+humanLogger.onlyAcceptsAHuman( makePerson() ); // <-- yay! It compiles!
+```
+
+# Quality Benefits of this approach
+
+Abstraction layers are a valuable concept in programming. Here's a classic example of why it will help you and your API consumers.
+
+Let's say that `some-external-dependency` changes their method to take a different data structure.
+
+If tsoa were to update your swagger documentation to thew new type, then your API consumers would get 400 BAD REQUEST errors. See here:
```
endor: the

ExternalInterfacesExplanation.MD
```diff
@@ -0,0 +1,166 @@
...
+
+export function doStuff(user: IUser) {
+    // ...
+}
+```
+Now your swagger documentation has changed, but the API consumer hasn't been notified of the breaking contract. So they call the API:
+```ts
+const response = someHttpLib.post('/api/Users/', {
+    name: "Bob Smith"
+}); // Throw 400 error: "firstName is a required parameter
```
endor: `"` missing

src/metadataGeneration/typeResolver.ts
```diff
@@ -432,7 +432,7 @@ export class TypeResolver {
       }) as UsableDeclaration[];
 
     if (!modelTypes.length) {
-      throw new GenerateMetadataError(`No matching model found for referenced type ${typeName}.`);
+      throw new GenerateMetadataError(`No matching model found for referenced type ${typeName}. If ${typeName} comes from a dependency, please create an interface in your own code that has the same structure. Tsoa can not utilize interfaces from external dependencies. Read more at https://github.com/lukeautry/tsoa/blob/master/ExternalInterfacesExplanation.MD`);
```
endor: wonder if the reference to the file should be local instead.. so `../../ExternalInterfacesExplanation.MD` or whatever. just in case the location of the repository changes at some point.

tests/fixtures/controllers/getController.ts
```diff
@@ -40,6 +40,12 @@ export class GetTestController extends Controller {
     modelsArray: new Array<TestSubModel>(),
     numberArray: [1, 2, 3],
     numberValue: 1,
+    object: {
```
endor: are these necessary for the tests to compile?

src/metadataGeneration/typeResolver.ts
dgreene1: I wanted to include the fully qualified link so that people could just copy and paste the link (or ctrl+click in vscode).

As far as the repo changing location at some point, I didn't even know that was possible.

tests/fixtures/controllers/getController.ts
dgreene1: Yea, unfortunately they are necessary for the tests to compile. It happened because one of the previous PRs got merged without the tests running (woops!). But now that the previous PR added husky, these kind of things will be caught.

dgreene1:
@endor I pushed the corrections. Thank you so much for reviewing this so thoroughly. It's wonderful to be collaborating with you. :)

PR #396, opened 2019-08-10
support controllers glob for dynamic imports

### LINKED ISSUE #395: Support dyanimc controllers import glob
Hey!

I think it could nice to add an option to the routes config (later on to the swagger as well) to support controllers glob like RoutingControllers library

Example in the PR

joni7777:
FIX #395

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -150,5 +150,5 @@ export interface RoutesConfig {
   /**
    * Controllers glob path, contains directories with wild card to file names.
    */
-  controllers?: string[];
+  controllerPathGlobs?: string[];
 }
--- a/src/config.ts
+++ b/src/config.ts
@@ -146,4 +146,9 @@ export interface RoutesConfig {
    * Authentication Module for express, hapi and koa
    */
   authenticationModule?: string;
+
+  /**
+   * Controllers glob path, contains directories with wild card to file names.
+   */
+  controllers?: string[];
 }
--- a/src/metadataGeneration/metadataGenerator.ts
+++ b/src/metadataGeneration/metadataGenerator.ts
@@ -1,23 +1,46 @@
 import * as mm from 'minimatch';
 import * as ts from 'typescript';
+import { importClassesFromDirectories } from '../utils/importClassesFromDirectories';
 import { ControllerGenerator } from './controllerGenerator';
 import { Tsoa } from './tsoa';
 
 export class MetadataGenerator {
   public readonly nodes = new Array<ts.Node>();
-  public readonly typeChecker: ts.TypeChecker;
-  private readonly program: ts.Program;
+  public typeChecker: ts.TypeChecker;
+  private program: ts.Program;
   private referenceTypeMap: Tsoa.ReferenceTypeMap = {};
   private circularDependencyResolvers = new Array<(referenceTypes: Tsoa.ReferenceTypeMap) => void>();
 
   public IsExportedNode(node: ts.Node) { return true; }
 
-  constructor(entryFile: string, compilerOptions?: ts.CompilerOptions, private readonly ignorePaths?: string[]) {
+  constructor(entryFile: string, private readonly compilerOptions?: ts.CompilerOptions, private readonly ignorePaths?: string[], private readonly controllers?: string[]) {
     this.program = ts.createProgram([entryFile], compilerOptions || {});
     this.typeChecker = this.program.getTypeChecker();
   }
 
   public Generate(): Tsoa.Metadata {
+    if (this.controllers) {
+      this.setProgramToDaynamicControllersFiles(this.controllers);
+    }
+
+    this.extraceNodeFromProgramSourceFiles();
+
+    const controllers = this.buildControllers();
+
+    this.circularDependencyResolvers.forEach((c) => c(this.referenceTypeMap));
+
+    return {
+      controllers,
+      referenceTypeMap: this.referenceTypeMap,
+    };
+  }
+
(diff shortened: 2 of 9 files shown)


src/config.ts
```diff
@@ -146,4 +146,9 @@ export interface RoutesConfig {
    * Authentication Module for express, hapi and koa
    */
   authenticationModule?: string;
+
+  /**
+   * Controllers glob path, contains directories with wild card to file names.
+   */
+  controllers?: string[];
```
dgreene1: Can you rename this to be `controllerPathGlobs` so that it’s clear to consumers?

src/utils/importClassesFromDirectories.ts
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * Loads all exported classes from the given directory.
+ */
+export function importClassesFromDirectories(directories: string[], formats = ['.ts']): string[] {
+  const allFiles = directories.reduce((allDirs, dir) => {
+    return allDirs.concat(require('glob').sync(path.normalize(dir)));
+  }, [] as string[]);
+
+  return allFiles
```
dgreene1: I don’t see anything in here that checks if zero files were found. I think it’s important that we throw an error when the glob doesn’t find any files because that will inform the user that the glob is probably incorrect. Otherwise, the user will wonder why their routes.ts was generated without any routes.

I’d also like to see an automated unit test or an integration test that confirms this.

tests/unit/swagger/definitionsGeneration/metadata.spec.ts
```diff
@@ -14,6 +14,22 @@ describe('Metadata generation', () => {
     });
   });
 
+  describe('DynamicControllerGenerator', () => {
+    it('should should throw \'globs found 0 controllers.\'', () => {
+      expect(() => {
+        // Non existing controllers folder to get 0 controllers found error
+        const NON_CONTROLLER_EXISTS_GLOB = './tests/unit/swagger';
+        new MetadataGenerator('./tests/fixtures/express-dynamic-controllers/server.ts', undefined, [], [NON_CONTROLLER_EXISTS_GLOB] ).Generate();
+      }).to.throw(/globs found 0 controllers./);
```
dgreene1: Thank you so much for adding this test

tests/unit/swagger/definitionsGeneration/metadata.spec.ts
```diff
@@ -14,6 +14,22 @@ describe('Metadata generation', () => {
...
+    });
+
+    it('should generate one controller', () => {
+      expect(metadata.controllers.length).to.equal(1);
```
dgreene1: It was probably just a slip up, but this second unit test isn’t actually testing the controllers glob feature because it uses the metadata variable at the top of the file. You’d need to start the test off with:
```
const metadataForControllerGlobs = new MetadataGenerator( /* same settings you did in the last test */ );

expect(metadataForControllerGlobs.controllers[0]...
```

src/metadataGeneration/metadataGenerator.ts
```diff
@@ -1,23 +1,52 @@
...
 
-  constructor(entryFile: string, compilerOptions?: ts.CompilerOptions, private readonly ignorePaths?: string[]) {
+  constructor(entryFile: string, private readonly compilerOptions?: ts.CompilerOptions, private readonly ignorePaths?: string[], private readonly controllers?: string[]) {
     this.program = ts.createProgram([entryFile], compilerOptions || {});
     this.typeChecker = this.program.getTypeChecker();
   }
 
   public Generate(): Tsoa.Metadata {
+    if (this.controllers) {
+      this.setProgramToDynamicControllersFiles(this.controllers);
```
dgreene1: It would be awesome if we could leave `typeChecker` and `program` as readonly. I think if you move this check into the controller, then you don’t have make those variables mutable.

dgreene1:
I found a couple of things, but ultimately this is a fantastic contribution. Thank you so much for doing this. :) I’m looking forward to merging this after you fix the one tests and return the variables to being readonly.

joni7777:
> I’m not sure why tsoa would search a file called routes.ts to find a controller. I think you meant `“./custom-filerecursive-dir/**/*.controller*.ts"` instead of `“./custom-filerecursive-dir/**/*.routes.ts"`

you are right, just in our code style we separate the controller to router and controller
in that way we achieve A real SRP with the controller only knowing business logic of the app and router to handle the http requests structure and validations

src/metadataGeneration/metadataGenerator.ts
joni7777: do you mean check this in the c'tor? still need to run all the logic of importClassesFromDirectories to get the classes, and make length check and then to init the program and the type checker 
so if ill move it all to the c'tor it wont be readable

dgreene1: You can achieve both goals of readability and immutability by changing:
`private setProgramToDynamicControllersFiles(controllers): void`
to be a pure function that can be used in the constructor, like:
`private createProgramToDynamicControllersFiles(controllers): ts.Program {`

So the constructor would look like:
```
this.program = !!controllers ?
    createProgramToDynamicControllersFiles(...) :
    ts.createProgram([entryFile], compilerOptions || {});
```

joni7777:
Do you know when this will get published to the npm?

I really need to use it :)

dgreene1:
For those that come accross this:
> Do you know when this will get published to the npm?

This was published in 2.4.7 I believe. I've added release notes. There will be a fix/enhancement coming in 2.4.9

Issue #394, opened 2019-08-10
TSOA unable to ignore constructor parameters


WoH:
When documenting classes, tsoa documents both properties and all constructor parameters not marked private or protected. This leads to issues with classes like this:

```
export class TsoaError {
  /**
   * @ignore
   */
  readonly status: number

  constructor(status: number, public readonly message: string) { }
}
```

When TSOA parses this class, it correctly iterates model properties and filters the ignored property. Then it iterates the constructor parameters, which is perfect since I want to document the message. However, it re-adds the ignored status, since the filter for constructor parameters only ignores parameters explicitly marked `private` or `protected`.

See [here](https://github.com/lukeautry/tsoa/blob/master/src/metadataGeneration/typeResolver.ts#L578) and [here](https://github.com/lukeautry/tsoa/blob/master/src/metadataGeneration/typeResolver.ts#L658)

I would really like to change this to add only explicitly marked public parameters to the metadata. I'd argue he benefit of having the unmarked parameters added to the metadata is not worth the inability to remove them.

I'd happily open the PR :)

dgreene1:
In TypeScript, the default accessor is public if you don’t explicitly define it’s accessor ((citation here)[https://www.typescriptlang.org/docs/handbook/classes.html#public-private-and-protected-modifiers]). So if we were to make the change that you’re requesting, that would be hiding public members. This would be confusing and would likely result in other users filing bugs against tsoa.

So we’ll have to close this issue since it is behaving as expected. Unless you mark a member as private or protected (or use the ignore JSDoc (which I was actually surprised that tsoa had)) then the memeber will be both public in tsoa and the generated swagger.

dgreene1:
But if you would like to submit a PR that fixes the fact that the ignore JSDoc tag isn’t being applied for constructor parameters, we’d accept that.

I just can’t accept this solution you described for the reasons I stated in the prior comment:
> “I would really like to change this to add only explicitly marked public parameters to the metadata.”

WoH:
That is semi-correct. You can parse the modifiers and check.

```
  private hasPublicModifier(node: ts.Node) {
    return !node.modifiers || node.modifiers.every((modifier) => {
      return modifier.kind !== ts.SyntaxKind.ProtectedKeyword && modifier.kind !== ts.SyntaxKind.PrivateKeyword;
    });
  }
```

WoH:
I should clarify, because I didn't explain the issue well I think.
[...]

PR #397, opened 2019-08-10
Ignore unspecified constructor args in type resolution (#394)

WoH:
#394 

Other types of constructor vars are already tested [here](https://github.com/lukeautry/tsoa/blob/master/tests/unit/swagger/definitionsGeneration/definitions.spec.ts#L163).

Class properties are already tested [here](https://github.com/lukeautry/tsoa/blob/master/tests/unit/swagger/definitionsGeneration/definitions.spec.ts#L154).

@dgreene1 I'd appreciate if you could take a look, I think I covered everything we discussed.

### DIFF
--- a/tests/fixtures/testModel.ts
+++ b/tests/fixtures/testModel.ts
@@ -328,6 +328,7 @@ export class TestClassModel extends TestClassBaseModel {
   constructor(
     public publicConstructorVar: string,
     protected protectedConstructorVar: string,
+    defaultConstructorArgument: string,
     public optionalPublicConstructorVar?: string,
   ) {
     super();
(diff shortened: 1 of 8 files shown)


tests/fixtures/testModel.ts
```diff
@@ -328,6 +328,7 @@ export class TestClassModel extends TestClassBaseModel {
   constructor(
     public publicConstructorVar: string,
     protected protectedConstructorVar: string,
+    defaultConstructorArgument: string,
```
WoH: I'd like to prefix this with an underscore, but the linter yells at me if I do :disappointed:

Issue #374, opened 2019-07-25
Support for index signatures


manast-apsis:
Seems like index signatures such as
```
class myClass {
    nodes: {
        [nodeName: string]: Node;
    };
}
```
are not currently supported.

https://www.typescriptlang.org/docs/handbook/advanced-types.html#index-types-and-string-index-signatures

nodes will just show as an empty object in the swagger file.

dgreene1:
I'll look into making a PR for this after https://github.com/lukeautry/tsoa/pull/385 merges (since that's a big PR with lots of changes.

In the meantime, although the swagger output will not contain the correct information, the request validation will ensure that the data is correct. There is an API test that confirms that the following interface actually gets validated.
```ts
export interface ValidateMapStringToNumber {
  [key: string]: number;
}
```
And here's the unit test that proves the validation works as expected: https://github.com/lukeautry/tsoa/blob/a5ee2904ae12ddb9d3951f85304be5f26f4560c9/tests/integration/express-server.spec.ts#L479

So long story short, please continue to use `tsoa` for your validation, and we'll add a PR for better swagger generation for indexed types soon. :)

I know it's a bit strange that tsoa's swagger output is separate from it's runtime validation, but that's advantageous because we return nice, human readable output.

javichi:
Looking forward for this feature :)

dgreene1:
I'm closing this because it's a duplicate of #50. However, until #50 is solved, the workaround for this particular case is:
```ts
interface INodeDictionary {
    [nodeName: string]: Node;
}
class myClass {
    nodes: INodeDictionary;
}
```

dgreene1:
We need to reopen this to make sure that the case presented by @javichi works: https://github.com/lukeautry/tsoa/issues/50#issuecomment-528213658

dgreene1:
@javichi / @manast-apsis this fix/feature was just [released in v2.5.1](https://github.com/lukeautry/tsoa/releases). Can you let us know how it worked out for you?

Side note: we're looking for companies that want to be featured in tsoa's readme: https://github.com/lukeautry/tsoa/issues/464

javichi:
Sorry for the delay. I can confirm that all index signatures are generated in v2.5.1. Great work guys!

Issue #50, opened 2017-01-31
Unknown type: TypeLiteral when using a dictionary on a model object


donahchoo:
Hi,

When I try and generate the model I get this error:

      Unknown type: TypeLiteral

I thought I knew what was causing this, but its clear now that I don't.  Its from one of my models, but that's about it.

Any suggestions?

alexey-pelykh:
I guess this is a common issue for anonymous types. Either a type name needs to be generated, or more meaningful error reported. @lukeautry what's your position on this? I was looking into this for hashmaps support (```field: {[key:string]: xxxxx};```) - may as well fix this if you'll approve type generation

lukeautry:
I think we should be able to process anonymous types, but I don't think we should generate interfaces for them or anything. I think we should simply have inline descriptions in the swagger spec with the properties contained in that anonymous type.

ThorstenBux:
Any update on that?

fhewitt:
+1 This would definitely be a great improvement : Being forced to define distinct classes for properties that will never exist by themselves is painful and add unnecessary complexity to the documentation.

dgreene1:
@ThorstenBux / @fhewitt / @donahchoo / @manast-apsis if `tsoa` were to generate the swagger model "on-the-fly" it doesn't have a name for the model. And that's the problem, swagger requires a name for the "$ref" ([see here](https://swagger.io/docs/specification/using-ref/)). So my question for you is, how would you like it create that name?

An example @manast-apsis provided looks like this:
```ts
class myClass {
    nodes: {
        [nodeName: string]: Node;
    };
}
```

I imagine that we could create something like `RecordOfNodeNameToNodeValue` since that would be `RecordOf{indexName}To{valueType}Value`. But even then, I'm wondering if we would get collisions. For instance:
```ts
export interface ModelToUser {
    toys: {
        [key: string]: any
    },
    chores {
        [key: string]: any
    }
}
```
That would produce to $refs with the same name `RecordOfKeyToAnyValue`.

I'm just raising some scenarios. **Also I'd love your input to know what you'd like the auto-generated name to look like.** I'll add the [Help Wanted] PR to this so new contributors can give it a shot.

---------

P.S. it's been mentioned before, but the workaround is to make explicit interfaces (even if that is unpalatable for some). So in the case of the `MyClass.nodes` issue I presented above, the current solution would be
```ts
interface INodeDictionary {
    [nodeName: string]: Node;
}
class myClass {
[...]

PR #400, opened 2019-08-11
Add support for OAPI3 Union and Intersection Types (fixes #393)

### LINKED ISSUE #393: Seeking contributions to add `anyOf` and `oneOf`
Swagger supports JSON Schema's `anyOf` and `oneOf`.
* `oneOf` is analogous to the `|` (i.e. TypeScript union)
  * Could find inspiration from `typescript-json-schema`'s implementation (but be wary because JSON Schema is **not** equivalent to Swagger/OpenApi): https://github.com/YousefED/typescript-json-schema/blob/4f2d3da6f3f1ae3e1641b6777324c888a21264a3/typescript-json-schema.ts#L220
* `allOf` is analogous to the `&` (i.e. TypeScript intersection/aggregate)
  * Could find inspiration from `typescript-json-schema`'s implementation (but be wary because JSON Schema is **not** equivalent to Swagger/OpenApi): https://github.com/YousefED/typescript-json-schema/blob/4f2d3da6f3f1ae3e1641b6777324c888a21264a3/typescript-json-schema.ts#L660
* Both keywords have separate validation criteria as described here: https://stackoverflow.com/a/34038277/706768

For those people who  need a workaround, please check out: https://github.com/lukeautry/tsoa/issues/391

WoH:
This is a work in progress to fix #393.

I am trying to maintain compatibility with the supported UnionType, because we can translate that to a proper OAPI2 definition meaning the behavior of UnionTypes that are LiteralTypes is still identical. 

Any other case (UnionType and not LiteralType) will still be an unspecified object for OAPI2.
I applied this behavior for IntersectionTypes that are not handled [here](https://github.com/lukeautry/tsoa/blob/5bdde097e55b79ed6cf71d0751b008bdc3445b43/src/metadataGeneration/typeResolver.ts#L535) aswell.

Since the behavior differs, I added OAPI3 definition specs.

I decided to open the PR early because I think it's ready for feedback.


src/swagger/specGenerator2.ts
```diff
@@ -253,6 +253,15 @@ export class SpecGenerator2 extends SpecGenerator {
     return properties;
   }
 
+  protected getSwaggerType(type: Tsoa.Type): Swagger.Schema {
+
+    if (type.dataType === 'union' || type.dataType === 'intersection') {
+      return { type: 'object' };
```
dgreene1: I think we should `console.warn` in here some text that says `Swagger 2.0 does not support intersection or unions. If you would like to take advantage of this, please change tsoa.json's "specVersion" to 3.`

tests/fixtures/testModel.ts
```diff
@@ -66,6 +66,12 @@ export interface TestModel extends Model {
   TypeAliasCase2?: TypeAliasModelCase2;
 }
 
+export interface UnionTestModel {
+  or: TypeAliasModel1 | TypeAliasModel2;
+  and: TypeAliasModel1 & TypeAliasModel2;
+  referenceAnd: TypeAliasModelCase1;
+}
+
```
dgreene1: There are a couple of other cases that users have asked for in the past in tsoa issues.

1) I'd like to see a case here where users try to create an "on-the-fly" union.

So given this interface:
```ts
export interface OnTheFlyUnion {
  twoObjects: { apple: string } |  { orange: string }
}
```
And we'll need some input from other people (@endor?) on what we should be testing for. As a first pass, we'll need to probably throw or warn that you must use interfaces.

Perhaps as a future enhancement we could create a name on the fly, but I don't know if we should start down the road of generating swagger model names for users. Like what would this one even be... `$ref: twoObjectsObjectWithAppleOrObjectWithOrange`? Because that won't scale very well if objects have lots of properties. Perhaps the easiest way would be to just create a UUID since that's unlikely to be a duplicate $ref. So like `$ref: #/definitions/Union913ced83-3873-43bc-ab5f-6a7c22af9181`

Another case that we'll want to test for since people have asked for it:
```ts
export interface IObjectWithUnionOfPrimativeAndObj {
  primOrObj : { apple: string } | number
}
```

and lastly:
```ts
export interface IObjectWithUnionOfNullAndObj {
  primOrNull : null | { apple: string }
}
```

WoH:
Rebased

tests/fixtures/testModel.ts
WoH: Due to the way these things are being handled, all of those cases should not be an issue.
I simply collect the types that are mixed (& or |) and resolve each one individually.
I added some tests to illustrate the behavior. :beers:

WoH: The `primOrNull` will throw (`Error: Unknown type: NullKeyword`) because null individually can not be resolved.
I'm not sure it should be, maybe there is an argument to catch that corner case.

WoH: > Perhaps as a future enhancement we could create a name on the fly, but I don't know if we should start down the road of generating swagger model names for users. Like what would this one even be... $ref: twoObjectsObjectWithAppleOrObjectWithOrange? Because that won't scale very well if objects have lots of properties. Perhaps the easiest way would be to just create a UUID since that's unlikely to be a duplicate $ref. So like $ref: #/definitions/Union913ced83-3873-43bc-ab5f-6a7c22af9181

This is also related to https://github.com/lukeautry/tsoa/pull/400#discussion_r313035787.

> collect the types that are mixed (& or |) and resolve each one individually

This PR will only $ref what it used to $ref individually before and inline what it used to inline before.
If this is a problem, and I don't believe it is (when we $ref, the user has a corresponding interface in his code, which lines up elegantly with the openapi schema), it's something that should be tackled for all of tsoa, not just unions/intersection types.

dgreene1: > The primOrNull will throw (Error: Unknown type: NullKeyword) because null individually can not be resolved.
I'm not sure it should be, maybe there is an argument to catch that corner case.

I think in OpenAPI 3.0 we could check if any of the union types are the null keyword, and if so we could mark it as nullable.

dgreene1:
@WoH You make a good point about inline types versus interface types
> it's something that should be tackled for all of tsoa, not just unions/intersection types.

I’m starting to think that the best thing to do would be to call `console.warn` whenever we find an in-line type. Take for instance issue https://github.com/lukeautry/tsoa/issues/374 where the user asssumed the problem was that he provided an index type. In reality, the issue is that his interface had an “on the fly” object which as you and I know @WoH showed up as an object in Swagger. But if we warned that an interface had a nested type that didn’t correspond to a $ref then we could point the user in the right direction (the right direction being that they should make every type an interface). This solution is much less brittle than enhancing tsoa to generate $ref names. We have precedent from all the way back in issue https://github.com/lukeautry/tsoa/issues/50#issuecomment-287143907 that tsoa chooses not to create names.

WoH:
> I’m starting to think that the best thing to do would be to call console.warn whenever we find an in-line type. Take for instance issue #374 where the user asssumed the problem was that he provided an index type. In reality, the issue is that his interface had an “on the fly” object which as you and I know @WoH showed up as an object in Swagger. But if we warned that an interface had a nested type that didn’t correspond to a $ref then we could point the user in the right direction (the right direction being that they should make every type an interface). This solution is much less brittle than enhancing tsoa to generate $ref names. We have precedent from all the way back in issue #50 (comment) that tsoa chooses not to create names.

Warning the user is a good idea, but I believe it's best to add an unrelated PR that improves the way we deal with these on the fly objects by digging into them and adding known properties. That would directly benefit the way we handle them in interfaces and indirectly this PR because it would automatically reuse it.

WoH:
Added initial validation. Right now, I have to ignore additional prop validation for Intersections.

I am not handling this (yet) because redirecting input like `{value1: true, value2: true}` to `A: { value1: bool } & B { value: 2 }` will complain about excess in A and in B.

tests/fixtures/testModel.ts
WoH: > I think in OpenAPI 3.0 we could check if any of the union types are the null keyword and if so, we could mark it as nullable.

I thought about this and while it sounds nice on paper and is easy to detect, it will be a very big change under the hood because validation will break. There, we currently do not distinguish between `required` and `nullable` properties, so I'd have to add that, check which version we are validating against and ideally add the distinction between `nullable` and `required` for all data types. The tests also seem to rely on the fact that `x-nullable` and `required` are somewhat the same concept for Swagger 2.

Therefore, I think this is more than a convenience and users should rely on the `?` operator for now.
Since `null` throws, everyone knows what's going on so communication works.

We can always follow up with a validation for OAPI3 that does not break and depends on a RouteGenerator that distinguishes between both concepts.

src/routeGeneration/templateHelpers.ts
```diff
@@ -359,6 +363,95 @@ export class ValidationService {
...
+    if (!subSchemas) { return; }
+
+    const subFieldErrors = subSchemas.map(subSchema => {
+      const subFieldErrors: FieldErrors = {};
+      this.ValidateParam(subSchema, value, name, subFieldErrors, parent);
+      return subFieldErrors;
+    });
+
+    subFieldErrors.forEach(subFieldError => {
+      (Object as any).entries(subFieldError).forEach(([key, value]) => {
```
dgreene1: @WoH can you please turn on this TsLint rule? https://palantir.github.io/tslint/rules/no-shadowed-variable/

I ask because it looks like you have a variable named `value` here and then you have `value` as a parameter of `validateIntersection`. For readability purposes, please change the `validateIntersection` property to `dataToValidate` (or something else if you can think of something shorter).

src/routeGeneration/templateHelpers.ts
```diff
@@ -359,6 +363,95 @@ export class ValidationService {
...
+        if (value.message.includes('excess')) {
```
dgreene1: Our code shouldn't rely on "magic strings." I know it's unlikely that the error will change as long as I'm one of the owners of the library, but since this line of code depends on the language of the error message thrown from another place, I think we have two options:
1) have the function that returns that error use some kind of statically typed error (like `type IExcessError = Error & {foundExcessErrors:true}`)
2) or probably the easier option, let's just have a unit test that ensures that this bit of code stays in-tact.

src/routeGeneration/templateHelpers.ts
dgreene1: I think we can't use `.entries` out of the box unless TypeScript is transpiling it down to a version that Node 6 supports ( https://node.green/#ES2017-features-Object-static-methods-Object-entries ). I realize that it's probably ridiculous to support Node 6 at this point, but to be safe, let's add `"es2017"` to the `"lib"` section of tsconfig.json.

tests/fixtures/testModel.ts
dgreene1: > Therefore, I think this (T|null) is more than a convenience and users should rely on the ? operator for now.
I mean it might be an issue for some since null is not undefined. But I'll mark this as resolved because

> Since null throws, everyone knows what's going on so communication works.

Yes, I suppose that's true. I'm sure that some user of tsoa will complain, but we can always enhance this code when/if that user ever files a PR.

src/routeGeneration/templateHelpers.ts
```diff
@@ -359,6 +363,95 @@ export class ValidationService {
...
+
+    const reportedExcess = refNames.map(refName => this.models[refName]).reduce((acc, subSchema) => {
+      return new Set([...acc, ...this.getExcessPropertiesFor(subSchema, new Set(Object.keys(value)))]);
+    }, new Set());
+
+    const allowedProperties = refNames.map(refName => this.models[refName]).reduce((acc, subSchema) => {
+      return new Set([...acc, ...this.getPropertiesFor(subSchema)]);
+    }, new Set());
+
+    const actualExcess = new Set([...reportedExcess].filter(property => !allowedProperties.has(property)));
```
dgreene1: I don't see any of this code using `noImplicitAdditionalProperties` (I'm talking about the boolean version). If `noImplicitAdditionalProperties` is set to `false | undefined` then you don't even need to do the excess property check.

src/routeGeneration/templateHelpers.ts
WoH: > Our code shouldn't rely on "magic strings." I know it's unlikely that the error will change as long as I'm one of the owners of the library, but since this line of code depends on the language of the error message thrown from another place

See above.
> E: Especially the way I ignore them right now is not acceptable. I will refactor the additional prop handling, but it somewhat depends on #411.

I noticed I can call validateModel and set additionalProps to false after that gets merged, which I am intending to use so I don't have to do any filtering at all.

src/routeGeneration/templateHelpers.ts
WoH: In that case, `reportedExcess` will be an empty Set.

src/routeGeneration/templateHelpers.ts
```diff
@@ -359,6 +363,95 @@ export class ValidationService {
...
+      return;
+    }
+
+    // Only Model definitions make sense here right now
+    const refNames = subSchemas.filter(subschema => subschema.ref).map(subschema => subschema.ref) as string[];
+    if (!refNames.every(refName => this.models[refName])) {
+      return value;
+    }
+
+    const reportedExcess = refNames.map(refName => this.models[refName]).reduce((acc, subSchema) => {
```
dgreene1: `allowedProperties` can stay a `Set` because you're using the `Set.prototype.has` method (which is Big-O(1). But it's much slower to have `reportedExcess` be a `Set` since you never use any Big-O(1) operations on it. Additionally when you "unwrap" the Set later via the spread operator, you're iterating over the Set a second unnecessary time.

Please remove any conversions to `Set` where you aren't using `.has`

I've created a JsPerf demonstrating the advantages of not using Set in this circumstance:
https://jsperf.com/reduce-with-new-set-vs-seed/1

WoH: The reason I am using a Set here is to avoid duplication here.

WoH: If there is an actual excess property, every single Model in the intersection will report that.
But I think I can create only one Set at the end.

src/routeGeneration/templateHelpers.ts
dgreene1: @WoH I noticed that this was marked as resolved. Was that by mistake? Shadowing a variable name is an almost universally discouraged practice.

src/routeGeneration/templateHelpers.ts
dgreene1: Sounds good. Since it doesn't look like #411 is going to get much more review anyway, I have merged that so you can continue unencumbered.

src/routeGeneration/templateHelpers.ts
WoH: Yes, because that's exactly the part to be removed because of #411.

WoH:
Should be good for a presumably final round of feedback, I believe I was able to address all concerns

src/swagger/specGenerator3.ts
```diff
@@ -316,4 +316,12 @@ export class SpecGenerator3 extends SpecGenerator {
   protected getSwaggerTypeForReferenceType(referenceType: Tsoa.ReferenceType): Swagger.BaseSchema {
     return { $ref: `#/components/schemas/${referenceType.refName}` };
   }
+
+  protected getSwaggerTypeForUnionType(type: Tsoa.UnionType) {
+    return { anyOf: type.types.map(x => this.getSwaggerType(x))};
```
dgreene1: Why did your last commit change it from oneOf to anyOf? From my understanding of the [swagger documentation](https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/#anyof) anyOf allows for an object to satisfy parts of either. Thats not how a TypeScript union works. You must satisfy all requirements of one of the types in the union. So this needs to be changed back to oneOf.

WoH:
```typescript
let a: {value1: number} | {value2: number} = {value1: 1, value2: 2}
```

Is valid.

dgreene1: It's valid because you're the way you've written it is telling TypeScript that you're asserting that it is a union. I believe you might have uncovered a bug in TypeScript with this example:
```ts
let a: {value1: number} | {value2: number} = {value1: 1, value2: 2}
```

But if you write it out step by step then TypeScript catches it and won't even let you set value2.

See below or in [this playground](https://www.typescriptlang.org/play/index.html#code/JYOwLgpgTgZghgYwgAgJIBkIzAFQJ4AOKA3gFDIXIBucANgK4QCMAXMiPQLYBG0ANKQC+pUJFiIUqAErAA5gAtchEuUo0GEAExsOPaENKkwy5AFEAHnE4FaEAKohgAexDIAvGkzZ8RZAB80GQUlIgBuQwQXAGcwZCgIKPpaMDYLKxt7Rxd3ZDJKajpGVmQAZgF89UY2TWQAelrkHxQAcmICjWLdXihQ9qr2Lm7ewWbkYCj2J1i4KKi5EDhuW2QwJxWTZrTrWwdnEGaAOlV8k5P6gHluACsIBFjaYDE6ZE44PGQXWneoogRgGHeAGsQE4AO6uAhQJxEKBgYAJPjIOAgAAmyGalQgoxRTgSk1iEHM41ioHWvk2lm2mT2hyEQA) I made:
```ts
interface ILeftType {
    value1: number,
}
interface IRightType {
    value2: number
}

type ExampleUnion = ILeftType | IRightType;

const result: ExampleUnion = {
    value1: 3,
    value: 2 // Type '{ value1: number; value: number; }' is not assignable to type 'ExampleUnion'.
                //Object literal may only specify known properties, and 'value' does not exist in type 'ExampleUnion'.
}
```

So yes we have to keep it as `oneOf` and the validation will also have to enforce that the input object can only be all of one of the two objects but not both.

WoH: Reverted in f9ce396

WoH:
@dgreene1 I rebased and would appreciate if you could reconsider your rfcs.

dgreene1:
This is my number 1 “non-day job” priority. But that being said, I have some personal things to take care of. Also, I really want to pull this down and do more testing with some of own personal use cases. I don’t expect anymore requested changes, but I’d feel better after bringing it down. I don’t expect that it will languish more than a week maximum.

tests/prepare.ts
```diff
@@ -58,6 +58,13 @@ const log = async <T>(label: string, fn: () => Promise<T>) => {
       middleware: 'express',
       routesDir: './tests/fixtures/express',
     }, defaultOptions, undefined, undefined, metadata)),
+    log('Express Route Generation, OpenAPI3, noImplicitAdditionalProperties', () => generateRoutes({
+        authenticationModule: './tests/fixtures/express-openapi3/authentication.ts',
+        basePath: '/v1',
+        entryFile: './tests/fixtures/server.ts',
+        middleware: 'express',
+        routesDir: './tests/fixtures/express-openapi3',
+      }, {...optionsWithNoAdditional, specVersion: 3 }, undefined, undefined, metadata)),
```
dgreene1: Wow, did we really not have an integration test with OpenAPI 3?

Good catch. :)

tests/integration/openapi3-express.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+        expect(body.fields['body.stringPatternAZaz'].message).to.equal('Not match in \'^[a-zA-Z]+$\'');
+        expect(body.fields['body.stringPatternAZaz'].value).to.equal(bodyModel.stringPatternAZaz);
+
+        expect(body.fields['body.arrayMax5Item'].message).to.equal('maxItems 5');
+        expect(body.fields['body.arrayMax5Item'].value).to.deep.equal(bodyModel.arrayMax5Item);
+        expect(body.fields['body.arrayMin2Item'].message).to.equal('minItems 2');
+        expect(body.fields['body.arrayMin2Item'].value).to.deep.equal(bodyModel.arrayMin2Item);
+        expect(body.fields['body.arrayUniqueItem'].message).to.equal('required unique array');
+        expect(body.fields['body.arrayUniqueItem'].value).to.deep.equal(bodyModel.arrayUniqueItem);
+        expect(body.fields['body.intersection'].message).to.deep.equal('Could not match the intersection against every type. Issues: [{"body.value2":{"message":"\'value2\' is required"}}]');
```
dgreene1: @WoH these 4 lines are the only assertions I was able to find that verify that the validation code is working. I’d like to see some unit tests that execute the validation logic. I think they would be valuable because you could explain the rationale as to why the assertions should be true. The fact that you can add a “because” string as the second parameter of any Chai assertion is one of the few things that I like about Chai/Mocha (I generally prefer Jest).

So it would help me and future devs a great deal if there were tests like:
```
// pseudo-code
const result = validation(model, data, errorDictionary)

expect(result).to.equal(“something”, “because unions have this special trait that I’m going to explain now”)
```

tests/unit/swagger/definitionsGeneration/definitions.spec.ts
```diff
@@ -333,6 +353,15 @@ describe('Definition generation', () => {
                 genericNestedArrayCharacter2: (propertyName, propertySchema) => {
                     expect(propertySchema.$ref).to.eq('#/definitions/GenericRequestTypeAliasModel2Array', `for property ${propertyName}.$ref`);
                 },
+                and: (propertyName, propertySchema) => {
+                  expect(propertySchema.type).to.eq('object');
```
dgreene1: All of these definition tests should be verifying that `additionalProperties` is set to the expected value. That goes for schemaDetails3.spec.ts too.

tests/unit/swagger/schemaDetails3.spec.ts
```diff
@@ -374,6 +374,27 @@ describe('Definition generation for OpenAPI 3.0.0', () => {
...
+                        ],
+                    });
+                },
+                referenceAnd: (propertyName, propertySchema) => {
+                    expect(propertySchema).to.deep.include({
+                        $ref: '#/components/schemas/TypeAliasModelCase1',
+                    });
+                },
+                or: (propertyName, propertySchema) => {
+                    expect(propertySchema).to.deep.include({
```
dgreene1: Gotta test `additionalProperties` here.

tests/integration/openapi3-express.spec.ts
dgreene1: Another reason why unit tests are valuable for this PR is that you’ll want to test for the various cases for additionalProperties in the validation tests. Off the top of my head, the intersection validation might not work with what I wrote last week for additionalProperties validation. But that’s why we need some exhaustive unit tests.

WoH: I am onto it. But in order to test some stuff, I discovered #432 (trying to verify 'string' | model)

dgreene1: K, 432 has been merged.

tests/unit/swagger/schemaDetails3.spec.ts
WoH: When would `additionalProperties` be set here / be true? :confused:

dgreene1: I’m not sure. But it’s something we have to figure out. Any schema for an object will have it set to true or false. Things that aren’t objects will have it set to undefined. I think the tests above this should provide guidance.

dgreene1: And if it’s undefined for this case, I’d still like to see an assertion that confirms that it’s supposed to be undefined and explains why.

I know that this seems too cautious, but it’s important so that we prevent future regressions.

tests/unit/swagger/definitionsGeneration/definitions.spec.ts
WoH: https://github.com/lukeautry/tsoa/pull/400/commits/62bac808fb811cab908398bf24e637a484c29ce5

tests/unit/swagger/schemaDetails3.spec.ts
WoH: https://github.com/lukeautry/tsoa/pull/400/commits/62bac808fb811cab908398bf24e637a484c29ce5

tests/integration/openapi3-express.spec.ts
WoH: https://github.com/lukeautry/tsoa/pull/400/commits/ce1b89d5e03766c3e06b1a4f456a92d2184556cc

src/routeGeneration/templateHelpers.ts
```diff
@@ -362,6 +366,130 @@ export class ValidationService {
...
+
+  public validateIntersection(
+    name: string,
+    value: any,
+    fieldErrors: FieldErrors,
+    swaggerConfig: SwaggerConfigRelatedToRoutes,
+    subSchemas: TsoaRoute.PropertySchema[] | undefined,
+    parent = '',
+  ): any {
+    if (!subSchemas) { return; }
```
dgreene1: Shouldn't this assert/throw? I ask that because the metadata generation would have had to fail in order for an object to be created that has `dataType === 'union'`  or `dataType === 'union'`  but that doesn't have subSchemas, right?

src/routeGeneration/templateHelpers.ts
```diff
@@ -362,6 +366,130 @@ export class ValidationService {
...
+    }
+
+    const allowedProperties = new Set(refNames.map(refName => this.models[refName]).reduce((acc, subSchema) => {
+      return [...acc, ...this.getPropertiesFor(subSchema)];
+    }, []));
+
+    const actualExcess = [...reportedExcess].filter(property => !allowedProperties.has(property));
+
+    if (actualExcess.length > 0) {
+      fieldErrors[parent + name] = {
```
dgreene1: @WoH we should only set these field errors if the tsoa user has configured it to throw. So I believe that this should be replaced with:
```ts
    if (actualExcess.length > 0) {
      if (swaggerConfig.noImplicitAdditionalProperties === 'throw-on-extras' || swaggerConfig.noImplicitAdditionalProperties === true){
        fieldErrors[parent + name] = {
          message: `The following properties are not allowed by any part of the intersection: ${actualExcess.join(', ')}`,
          value,
        };
      } else if (swaggerConfig.noImplicitAdditionalProperties === 'silently-remove-extras'){
        actualExcess.forEach(excessKey => {
          delete value[excessKey];
        });
      } else if (swaggerConfig.noImplicitAdditionalProperties === false || !swaggerConfig.noImplicitAdditionalProperties){
        // then excess properties are acceptable
      } else {
        assertNever(swaggerConfig.noImplicitAdditionalProperties)
      }
    }
```

tests/integration/openapi3-express.spec.ts
dgreene1: Thank you for adding integrative tests. I should have been more clear that I was hoping for true unit tests that test one function at a time. So I've added a couple of unit tests right here in this comment. Please add these to `templateHelpers.spec.ts`.

```ts
        it('should allow additionalProperties (on a union) if noImplicitAdditionalProperties is set to silently-remove-extras', () => {
            // Arrange
            const refName = 'ExampleModel';
            const subSchemas: TsoaRoute.PropertySchema[] = [
                { "ref": "TypeAliasModel1" },
                { "ref": "TypeAliasModel2" }
            ]
            const models: TsoaRoute.Models = {
                [refName]:{
                    properties: {
                        or: {
                            "dataType": "union",
                            "subSchemas": subSchemas,
                            "required": true
                        },
                    }
                },
                "TypeAliasModel1": {
                    "properties": {
                      "value1": { "dataType": "string", "required": true },
                    },
                    "additionalProperties": false,
                },
                "TypeAliasModel2": {
                    "properties": {
                      "value2": { "dataType": "string", "required": true },
                    },
                    "additionalProperties": false,
                },
            }
            const v = new ValidationService(models);
            const minimalSwaggerConfig: SwaggerConfigRelatedToRoutes = {
                noImplicitAdditionalProperties: 'silently-remove-extras',
            };
            const errorDictionary: FieldErrors = {};
            const nameOfAdditionalProperty = 'I am the bad key name';
            let dataToValidate: {or: TypeAliasModel1} = {
                or: {
                    value1: "this is value 1"
                }
            };
            (dataToValidate.or as any)[nameOfAdditionalProperty]= 'something extra';

            // Act
            const name = 'dataToValidate';
            const result = v.validateUnion('or', dataToValidate, errorDictionary, minimalSwaggerConfig, subSchemas, name+".");

            // Assert
            expect(errorDictionary).to.deep.eq({});
            expect(result).to.eql({
                or: {
                    value1: "this is value 1",
                    [nameOfAdditionalProperty]: 'something extra',
                }
            });
            if (dataToValidate[nameOfAdditionalProperty]) {
                throw new Error(`dataToValidate.${nameOfAdditionalProperty} should have been removed because "silently-remove-extras" requires that excess properties be stripped.`);
            }
        });

        it('should throw if the data has additionalProperties (on a union) if noImplicitAdditionalProperties is set to throw-on-extras', () => {
            // Arrange
            const refName = 'ExampleModel';
            const subSchemas: TsoaRoute.PropertySchema[] = [
                { "ref": "TypeAliasModel1" },
                { "ref": "TypeAliasModel2" }
            ]
            const models: TsoaRoute.Models = {
                [refName]:{
                    properties: {
                        or: {
                            "dataType": "union",
                            "subSchemas": subSchemas,
                            "required": true
                        },
                    }
                },
                "TypeAliasModel1": {
                    "properties": {
                      "value1": { "dataType": "string", "required": true },
                    },
                    "additionalProperties": false,
                },
                "TypeAliasModel2": {
                    "properties": {
                      "value2": { "dataType": "string", "required": true },
                    },
                    "additionalProperties": false,
                },
            }
            const v = new ValidationService(models);
            const minimalSwaggerConfig: SwaggerConfigRelatedToRoutes = {
                noImplicitAdditionalProperties: 'throw-on-extras',
            };
            const errorDictionary: FieldErrors = {};
            const nameOfAdditionalProperty = 'I am the bad key name';
            let dataToValidate: TypeAliasModel1 = {
                value1: "valueOne"
            };
            (dataToValidate as any)[nameOfAdditionalProperty]= 'something extra';

            // Act
            const name = 'dataToValidate';
            v.validateUnion('or', dataToValidate, errorDictionary, minimalSwaggerConfig, subSchemas, name+".");

            // Assert
            const errorKeys = Object.keys(errorDictionary);
            expect(errorKeys).to.have.lengthOf(1);
            const firstAndOnlyErrorKey = errorKeys[0];
            expect(errorDictionary[firstAndOnlyErrorKey].message).to.include(`Could not match the union against any of the items.`);
            expect(errorDictionary[firstAndOnlyErrorKey].message).to.include(nameOfAdditionalProperty);
            expect(errorDictionary[firstAndOnlyErrorKey].message).to.include(`is an excess property and therefore is not allowed`);
            if (!dataToValidate[nameOfAdditionalProperty]) {
                throw new Error(`dataToValidate.${nameOfAdditionalProperty} should have been there because .validateModel should NOT have removed it since it took the more severe option of producing an error instead.`);
            }
        });

        it('should throw if the data has additionalProperties (on a intersection) if noImplicitAdditionalProperties is set to throw-on-extras', () => {
            // Arrange
            const refName = 'ExampleModel';
            const subSchemas: TsoaRoute.PropertySchema[] = [
                { "ref": "TypeAliasModel1" },
                { "ref": "TypeAliasModel2" }
            ]
            const models: TsoaRoute.Models = {
                [refName]:{
                    properties: {
                        and: {
                            "dataType": "intersection",
                            "subSchemas": subSchemas,
                            "required": true
                        },
                    }
                },
                "TypeAliasModel1": {
                    "properties": {
                      "value1": { "dataType": "string", "required": true },
                    },
                    "additionalProperties": false,
                },
                "TypeAliasModel2": {
                    "properties": {
                      "value2": { "dataType": "string", "required": true },
                    },
                    "additionalProperties": false,
                },
            }
            const v = new ValidationService(models);
            const minimalSwaggerConfig: SwaggerConfigRelatedToRoutes = {
                noImplicitAdditionalProperties: 'throw-on-extras',
            };
            const errorDictionary: FieldErrors = {};
            const nameOfAdditionalProperty = 'extraKeyName';
            const expectedErrorMsg = `The following properties are not allowed by any part of the intersection: ${nameOfAdditionalProperty}`;
            let dataToValidate: TypeAliasModel1 & TypeAliasModel2 = {
                value1: "this is value 1",
                value2: "this is value 2"
            };
            (dataToValidate as any)[nameOfAdditionalProperty]= 'something extra';

            // Act
            const name = 'dataToValidate';
            v.validateIntersection('and', dataToValidate, errorDictionary, minimalSwaggerConfig, subSchemas, name+".");

            // Assert
            const errorKeys = Object.keys(errorDictionary);
            expect(errorKeys).to.have.lengthOf(1);
            const firstAndOnlyErrorKey = errorKeys[0];
            expect(errorDictionary[firstAndOnlyErrorKey].message).to.eq(expectedErrorMsg);
            if (!dataToValidate[nameOfAdditionalProperty]) {
                throw new Error(`dataToValidate.${nameOfAdditionalProperty} should have been there because .validateModel should NOT have removed it since it took the more severe option of producing an error instead.`);
            }
        });

        it('should not throw if the data has additionalProperties (on a intersection) if noImplicitAdditionalProperties is set to silently-remove-extras', () => {
            // Arrange
            const refName = 'ExampleModel';
            const subSchemas: TsoaRoute.PropertySchema[] = [
                { "ref": "TypeAliasModel1" },
                { "ref": "TypeAliasModel2" }
            ]
            const models: TsoaRoute.Models = {
                [refName]:{
                    properties: {
                        and: {
                            "dataType": "intersection",
                            "subSchemas": subSchemas,
                            "required": true
                        },
                    }
                },
                "TypeAliasModel1": {
                    "properties": {
                      "value1": { "dataType": "string", "required": true },
                    },
                    "additionalProperties": false,
                },
                "TypeAliasModel2": {
                    "properties": {
                      "value2": { "dataType": "string", "required": true },
                    },
                    "additionalProperties": false,
                },
            }
            const v = new ValidationService(models);
            const minimalSwaggerConfig: SwaggerConfigRelatedToRoutes = {
                noImplicitAdditionalProperties: 'silently-remove-extras',
            };
            const errorDictionary: FieldErrors = {};
            const nameOfAdditionalProperty = 'extraKeyName';
            let dataToValidate: TypeAliasModel1 & TypeAliasModel2 = {
                value1: "this is value 1",
                value2: "this is value 2"
            };
            (dataToValidate as any)[nameOfAdditionalProperty]= 'something extra';

            // Act
            const name = 'dataToValidate';
            const result = v.validateIntersection('and', dataToValidate, errorDictionary, minimalSwaggerConfig, subSchemas, name+".");

            // Assert
            expect(errorDictionary).to.deep.eq({});
            if (dataToValidate[nameOfAdditionalProperty]) {
                throw new Error(`dataToValidate.${nameOfAdditionalProperty} should have been removed because "silently-remove-extras" requires that excess properties be stripped.`);
            }
            expect(result).to.eql({
                value1: "this is value 1",
                value2: "this is value 2",
            });
        });

        it('should not throw if the data has additionalProperties (on a intersection) if noImplicitAdditionalProperties is set to false', () => {
            // Arrange
            const refName = 'ExampleModel';
            const subSchemas: TsoaRoute.PropertySchema[] = [
                { "ref": "TypeAliasModel1" },
                { "ref": "TypeAliasModel2" }
            ]
            const models: TsoaRoute.Models = {
                [refName]:{
                    properties: {
                        and: {
                            "dataType": "intersection",
                            "subSchemas": subSchemas,
                            "required": true
                        },
                    }
                },
                "TypeAliasModel1": {
                    "properties": {
                      "value1": { "dataType": "string", "required": true },
                    },
                    "additionalProperties": false,
                },
                "TypeAliasModel2": {
                    "properties": {
                      "value2": { "dataType": "string", "required": true },
                    },
                    "additionalProperties": false,
                },
            }
            const v = new ValidationService(models);
            const minimalSwaggerConfig: SwaggerConfigRelatedToRoutes = {
                noImplicitAdditionalProperties: false,
            };
            const errorDictionary: FieldErrors = {};
            const nameOfAdditionalProperty = 'extraKeyName';
            let dataToValidate: TypeAliasModel1 & TypeAliasModel2 = {
                value1: "this is value 1",
                value2: "this is value 2"
            };
            (dataToValidate as any)[nameOfAdditionalProperty]= 'something extra';

            // Act
            const name = 'dataToValidate';
            const result = v.validateIntersection('and', dataToValidate, errorDictionary, minimalSwaggerConfig, subSchemas, name+".");

            // Assert
            expect(errorDictionary).to.deep.eq({});
            expect(result).to.eql({
                value1: "this is value 1",
                value2: "this is value 2",
                [nameOfAdditionalProperty]: 'something extra',
            });
        });
```

1 of those tests won't pass. I think it's because it's incredibly challenging to match an object against a union of objects. So I might be okay with changing `should allow additionalProperties (on a union) if noImplicitAdditionalProperties is set to silently-remove-extras` to instead assert `should not allow additionalProperties (on a union) regardless of the noImplicitAdditionalProperties config due to the complexity of unions`.

Two of the other tests won't pass until you fix this other part that I gave the solution for in https://github.com/lukeautry/tsoa/pull/400/files#r316392569

After these unit tests are added and the other comments I've made are resolved, we can merge this PR. :)

src/routeGeneration/templateHelpers.ts
```diff
@@ -362,6 +366,130 @@ export class ValidationService {
...
 
+  public validateUnion(
+    name: string,
+    value: any,
+    fieldErrors: FieldErrors,
+    minimalSwaggerConfig: SwaggerConfigRelatedToRoutes,
+    subSchemas: TsoaRoute.PropertySchema[] | undefined,
+    parent = '',
+  ): any {
+    if (!subSchemas) { return; }
```
dgreene1: I have the same question I have here as I did in comment https://github.com/lukeautry/tsoa/pull/400/files#r316279032

src/routeGeneration/templateHelpers.ts
WoH: If there is reported excess, that means the user has already set it to throw.
See [here](https://github.com/lukeautry/tsoa/pull/400/files#diff-054445345683aa77f57a22be4b3d8b28R473), which is the helper that determines reported Excess for one object/model.
If the user sets 'remove', the value return from model checking will be stripped from additional props -> nothing else to do
If no setting -> reported excess is [] -> actual excess is [].

dgreene1: I don’t understand your response because the unit tests that I provided for you don’t pass unless that code is in there.

src/routeGeneration/templateHelpers.ts
WoH: This will not ever happen, correct. However, TypeScript does not know that.

src/routeGeneration/templateHelpers.ts
WoH: The setup of one specific test was incorrect. It should've been passing `{ value1: '' }`, not `{ or: {value1: '' }}`.
On the silently removing stuff, yes, union did that already, intersection now does the same logical setup. Therefore the removing excess works aswell.

WoH: Also, please note: You verified vs. the dataToValidate, not the result in some cases where that really matters. The result is correct.

WoH:
Added the (adapted) tests as suggested (I usually allow push for maintainers, you don't even have to ask :wink: ).

src/routeGeneration/templateHelpers.ts
WoH: Is there a way you would prefer to handle that? Throw a (specific) error?

dgreene1: I was thinking something like:
```ts
throw new Error('internal tsoa error: the metadata that was generated should have had sub schemas since it’s for a intersection, however it did not. Please file an issue with tsoa at https://github.com/lukeautry/tsoa/issues’);
```
And then a similar one for in validateUnion.

The alternative is to use a [typeGuard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards) outside of the function that narrows the type of the metadata to a type that definitely had subSchemas. But that’s a lot more work. I think throwing the error is fine since you pointed out that it will never be hit.

src/routeGeneration/templateHelpers.ts
```diff
@@ -362,6 +366,155 @@ export class ValidationService {
...
+  ): string[] {
+    if (!modelDefinition || !modelDefinition.properties) {
+      return properties;
+    }
+
+    const modelProperties = new Set(Object.keys(modelDefinition.properties));
+
+    if (modelDefinition.additionalProperties) {
+      return [];
+    } else if (!config.noImplicitAdditionalProperties) {
```
dgreene1: It took me a while to figure out how it was possible that the unit test would even pass, and it turns out that it's because of this line. Please change this to exhaustively check each of the members of the union so that the typescript compiler will inform us if a new member is added to the `noImplicitAdditionalProperties` union and if we haven't considered it here. And if that seems far fetched, consider that you asked for their to be an `"ignore"` option which would you would not want to return `[]` for.

So the code should look like this:
```ts
    if (modelDefinition.additionalProperties) {
      return [];
    } else {
      if (
        config.noImplicitAdditionalProperties === false ||
        config.noImplicitAdditionalProperties === undefined
      ) {
        return [];
      } else if (
        config.noImplicitAdditionalProperties === 'throw-on-extras' ||
        config.noImplicitAdditionalProperties === true ||
        config.noImplicitAdditionalProperties === 'silently-remove-extras'
      ) {
        return [...properties].filter(property => !modelProperties.has(property));
      } else {
        assertNever(config.noImplicitAdditionalProperties);
      }
    }
```

And sorry, I said that I wouldn't make any more corrections. But if you are going to get in the code to fix this other (admittedly minor) request https://github.com/lukeautry/tsoa/pull/400/files#r316668417, then maybe you can make this exhaustiveness checking change too.

WoH: Ok, so there are a few notes:

a) `!config.noImplicitAdditionalProperties` is equal to 
```
config.noImplicitAdditionalProperties === false ||
config.noImplicitAdditionalProperties === undefined
```

b) the `else if` should only catch `throw` and `true`, I'll push a commit to reuse the `resolveAdditionalProps` helper I added.

c) the `else if` can't be `silently remove` because if this were true the validation would've already returned [here](https://github.com/lukeautry/tsoa/pull/400/files#diff-054445345683aa77f57a22be4b3d8b28R400) or [here](https://github.com/lukeautry/tsoa/pull/400/files#diff-054445345683aa77f57a22be4b3d8b28R449).
However, factoring in b) since now the `resolveAdditionalPropSetting` helper is [used](https://github.com/lukeautry/tsoa/pull/400/commits/51d50e0917859b48be5cf2ee75617709fd8f7ffa), this branch can only be `silently-remove` or `throw` (due to exhaustiveness of `resolveAdditionalPropSetting`), which is equivalent to the code you suggested.

src/routeGeneration/templateHelpers.ts
WoH: https://github.com/lukeautry/tsoa/pull/400/commits/fb1f5231978893c73b3fcdd22118254f4bbf209c

src/routeGeneration/templateHelpers.ts
WoH: https://github.com/lukeautry/tsoa/pull/400/commits/fb1f5231978893c73b3fcdd22118254f4bbf209c

src/routeGeneration/templateHelpers.ts
dgreene1: > !config.noImplicitAdditionalProperties is equal to
I know that. It's just not very "scan-able." Your newest push is much easier to read and to discern the meaning. Thank you.

dgreene1:
Thank you @WoH for your impressive work on this PR. And thank you for working with me to make sure that the first release of the feature should avoid any surprise bug reports from users. :)

src/swagger/specGenerator3.ts
WoH: FYI: It actually seems like this was a bug that's fixed as of [3.5](https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#improved-excess-property-checks-in-union-types):

Issue #389, opened 2019-08-07
Autogenerated swagger.json Comes with Errors as per OpenApi 3 spec


luli0822:
Hello,

I am using tsoa that is autogenerating swagger.json. But it failed because of some errors in the browser console. I tried it in [editor.swagger.io](https://editor.swagger.io/), and it reported like the screenshot attached here. 
Am I missing something somewhere?
Also, how do we verify api with headers and params, e.g. Authorization, and others?
Thanks and appreciate any advice and feedbacks.

![image](https://user-images.githubusercontent.com/11763452/62630582-365fde80-b8fd-11e9-8c3f-a75afb53c6e0.png)

The __swagger.json__ is:
```
    {

    "components": {
        "examples": {},
        "headers": {},
        "parameters": {},
        "requestBodies": {},
        "responses": {},
        "schemas": {
            "ApiResponse": {
                "description": "Swagger API response",
                "properties": {
                    "message": {
                        "type": "string"
                    },
                    "status": {
                        "type": "object"
                    },
                    "data": {
                        "type": "object",
                        "nullable": true
                    }
                },
                "required": [
                    "message",
                    "status"
                ],
                "type": "object"
            },

            "UserResponse": {
                "properties": {
                    "data": {
                        "type": "object"
                    },
                    "link": {
                        "type": "string",
                        "nullable": true
                    }
                },
                "required": [
                    "data"
                ],
                "type": "object"
            },

            "ContactRequest": {
                "description": "Request body to create Contact",
                "properties": {
                    "firstName": {
                        "type": "string"
                    },
                    "lastName": {
                        "type": "string"
                    },
                    "email": {
                        "type": "string",
                        "nullable": true
                    },
                    "phoneNumber": {
                        "type": "string"
                    }
                },
                "required": [
                    "firstName",
                    "lastName",
                    "phoneNumber"
                ],
                "type": "object"
            },

            "AddressRequest": {
                "properties": {
                    "addressLine1": {
                        "type": "string"
                    },
                    "addressLine2": {
                        "type": "string",
                        "nullable": true
                    },
                    "province": {
                        "type": "string"
                    },
                    "city": {
                        "type": "string"
                    },
                    "country": {
                        "type": "string"
                    },
                    "postCode": {
                        "type": "string"
                    }
                },
                "required": [
                    "addressLine1",
                    "province",
                    "city",
                    "country",
                    "postCode"
                ],
                "type": "object"
            },

            "EntityType": {
                "enum": [
[...]

PR #401, opened 2019-08-11
OpenAPI 3: Generate valid operations/parameters

WoH:
This PR should ensure we generate a valid spec for the GetController in the tests.

I moved the schema and example to the correct position to ensure we mostly* comply with the specification.

*There are still $refs that should be allowed in 3.1, so I don't see a reason to fix these.

### DIFF
--- a/src/swagger/specGenerator.ts
+++ b/src/swagger/specGenerator.ts
@@ -10,28 +10,6 @@ export class SpecGenerator {
     return this.getSwaggerType(type);
   }
 
-  protected buildOperation(controllerName: string, method: Tsoa.Method): Swagger.Operation {
-    const swaggerResponses: any = {};
-
-    method.responses.forEach((res: Tsoa.Response) => {
-      swaggerResponses[res.name] = {
-        description: res.description,
-      };
-      if (res.schema && res.schema.dataType !== 'void') {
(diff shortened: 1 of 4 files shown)


dgreene1:
@WoH there have been a lot of changes to the master branch. I can see that Github doesn't appear to see an issue, but that's because we dont (yet) have the unit tests running here. So can you please update this branch with master and then fix any breaking tests.

Additionally, would you be willing to see if this PR fixes this [other open issue](https://github.com/lukeautry/tsoa/issues/389) https://github.com/lukeautry/tsoa/issues/389 with OpenAPI 3.0?

WoH:
I saw that,  I rebased from https://github.com/lukeautry/tsoa/commit/7f96d8d9a02329b43bbde11f0e69a0bc97bc8c3e to 
https://github.com/lukeautry/tsoa/commit/465113f51c94f95f490287f5dc6beb1210130041 to address the changes.

WoH:
I ran the output through validation, these are 3 types of errors in the output from #389:

- Property produces is not allowed.
- Property schema is not allowed.
- Property type not allowed.

All 3 of these issues should be handled.
Maybe @luli0822 wants to run his code against this PR ;)

PR #411, opened 2019-08-12
Improve "noImplicitAdditionalProperties" to allow for flexibility

dgreene1:
I showed v2.4.5 to friends at work and the asked if there was an option to accept the request but to safely discard the additional properties. In fact, they told me that it was their belief that it was generally more preferable to strip the additional properties. They said that it would be burdensome to consumers if the consumer had to always watch what they were sending. They cited the "Robustness Principle":

> Be conservative in what you do, be liberal in what you accept from others.

Personally, I can't decide if it's better to throw or to fix the object, so I'm providing both options.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -26,6 +26,11 @@ export interface Config {
   compilerOptions?: ts.CompilerOptions;
 }
 
+/**
+ * these options will be removed in a future version since we would prefer consumers to explicitly state their preference that the tsoa validation throws or removes additional properties
+ */
+export type DeprecatedOptionForAdditionalPropertiesHandling = true | false;
+
 export interface SwaggerConfig {
   /**
    * Generated SwaggerConfig.json will output here
@@ -40,7 +45,7 @@ export interface SwaggerConfig {
   /**
    * Modes that allow you to prevent input data from entering into your API. This will document your decision in the swagger.yaml and it will turn on excess-property validation (at runtime) in your routes.
    */
-  noImplicitAdditionalProperties?: 'throw-on-extras' | 'silently-remove-extras';
+  noImplicitAdditionalProperties?: 'throw-on-extras' | 'silently-remove-extras' | DeprecatedOptionForAdditionalPropertiesHandling;
 
   /**
    * API host, expressTemplate.g. localhost:3000 or myapi.com
--- a/src/config.ts
+++ b/src/config.ts
@@ -38,9 +38,9 @@ export interface SwaggerConfig {
   entryFile: string;
 
   /**
-   * Set this to true if you want to prevent requests from coming in with more properties than are known/allowed
+   * Modes that allow you to prevent input data from entering into your API. This will document your decision in the swagger.yaml and it will turn on excess-property validation (at runtime) in your routes.
    */
-  noImplicitAdditionalProperties?: boolean;
+  noImplicitAdditionalProperties?: 'throw-on-extras' | 'silently-remove-extras';
 
   /**
    * API host, expressTemplate.g. localhost:3000 or myapi.com
--- a/src/module/generate-routes.ts
+++ b/src/module/generate-routes.ts
@@ -1,13 +1,13 @@
 import * as path from 'path';
 import * as ts from 'typescript';
-import { RoutesConfig, SwaggerConfig } from '../config';
+import { RoutesConfig } from '../config';
 import { MetadataGenerator } from '../metadataGeneration/metadataGenerator';
 import { Tsoa } from '../metadataGeneration/tsoa';
-import { RouteGenerator } from '../routeGeneration/routeGenerator';
+import { RouteGenerator, SwaggerConfigRelatedToRoutes } from '../routeGeneration/routeGenerator';
 
 export const generateRoutes = async (
   routesConfig: RoutesConfig,
-  swaggerConfig: SwaggerConfig,
+  minimalSwaggerConfig: SwaggerConfigRelatedToRoutes,
   compilerOptions?: ts.CompilerOptions,
   ignorePaths?: string[],
   /**
@@ -24,7 +24,7 @@ export const generateRoutes = async (
     ).Generate();
   }
 
-  const routeGenerator = new RouteGenerator(metadata, routesConfig, swaggerConfig);
+  const routeGenerator = new RouteGenerator(metadata, routesConfig, exactly(minimalSwaggerConfig));
 
   let pathTransformer;
   let template;
@@ -52,4 +52,29 @@ export const generateRoutes = async (
   await routeGenerator.GenerateCustomRoutes(template, pathTransformer);
 
   return metadata;
+};
+
+const exactly = (input: SwaggerConfigRelatedToRoutes): SwaggerConfigRelatedToRoutes => {
+  // Validate the config values first
+  if (
+    input.noImplicitAdditionalProperties === undefined ||
+    input.noImplicitAdditionalProperties === 'throw-on-extras' ||
+    input.noImplicitAdditionalProperties === 'silently-remove-extras'
+  ) {
+    // then it's good to go
+  } else {
+    throw new Error(`noImplicitAdditionalProperties is set to an invalid value. See https://github.com/lukeautry/tsoa/blob/master/src/config.ts for available options.`);
+  }
+
+  // Make an exact copy that doesn't have other properties
+  const recordOfProps: Record<keyof SwaggerConfigRelatedToRoutes, 'right side does not matter'> = {
+    noImplicitAdditionalProperties: 'right side does not matter',
+  };
+
+  const exactObj: SwaggerConfigRelatedToRoutes = {};
+  Object.keys(recordOfProps).forEach((key) => {
+    const strictKey = key as keyof SwaggerConfigRelatedToRoutes;
+    exactObj[strictKey] = input[strictKey];
+  });
+  return exactObj;
 };
--- a/src/module/generate-routes.ts
+++ b/src/module/generate-routes.ts
@@ -4,6 +4,7 @@ import { RoutesConfig } from '../config';
 import { MetadataGenerator } from '../metadataGeneration/metadataGenerator';
 import { Tsoa } from '../metadataGeneration/tsoa';
 import { RouteGenerator, SwaggerConfigRelatedToRoutes } from '../routeGeneration/routeGenerator';
+import { warnAdditionalPropertiesDeprecation } from '../utils/deprecations';
 
 export const generateRoutes = async (
   routesConfig: RoutesConfig,
@@ -56,7 +57,11 @@ export const generateRoutes = async (
 
 const exactly = (input: SwaggerConfigRelatedToRoutes): SwaggerConfigRelatedToRoutes => {
   // Validate the config values first
-  if (
+  if (input.noImplicitAdditionalProperties === true) {
+    warnAdditionalPropertiesDeprecation(input.noImplicitAdditionalProperties);
+  } else if (input.noImplicitAdditionalProperties === false) {
+    warnAdditionalPropertiesDeprecation(input.noImplicitAdditionalProperties);
+  } else if (
     input.noImplicitAdditionalProperties === undefined ||
     input.noImplicitAdditionalProperties === 'throw-on-extras' ||
     input.noImplicitAdditionalProperties === 'silently-remove-extras'
--- a/src/routeGeneration/routeGenerator.ts
+++ b/src/routeGeneration/routeGenerator.ts
@@ -3,6 +3,7 @@ import * as path from 'path';
 import * as tsfmt from 'typescript-formatter';
 import { Tsoa } from '../metadataGeneration/tsoa';
 import { assertNever } from '../utils/assertNever';
+import { warnAdditionalPropertiesDeprecation } from '../utils/deprecations';
 import { fsReadFile, fsWriteFile } from '../utils/fs';
 import { RoutesConfig, SwaggerConfig } from './../config';
 import { normalisePath } from './../utils/pathUtils';
@@ -55,6 +56,12 @@ export class RouteGenerator {
         return JSON.stringify(false);
       } else if (this.minimalSwaggerConfig.noImplicitAdditionalProperties === undefined) {
         // Since Swagger defaults to allowing additional properties, then that will be our default
+        return JSON.stringify(true);
+      } else if (this.minimalSwaggerConfig.noImplicitAdditionalProperties === true) {
+        warnAdditionalPropertiesDeprecation(this.minimalSwaggerConfig.noImplicitAdditionalProperties);
+        return JSON.stringify(false);
+      } else if (this.minimalSwaggerConfig.noImplicitAdditionalProperties === false) {
+        warnAdditionalPropertiesDeprecation(this.minimalSwaggerConfig.noImplicitAdditionalProperties);
         return JSON.stringify(true);
       } else {
         return assertNever(this.minimalSwaggerConfig.noImplicitAdditionalProperties);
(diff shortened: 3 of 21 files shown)


src/module/generate-routes.ts
```diff
@@ -53,3 +53,28 @@ export const generateRoutes = async (
...
+const exactly = (input: SwaggerConfigRelatedToRoutes): SwaggerConfigRelatedToRoutes => {
+  // Validate the config values first
+  if (
+    input.noImplicitAdditionalProperties === undefined ||
+    input.noImplicitAdditionalProperties === 'throw-on-extras' ||
+    input.noImplicitAdditionalProperties === 'silently-remove-extras'
+  ) {
+    // then it's good to go
+  } else {
+    throw new Error(`noImplicitAdditionalProperties is set to an invalid value. See https://github.com/lukeautry/tsoa/blob/master/src/config.ts for available options.`);
```
dgreene1: @endor I suppose it is a "breaking change" that I'm no longer accepting `true` or `false` as possible values, but:
(a) the code got confusing and gross if I were to allow booleans
(b) I couldn't determine which of the two "modes" (`'throw-on-extras'` or `'silently-remove-extras'`) would `true` default to?
(c) I felt that since the 2.4.4 release just went out this morning that it would be okay to make a breaking change when it's probably very unlikely that anyone has installed the new version and utilized this un-advertised feature. That being said, if they were to upgrade to this code they would at least get this (hopefully) very helpful error message.

src/module/generate-routes.ts
```diff
@@ -53,3 +53,28 @@ export const generateRoutes = async (
...
+  // Make an exact copy that doesn't have other properties
+  const recordOfProps: Record<keyof SwaggerConfigRelatedToRoutes, 'right side does not matter'> = {
+    noImplicitAdditionalProperties: 'right side does not matter',
+  };
+
+  const exactObj: SwaggerConfigRelatedToRoutes = {};
+  Object.keys(recordOfProps).forEach((key) => {
+    const strictKey = key as keyof SwaggerConfigRelatedToRoutes;
+    exactObj[strictKey] = input[strictKey];
+  });
```
dgreene1: I know this is very "meta" but I wanted to explain why I'm stripping the config object. The reason why I'm doing this is because it is possible that people could pass in more values since the CLI and JSON are not typesafe. So this method prunes the config object to only what's allowed. This is necessary since we plop the object into the handlebars template and it shows up as an object literal which is subject to what TypeScript calls ["excess property checks."](https://www.typescriptlang.org/docs/handbook/interfaces.html#excess-property-checks)

src/routeGeneration/routeGenerator.ts
```diff
@@ -48,11 +49,15 @@ export class RouteGenerator {
...
+      } else if (this.minimalSwaggerConfig.noImplicitAdditionalProperties === 'silently-remove-extras') {
         return JSON.stringify(false);
-      } else {
+      } else if (this.minimalSwaggerConfig.noImplicitAdditionalProperties === 'throw-on-extras') {
+        return JSON.stringify(false);
+      } else if (this.minimalSwaggerConfig.noImplicitAdditionalProperties === undefined) {
         // Since Swagger defaults to allowing additional properties, then that will be our default
         return JSON.stringify(true);
+      } else {
+        return assertNever(this.minimalSwaggerConfig.noImplicitAdditionalProperties);
```
dgreene1: I'm using "exhaustiveness checking" so that if we ever add new options to `noImplicitAdditionalProperties` we can get that awareness at compile-time.

src/swagger/specGenerator2.ts
```diff
@@ -67,12 +66,7 @@ export class SpecGenerator2 extends SpecGenerator {
         } else {
             // Since additionalProperties was not explicitly set in the TypeScript interface for this model
             //      ...we need to make a decision
-            if (config.noImplicitAdditionalProperties) {
-                definitions[referenceType.refName].additionalProperties = false;
-            } else {
-                // we'll explicitly set the default, which for swagger
-                definitions[referenceType.refName].additionalProperties = true;
-            }
+            definitions[referenceType.refName].additionalProperties = this.determineImplicitAdditionalPropertiesValue();
```
dgreene1: "Always be refactoring" :)

src/routeGeneration/templateHelpers.ts
```diff
@@ -359,7 +361,16 @@ export class ValidationService {
     return new Buffer(value);
   }
 
-  public validateModel(name: string, value: any, refName: string, fieldErrors: FieldErrors, parent = ''): any {
+  public validateModel( input: {
```
dgreene1: I was getting all kinds of problems when trying to add a required property after an optional property, so I switched over to an object input so that the parameters don't have an order. This also helped with readability in the tests. I think this approach is sometimes called "named properties."

dgreene1:
@WoH since I'm hoping to get this PR out before too many people start using v2.44 (read above), I'd love to get some review on this. Would you be interested in that?

nenadalm:
@dgreene1 I am not using this library - I used it only on one project that was cancelled. I use different approach now for js/ts inspired by reitit (it is part of one project that is not open source): https://github.com/metosin/reitit. In Clojure(Script) I use reitit directly.

My ideas were not in alignment with how tsoa works. See https://github.com/lukeautry/tsoa/pull/132.

(the above is just explanation why I won't review this pr as requested)

src/config.ts
```diff
@@ -38,9 +38,9 @@ export interface SwaggerConfig {
   entryFile: string;
 
   /**
-   * Set this to true if you want to prevent requests from coming in with more properties than are known/allowed
+   * Modes that allow you to prevent input data from entering into your API. This will document your decision in the swagger.yaml and it will turn on excess-property validation (at runtime) in your routes.
    */
-  noImplicitAdditionalProperties?: boolean;
+  noImplicitAdditionalProperties?: 'throw-on-extras' | 'silently-remove-extras';
```
WoH: Can we add another flag like `trimAdditionalProperties`?
This would make sure there are no breaking changes.
If someone sets both to true, we can detect, warn and use the stricter `noImplicitAdditionalProperties`.
This reminds me of the various feature flags TS itself has. While it surely is not ideal, it punts the breaking change to 3.0, where we can consolidate both flags into on. What do you think @dgreene1?

dgreene1: That was my original thought, but I really don’t like the idea that “true” would default to one of the two options and that it wouldn’t be clear to the user which one we’ve selected for them.

I don’t have the answer here. I’m kinda lost.

dgreene1: @WoH I just spoke to one of the architects at my day job and they said that in their option the default (ie the value true relates to) should be the silently remove option. This leads me to think that if I were to default true to something, that’s what I would choose. However, that would also be a breaking change since when I released yesterday that first introduced this only had the throw option.

WoH: I guess I still prefer warning users of the new version about invalid config (both flags true) over warning users impacted that their config is invalid (with the 2 flag case someone is actively setting something and presumably paying attention), although the recency of the change definitely makes it a lot more reasonable to go the other way.

endor:
Ah, sorry, I'm a bit lost. So what's the difference between just having `throw-on-extras=false` vs `silently-remove-extras=true`? Is the idea that silently removing extra properties in the framework would allow users to directly insert the data into a database for example?

Personally I like opinionated frameworks. So for my taste, I'd like us to stick to `throw-on-extras` and not have the `silently-remove-extras`. For one, why is it less robust? If expected fields change, then the API version or endpoint should change as well. Not only that, but it actually actively tells the user: look, something has changed! If we just remove the fields, the user might not even notice that their data is not being evaluated.

src/routeGeneration/templateHelpers.ts
```diff
@@ -383,16 +394,27 @@ export class ValidationService {
       } else if (additionalProperties === false) {
         Object.keys(value).forEach((key: string) => {
           if (!alreadyValidatedProperties.has(key)) {
-            fieldErrors[parent + '.' + key] = {
-              message: `${key} is an excess property and therefore is not allowed`,
-              value: key,
-            };
+            if (swaggerConfig.noImplicitAdditionalProperties === 'throw-on-extras') {
+              fieldErrors[parent + '.' + key] = {
```
WoH:
```suggestion
              fieldErrors[parent + key] = {
```

dgreene1: @WoH I can’t change that since it would be a breaking contract for consumers. I actually have my own custom TSOA error parser and it depends on the specific format.

dgreene1:
Yea @endor you make some good points. I think it’s just that many API developers prefer to accept the values they asked for and to ignore the ones that they didn’t ask for. In fact, that’s the default behavior for C# serializers and potentially Swift, Kotlin, and Java from my minimal research.

So that’s one of the reasons why I’d prefer that consumers of tsoa are explicit about which type of reaction they want the API to have. In some applications it might be more appropriate to throw, and in others might be more appropriate to be more forgiving like the principle states.

That’s why I abhors the idea of “true” being set to one of the actions. But I only have to do that if we’re not allowed to make a breaking change.

The way I’m leaning towards now is to avoid the breaking change but to print a ton of warnings if you use `true` which will be deprecated.

src/routeGeneration/templateHelpers.ts
WoH: Just means the next release should be a major version.
This should definitely be fixed IMHO.

src/routeGeneration/templateHelpers.ts
```diff
@@ -383,16 +394,27 @@ export class ValidationService {
       } else if (additionalProperties === false) {
         Object.keys(value).forEach((key: string) => {
           if (!alreadyValidatedProperties.has(key)) {
```
WoH: I think there should be and extracted function getExcessPropertyForModel(model, properties) that returns an array of excess properties. That would be a lot easier to read.

src/routeGeneration/templateHelpers.ts
WoH: minimalSwaggerConfig => handleImplicitProperties
This should only be { noImplicitAdditionalProperties: "..." }, but only 'throw-on-extras' | 'silently-remove-extras' | 'ignore'

dgreene1:
Alright @WoH & @endor, I submitted a new commit that has the backwards compatibility in place. I've left `true` to be equated with `throw-on-extras` since (a) that's the backwards compatible choice and (b) since @endor prefers that consumers get notified if they submit additional properties.

I did my best to write this in a way that adds as little "cruft" to the code. That being said, I'm purposefully choosing to blast the user's console with warnings if they choose to submit `true` or `false`. I don't think it's fair to make a choice for the consumer without giving them a ton of notice.

@WoH I also made your change where you requested greater readability.

src/routeGeneration/templateHelpers.ts
dgreene1: I created a new function called `isAnExcessProperty` that still utilizes the `Set` (for performance reasons) but should make the code more readable.

WoH:
I am not really convinced by the performance argument for not extracting a `getExcessPropertyForModel`. It only just checks the given property names and subtracts the ones specified on the model (no further validation).
If there is a legitimate performance concern here, nothing convinces me faster than a benchmark.
I will admit, however, that I am kinda biased here since I know 2 cases where I'd like to reuse that.

Also, is there a reason to pass `{ noImplicitAdditionalProperties: "..." }` instead of a string literal?

src/routeGeneration/templateHelpers.ts
dgreene1: I don't understand your comment here @WoH. Are you saying that the name here is wrong? Or that the type is wrong?

Either way, I still don't understand since it's a type that I use in a lot of different places and it represents a minimal interface for exactly the data that I'm expecting.

I'm also confused about the `'ignore'` in your comment about `'throw-on-extras' | 'silently-remove-extras' | 'ignore'` since I don't provide `'ignore'` as a valid option.

dgreene1:
> I am not really convinced by the performance argument for not extracting a getExcessPropertyForModel. 
> It only just checks the given property names and subtracts the ones specified on the model (no further validation).
> If there is a legitimate performance concern here, nothing convinces me faster than a benchmark.

We've never had a requirement to have performance checks to be able to contribute to tsoa. That being said, `Set` is a Big-O(1) operation. So if you have some other recommended algorithm, then please feel free to add the implementation here in the comments or make a branching PR that we can discuss. It's difficult to make a performance benchmark of a concrete implementation as compared to a vaporware implementation.

> I will admit, however, that I am kinda biased here since I know 2 cases where I'd like to reuse that.

If and when you have a need for a public function we can pull it out. But until then it's simply a [YAGNI](http://www.agilenutshell.com/yagni).

> Also, is there a reason to pass { noImplicitAdditionalProperties: "..." } instead of a string literal?

Yes, there is a very important reason: That config option can also be `true`, or `false`, or `undefined`.

WoH:
> If and when you have a need for a public function we can pull it out. But until then it's simply a YAGNI. When you do need it, then you can add it as part of that PR and you can refactor this code to use your public function (if it is in-fact faster).

Fair. Should still be an internal thing, but I'll need it for the Intersection and the nested object prop checking :smiley: 

> Yes, there is a very important reason: That config option can also be true, or false, or undefined.

But you parse that [here](https://github.com/lukeautry/tsoa/blob/75082b9d0cafcdb880f5a7765985fb9b4f0c4025/src/module/generate-routes.ts#L58). What I wanted to suggest is returning 'silently-remove-extras'' | 'throw-on-extras' | 'ignore' from that function.
So 
```ts
export interface SwaggerConfigRelatedToRoutes {
  noImplicitAdditionalProperties?: SwaggerConfig['noImplicitAdditionalProperties'];
}
```
becomes
```ts
type ImplicitAdditionalProperties =
    | "silently-remove-extras'"
    | "throw-on-extras"
    | "ignore";
```

src/routeGeneration/templateHelpers.ts
dgreene1: I clarify below why I’d like to keep the data immutable.

src/routeGeneration/templateHelpers.ts
dgreene1: @WoH if you think that this `parent + ‘.’ key` syntax is an issue, please open up a new PR explaining why it’s a bug. It has not been made clear why this is a bug. Additionally it’s been a part of tsoa for a long time.

dgreene1:
@WoH that suggestion is just a stylistic preference of yours.

> But you parse that here. What I wanted to suggest is returning 'silently-remove-extras'' | 'throw-on-extras' | 'ignore' from that function.

I prefer to treat data immutably. But I come from a declarative programming background. I’ve even considered marking each config type as DeepReadonly (from the utility-types npm Module) so that the config is purposefully immutable.

But for an open source project I think it’s acceptable and maybe even encouraged that each contributor uses the programming paradigm that they are most comfortable with.

That being said, I believe that immutable data is considered to be a positive practice even within OO communities.

WoH:
> I prefer to treat data immutably. But I come from a declarative programming background. I’ve even considered marking each config type as DeepReadonly (from the utility-types npm Module) so that the config is purposefully immutable.
> That being said, I believe that immutable data is considered to be a positive practice even within OO communities.

I believe that does not affect the suggestion, you can and should still treat the config immutably.

> @WoH that suggestion is just a stylistic preference of yours.

Yes, it is, see below.

> But for an open-source project I think it’s acceptable and maybe even encouraged that each contributor uses the programming paradigm that they are most comfortable with.

What I meant to do is suggest my stylistic preference for consideration and try to make you understand my point of view that lead me to make the suggestion. I did not mean to push you in a direction you do not feel comfortable with. When you felt pressured or uncomfortable, that was poor communication/manners on my part and I apologize. Since both ways do not affect the correctness, you should feel happy with the great work on this PR however you decide to pass that piece of config around.

dgreene1:
Okay thank you @WoH. Yea, I wasn’t sure if it was the case of “I’m not going to approve this until the change is made” or more of a “here’s a cool idea” kind of thing. It was a cool idea by the way.

Okay, I guess I’ll leave this open for the weekend just to leave @endor or anyone else a chance to comment on it and then I’ll merge it. Thank you everyone for your great comments so far. :)

dgreene1:
I know I said that I'd wait the weekend to allow for more review, but this code is needed for two other PRs that @WoH is working on. So I'm going to merge this now.

Issue #353, opened 2019-05-14
Dynamic Query/Header Param


oreporan:
I have a request which should look like this

`users/query?someKey=someValue&someKey2=someValue2`

These are dynamic query params that I handle in the server. Is there any way to create this using tsoa?

I tried something like this - 

```ts
export interface QueryFilters {
    [x: string]: string;
}
```

```ts
...
 public async get(
        @Query() filters?: QueryFilters
    ): Promise<User[]> {
        return getUsers(filters);
    }
```

but I'm getting error - 
```
 Error: @Query('filters') Can't support 'refObject' type.
```

HoldYourWaffle:
I'm currently working on an implementation of `refObject` for `@Query` (and possibly `@Header` too) by "flattening" the properties as individual parameters and reassembling them at the server.

I have a **very basic** but (mostly) working implementation ready, which is good enough for my own project for now. When I have the time I'll add the missing parts and create a PR.

dgreene1:
@HoldYourWaffle we’d love if you could submit your PR for this.
@dmitrirussu had a similar and interesting request here: https://github.com/lukeautry/tsoa/issues/324#issue-391707399

I also work with another team that would find this really valuable as they use OData filters. So suffice to say, you’d be quite the hero if you were to submit this! :)

HoldYourWaffle:
Sorry for the late response, haven't touched this in a while due to summer break.

I don't remember much and I'm not able to really look into the code right now, but from what I remember my implementation was very basic and specific to my use-case. I'll see what I can do to fix it up and maybe make it more usable for other use-cases.

dgreene1:
I believe that https://github.com/lukeautry/tsoa/issues/219#issuecomment-526403819 is a duplicate of this. I’ll close that one to keep this open.

dgreene1:
@Templum @anro87 when we introduce array like query params then this is the issue number to track. I’m cc’ing you so you’re not left out. We apologize that your original feature request was not added in a timely manner.

givethemheller:
@WoH can you point me to where the changes in the code would need to be to get this working? Its currently blocking for me. Would be happy to work on it (maybe tonight/this weekend)

WoH:
This should not be a blocker, albeit you'd have to write the OpenAPI for yourself. Assuming Express, but you can inject the `Request` (and look at req.headers) and add the Spec to your tsoa.json.
I'm not sure how the API should look like for this, the one suggested is a non-starter.
[...]

PR #415, opened 2019-08-14
ObjectTypeLiteral support

WoH:
Currently open issue: https://github.com/lukeautry/tsoa/issues/50

R4R

### DIFF
--- a/src/metadataGeneration/tsoa.ts
+++ b/src/metadataGeneration/tsoa.ts
@@ -87,12 +87,13 @@ export namespace Tsoa {
     | 'any'
     | 'refEnum'
     | 'refObject'
+    | 'objectLiteral'
     | 'union'
     | 'intersection';
 
   export type RefTypeLiteral = 'refObject' | 'refEnum';
 
-  export type PrimitiveTypeLiteral = Exclude<TypeStringLiteral, RefTypeLiteral | 'enum' | 'array' | 'void' | 'union' | 'intersection'>;
+  export type PrimitiveTypeLiteral = Exclude<TypeStringLiteral, RefTypeLiteral | 'enum' | 'array' | 'void' | 'objectLiteral' | 'union' | 'intersection'>;
 
   export interface Type {
     dataType: TypeStringLiteral;
@@ -105,6 +106,11 @@ export namespace Tsoa {
   export interface EnumerateType extends Type {
     dataType: 'enum';
     enums: string[];
+  }
+
+  export interface ObjectLiteralType extends Type {
+    dataType: 'objectLiteral';
+    properties: Property[];
   }
 
   export interface ArrayType extends Type {
--- a/src/metadataGeneration/tsoa.ts
+++ b/src/metadataGeneration/tsoa.ts
@@ -87,13 +87,13 @@ export namespace Tsoa {
     | 'any'
     | 'refEnum'
     | 'refObject'
-    | 'objectLiteral'
+    | 'nestedObjectLiteral'
     | 'union'
     | 'intersection';
 
   export type RefTypeLiteral = 'refObject' | 'refEnum';
 
-  export type PrimitiveTypeLiteral = Exclude<TypeStringLiteral, RefTypeLiteral | 'enum' | 'array' | 'void' | 'objectLiteral' | 'union' | 'intersection'>;
+  export type PrimitiveTypeLiteral = Exclude<TypeStringLiteral, RefTypeLiteral | 'enum' | 'array' | 'void' | 'nestedObjectLiteral' | 'union' | 'intersection'>;
 
   export interface Type {
     dataType: TypeStringLiteral;
@@ -109,8 +109,9 @@ export namespace Tsoa {
   }
 
   export interface ObjectLiteralType extends Type {
-    dataType: 'objectLiteral';
+    dataType: 'nestedObjectLiteral';
     properties: Property[];
+    additionalProperties?: Type;
   }
 
   export interface ArrayType extends Type {
--- a/src/metadataGeneration/typeResolver.ts
+++ b/src/metadataGeneration/typeResolver.ts
@@ -17,7 +17,7 @@ syntaxKindMap[ts.SyntaxKind.VoidKeyword] = 'void';
 const localReferenceTypeCache: { [typeName: string]: Tsoa.ReferenceType } = {};
 const inProgressTypes: { [typeName: string]: boolean } = {};
 
-type UsableDeclaration = ts.InterfaceDeclaration | ts.ClassDeclaration | ts.TypeAliasDeclaration;
+type UsableDeclaration = ts.InterfaceDeclaration | ts.ClassDeclaration | ts.TypeAliasDeclaration | ts.PropertySignature;
 
 export class TypeResolver {
   constructor(private readonly typeNode: ts.TypeNode, private readonly current: MetadataGenerator, private readonly parentNode?: ts.Node, private readonly extractEnum = true) {}
@@ -90,22 +90,44 @@ export class TypeResolver {
     }
 
     if (this.typeNode.kind === ts.SyntaxKind.TypeLiteral) {
-      const properties = (this.typeNode as ts.TypeLiteralNode).members
+      const typeLiteralNode = this.typeNode as ts.TypeLiteralNode;
+      const properties = typeLiteralNode.members
         .filter(member => member.kind === ts.SyntaxKind.PropertySignature)
         .reduce((res, propertySignature: ts.PropertySignature) => {
           const type = new TypeResolver(propertySignature.type as ts.TypeNode, this.current, this.typeNode).resolve();
 
           return [
             {
+              default: getJSDocComment(propertySignature, 'default'),
+              description: this.getNodeDescription(propertySignature),
+              format: this.getNodeFormat(propertySignature),
               name: (propertySignature.name as ts.Identifier).text,
               required: !propertySignature.questionToken,
               type,
+              validators: getPropertyValidators(propertySignature),
             } as Tsoa.Property,
             ...res,
           ];
         }, []);
 
-      return { dataType: 'objectLiteral', properties } as Tsoa.ObjectLiteralType;
+      const indexMember = typeLiteralNode.members.find(member => member.kind === ts.SyntaxKind.IndexSignature);
+      let additionalType: Tsoa.Type | undefined;
+
+      if (indexMember) {
+        const indexSignatureDeclaration = indexMember as ts.IndexSignatureDeclaration;
+        const indexType = new TypeResolver(indexSignatureDeclaration.parameters[0].type as ts.TypeNode, this.current).resolve();
+        if (indexType.dataType !== 'string') {
+          throw new GenerateMetadataError(`Only string indexers are supported.`);
+        }
+
+        additionalType = new TypeResolver(indexSignatureDeclaration.type as ts.TypeNode, this.current).resolve();
+      }
+
+      return {
+        additionalProperties: indexMember && additionalType,
+        dataType: 'nestedObjectLiteral',
+        properties,
+      } as Tsoa.ObjectLiteralType;
     }
 
     if (this.typeNode.kind === ts.SyntaxKind.ObjectKeyword) {
@@ -507,7 +529,7 @@ export class TypeResolver {
 
       const modelTypeDeclaration = node as UsableDeclaration;
       return (modelTypeDeclaration.name as ts.Identifier).text === typeName;
-    }) as UsableDeclaration[];
+    }) as Array<Exclude<UsableDeclaration, ts.PropertySignature>>;
 
     if (!modelTypes.length) {
       throw new GenerateMetadataError(
@@ -709,7 +731,7 @@ export class TypeResolver {
     return undefined;
   }
 
-  private getModelInheritedProperties(modelTypeDeclaration: UsableDeclaration): Tsoa.Property[] {
+  private getModelInheritedProperties(modelTypeDeclaration: Exclude<UsableDeclaration, ts.PropertySignature>): Tsoa.Property[] {
     const properties = [] as Tsoa.Property[];
     if (modelTypeDeclaration.kind === ts.SyntaxKind.TypeAliasDeclaration) {
       return [];
--- a/src/metadataGeneration/typeResolver.ts
+++ b/src/metadataGeneration/typeResolver.ts
@@ -90,7 +90,22 @@ export class TypeResolver {
     }
 
     if (this.typeNode.kind === ts.SyntaxKind.TypeLiteral) {
-      return { dataType: 'any' } as Tsoa.Type;
+      const properties = (this.typeNode as ts.TypeLiteralNode).members
+        .filter(member => member.kind === ts.SyntaxKind.PropertySignature)
+        .reduce((res, propertySignature: ts.PropertySignature) => {
+          const type = new TypeResolver(propertySignature.type as ts.TypeNode, this.current, this.typeNode).resolve();
+
+          return [
+            {
+              name: (propertySignature.name as ts.Identifier).text,
+              required: !propertySignature.questionToken,
+              type,
+            } as Tsoa.Property,
+            ...res,
+          ];
+        }, []);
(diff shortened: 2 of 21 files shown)


tests/unit/swagger/definitionsGeneration/definitions.spec.ts
```diff
@@ -318,6 +318,29 @@ describe('Definition generation', () => {
                     expect(propertySchema).to.not.haveOwnProperty('additionalProperties', `for property ${propertyName}`);
                     expect(propertySchema['x-nullable']).to.eq(true, `for property ${propertyName}[x-nullable]`);
                 },
+                objLiteral: (propertyName, propertySchema) => {
+                    expect(propertySchema).to.deep.include({
```
dgreene1: Can you use deep.equal instead?

tests/unit/swagger/definitionsGeneration/definitions.spec.ts
```diff
@@ -318,6 +318,29 @@ describe('Definition generation', () => {
...
+                      properties: {
+                        name: {
+                          type: 'string',
+                        },
+                        nested: {
+                          properties: {
+                            allOptional: {
```
dgreene1: I was unaware that swagger let you specify a type within a type. I thought it only allowed $refs. I’m assuming you’ve checked that the output is valid Schema within the “pet shop” Swagger editor?

If it really is that that simple then this PR is going to make a lot of people happy! :)

src/metadataGeneration/tsoa.ts
```diff
@@ -68,7 +68,7 @@ export namespace Tsoa {
     validators: Validators;
   }
 
-  export type TypeStringLiteral = 'string' | 'boolean' | 'double' | 'float' | 'integer' | 'long' | 'enum' | 'array' | 'datetime' | 'date' | 'binary' | 'buffer' | 'byte' | 'void' | 'object' | 'any' | 'refEnum' | 'refObject';
+  export type TypeStringLiteral = 'string' | 'boolean' | 'double' | 'float' | 'integer' | 'long' | 'enum' | 'array' | 'datetime' | 'date' | 'binary' | 'buffer' | 'byte' | 'void' | 'object' | 'any' | 'refEnum' | 'refObject' | 'objectLiteral';
```
dgreene1: What would you think about renaming it to be “nestedObject”? I ask this because we already have the “Object” type and the “refObject” so it makes for some confusing code. I realize that the type your describing is most accurately described as an objectLiteral, but I think a different name might lead to more clear code for future contributors.

tests/unit/swagger/definitionsGeneration/definitions.spec.ts
WoH: I checked [swagger editor](https://editor.swagger.io) and [swagger parser](https://apidevtools.org/swagger-parser/online/).
Just like you, I expected to move this to the OAPI3 generator but tried with the OAPI2 generator mostly to find hints what the errors would be. Turns out no errors :smile: 

For convenience, here is a small-ish YAML:
```yaml
basePath: /v1
consumes:
  - application/json
definitions:
  EnumIndexValue:
    description: EnumIndexValue.
    enum:
      - '0'
      - '1'
    type: string
  TestModel:
    description: This is a description of a model
    properties:
      objLiteral:
        properties:
          name:
            type: string
          nested:
            properties:
              bool:
                type: boolean
              optional:
                type: number
                format: double
              allOptional:
                properties:
                  one:
                    $ref: '#/definitions/EnumIndexValue'
                  two:
                    type: string
                type: object
            required:
              - allOptional
              - bool
            type: object
        required:
          - name
        type: object
    required:
      - objLiteral
info:
  title: Test API
  version: 1.0.0
  description: Description of a test API
  license:
    name: MIT
paths:
  /GetTest:
    get:
      operationId: GetModel
      produces:
        - application/json
      responses:
        '200':
          description: Returns TestModel
          schema:
            $ref: '#/definitions/TestModel'
          examples:
            application/json:
              objLiteral:
                name: a string
                nested:
                  bool: true
      description: |-
        This is a description of the getModel method
        this is some more text on another line
      security: []
      parameters: []
produces:
  - application/json
swagger: '2.0'
securityDefinitions:
  basic:
    type: basic
host: 'localhost:3000'
```

tests/unit/swagger/definitionsGeneration/definitions.spec.ts
WoH: Prefer not to. I'm not setting any of default, description or format (which I understand to be not allowed in this context). Therefore, adding these 3 undefined props add a lot of unnecessary noise imho.

tests/integration/dynamic-controllers-express-server.spec.ts
```diff
@@ -783,6 +783,9 @@ describe('Express Server', () => {
       },
       numberArray: [1, 2],
       numberValue: 5,
+      objLiteral: {
```
WoH: @dgreene1 Every time I add a test to the TestModel, I end up with a lot of these additions in various places. Would you be open to someone refactoring this (outside of this PR)?

dgreene1: @WoH I didn’t write the first TestModel, so I can’t speak to if it’s okay to refactor it to be used in less places. However, I don’t expect that we will be introducing significantly more types so I’m not sure that it’s worth refactoring without bowing what we’d lose in the process. Like it’s low value and potentially high risk?

tests/unit/swagger/definitionsGeneration/definitions.spec.ts
dgreene1: I understand that it adds noise, but it could potentially prevent a future bug if it caches something. I always prefer quality control over readability.

src/metadataGeneration/tsoa.ts
dgreene1: Another option is “InLineObject”

tests/integration/dynamic-controllers-express-server.spec.ts
WoH: The value is not going through multiple places to copy/paste a single prop...

Some of the most obvious candidates are:
- https://github.com/lukeautry/tsoa/blob/master/tests/integration/dynamic-controllers-express-server.spec.ts#L770
- https://github.com/lukeautry/tsoa/blob/master/tests/integration/express-server.spec.ts#L770
- https://github.com/lukeautry/tsoa/blob/master/tests/integration/hapi-server.spec.ts#L712
- https://github.com/lukeautry/tsoa/blob/master/tests/integration/koa-server.spec.ts#L699

They are already out of sync and I don't think it's on purpose ;)

src/metadataGeneration/tsoa.ts
WoH: I understand your concern for new contributors, I was siding more with the expectation of someone comfortable with the internal TS terminology with my initial proposal.
Maybe as a 
Is `nestedObjectTypeLiteral` something you would agree on as a compromise between the two?

dgreene1: Yes, `nestedObjectTypeLiteral` is a wonderful compromise. :)

tests/unit/swagger/definitionsGeneration/definitions.spec.ts
WoH: https://github.com/lukeautry/tsoa/pull/415/commits/da71a8960e79c534cf951deec603fe3e155a2aa7

dgreene1:
I see that this is still marked as "Draft" Do you want to remove that @WoH or are you still waiting for more feedback from other members or something?

HoldYourWaffle:
I'm a little confused (probably due to severe lack of sleep) so sorry if this is a stupid question, but what kind of 'objects' can we use know? Is it an implementation of #353?

WoH:
> I'm a little confused (probably due to severe lack of sleep) so sorry if this is a stupid question, but what kind of 'objects' can we use know? Is it an implementation of #353?

No. The primary concern of this PR are type definitions wrapped inside `{}`.
See [here](https://github.com/lukeautry/tsoa/pull/415/files#diff-c5f67349853619c9f3b6b1b7b9bb6df7) for reference:

```ts
interface Model {
  prop: {
    nested: string
  }
}
```

or 
```ts
function(): { value: number } {
  ...
}
```

> I see that this is still marked as "Draft" Do you want to remove that @WoH or are you still waiting for more feedback from other members or something?

Mostly I'm dreading the tests I'll want to write around this. This PR should already cover nested JSDoc annotations and pretty much everything we support on interface declarations. However, that means we want to make sure it does and will not regress at any point in the future.

HoldYourWaffle:
> The primary concern of this PR are type definitions wrapped inside {}

Wait how is that different from my implementation of #353 (supporting `refObject`s)? Does this only support 'objects' with a single property?
Again sorry for my stupidity, I just can't get it to click in my head for some reason...

WoH:
> Wait how is that different from my implementation of #353 (supporting refObjects)? Does this only support 'objects' with a single property?
Again sorry for my stupidity, I just can't get it to click in my head for some reason...

Ok, so a `refObject` is the internal representation for a named interface, sometimes a type alias in tsoa. Simply called a `referenceType`, so that's the term for this I will use in the context of this explanation. This thing is not supported as QueryParam.

What I'm doing is adding a way to resolve `ObjectLiteralTypes`. That's what would allow you to not create a new interface every time you want to add an arbitrary amount of sub-properties to your property (of a `referenceType`) but to do it the TypeScript way of adding another set of angle brackets and specifying the sub-properties between. These types would then be inlined into the property schema of the `referenceType`

HoldYourWaffle:
Oh wait I think I get it, you're adding support for an "anonymous" object, just not in `@Query`.
That makes sense, thank you for making my dumb and tired brain understand.

As a side note, I might post a very work-in-progress PR with my implementation of `refObject` (and I suppose these literals too in the future) in `@Query`. I made this implementation 2 months ago in a really big rush (needed it for a project with a tight deadline) so it's nowhere near a "mergeable" state (I'm not even sure what some parts are doing anymore), but it might be good to get some feedback on the general direction I'm going.

WoH:
I'll open this one up, got some (admittedly lazy) JSDoc validations in.
Excess prop handling should be covered in the koa no additional integration spec, but there's probably room for more.

HoldYourWaffle:
Just like with #444 we should consider which PR we want to merge first, my cleanup (#445) or this.
(Again assuming my cleanup gets merged)

WoH:
Anticipating some of @dgreene1's review in https://github.com/lukeautry/tsoa/pull/415/commits/f3d9fb0f8d6991e864ba52a29cf3a3a3a8d9e532 and https://github.com/lukeautry/tsoa/pull/415/commits/55f9ca79bb1d8f60c52d80f693faf903ca623bf3.

I tried to improve UX for validations, throw if metadata is missing, added an assertion that a nested object value is an object in the first place.

Got around to spec excess prop handling (ignore, remove, throw), see https://github.com/lukeautry/tsoa/pull/415/commits/55f9ca79bb1d8f60c52d80f693faf903ca623bf3

Happy to get reviews at this point.

tests/fixtures/testModel.ts
```diff
@@ -333,12 +422,6 @@ export class TestClassModel extends TestClassBaseModel {
   stringProperty: string;
   protected protectedStringProperty: string;
 
-  public static typeLiterals = {
```
WoH: I removed these since they were messing up type resolution (`Boolean, Number and String`) will throw.

WoH:
Rebased for the green checkmark.

WoH:
@dgreene1 I just rebased and realized checks are not running anymore. Is this intended?

dgreene1:
> @dgreene1 I just rebased and realized checks are not running anymore. Is this intended?

I don’t know what’s up with github actions. I just made a change that should hopefully force the tests to be run on each commit.

src/metadataGeneration/typeResolver.ts
```diff
@@ -89,8 +89,44 @@ export class TypeResolver {
...
+          return [
+            {
+              default: getJSDocComment(propertySignature, 'default'),
+              description: this.getNodeDescription(propertySignature),
+              format: this.getNodeFormat(propertySignature),
+              name: (propertySignature.name as ts.Identifier).text,
+              required: !propertySignature.questionToken,
+              type,
+              validators: getPropertyValidators(propertySignature),
+            } as Tsoa.Property,
```
dgreene1: @WoH can you create a temp variable of type Tsoa.Property instead of using `as`?

So like this instead:
```ts
const aProperty: Tsoa.Property = {
```

WoH: Done

WoH:
> I don’t know what’s up with github actions. I just made a change that should hopefully force the tests to be run on each commit.

[Works now](https://github.com/lukeautry/tsoa/pull/415/checks?check_run_id=209190518)

tests/integration/dynamic-controllers-express-server.spec.ts
```diff
@@ -600,6 +667,47 @@ describe('Express Server', () => {
...
+
+          expect(body.fields['body.nestedObject.arrayMax5Item'].message).to.equal('maxItems 5');
+          expect(body.fields['body.nestedObject.arrayMax5Item'].value).to.deep.equal(bodyModel.arrayMax5Item);
+          expect(body.fields['body.nestedObject.arrayMin2Item'].message).to.equal('minItems 2');
+          expect(body.fields['body.nestedObject.arrayMin2Item'].value).to.deep.equal(bodyModel.arrayMin2Item);
+          expect(body.fields['body.nestedObject.arrayUniqueItem'].message).to.equal('required unique array');
+          expect(body.fields['body.nestedObject.arrayUniqueItem'].value).to.deep.equal(bodyModel.arrayUniqueItem);
+          expect(body.fields['body.nestedObject.model'].message).to.equal('invalid object');
+          expect(body.fields['body.nestedObject.model'].value).to.deep.equal(bodyModel.model);
+          expect(body.fields['body.nestedObject.mixedUnion'].message).to.equal(
```
dgreene1: We should have a negative assertion for `intersection` here too.

tests/integration/dynamic-controllers-express-server.spec.ts
```diff
@@ -527,6 +550,28 @@ describe('Express Server', () => {
...
+          expect(body.nestedObject.numberMin5).to.equal(bodyModel.nestedObject.numberMin5);
+          expect(body.nestedObject.stringMax10Lenght).to.equal(bodyModel.nestedObject.stringMax10Lenght);
+          expect(body.nestedObject.stringMin5Lenght).to.equal(bodyModel.nestedObject.stringMin5Lenght);
+          expect(body.nestedObject.stringPatternAZaz).to.equal(bodyModel.nestedObject.stringPatternAZaz);
+
+          expect(body.nestedObject.arrayMax5Item).to.deep.equal(bodyModel.nestedObject.arrayMax5Item);
+          expect(body.nestedObject.arrayMin2Item).to.deep.equal(bodyModel.nestedObject.arrayMin2Item);
+          expect(body.nestedObject.arrayUniqueItem).to.deep.equal(bodyModel.nestedObject.arrayUniqueItem);
+          expect(body.nestedObject.model).to.deep.equal(bodyModel.nestedObject.model);
+          expect(body.nestedObject.mixedUnion).to.deep.equal(bodyModel.nestedObject.mixedUnion);
```
dgreene1: We might as well have a positive test for intersection here. I think that’s the only coverage that I couldn’t find. Great work on this PR! :)

WoH: https://github.com/lukeautry/tsoa/pull/415/commits/d0fe540cc6734d10a10b189fe3c24c4722979505

tests/integration/dynamic-controllers-express-server.spec.ts
WoH: https://github.com/lukeautry/tsoa/pull/415/commits/d0fe540cc6734d10a10b189fe3c24c4722979505

WoH: Thanks for your review. I added intersection and nested intersection specs, both valid and invalid for all the controllers that have these tests.

Issue #417, opened 2019-08-15
Internal typeResolver cache causes issues when running generateRoutes before generateSwaggerSpec


ryankeener:
This cache is re-used when running `generateSwaggerSpec` after `generateRoutes` within the same node process.
https://github.com/lukeautry/tsoa/blob/200b887c5d472ec53810c18f4d8b90a774b5785d/src/metadataGeneration/typeResolver.ts#L17

This bit of code
https://github.com/lukeautry/tsoa/blob/200b887c5d472ec53810c18f4d8b90a774b5785d/src/metadataGeneration/typeResolver.ts#L256-L259

short circuits the code here
https://github.com/lukeautry/tsoa/blob/200b887c5d472ec53810c18f4d8b90a774b5785d/src/metadataGeneration/typeResolver.ts#L274
and here
https://github.com/lukeautry/tsoa/blob/200b887c5d472ec53810c18f4d8b90a774b5785d/src/metadataGeneration/typeResolver.ts#L529

which results in some models not being added to the `referenceTypeMap` here
https://github.com/lukeautry/tsoa/blob/200b887c5d472ec53810c18f4d8b90a774b5785d/src/metadataGeneration/typeResolver.ts#L123

dgreene1:
@ryankeener, would you be interested in submitting a PR to fix this?

ryankeener:
I dug into it far enough to provide the information above but I don't think I have a deep enough understanding of what's going on (nor the time to do so right now) to create a PR. For now I'm just brute-force clearing `tsoa` and its components from the `require.cache` before executing the second stage.

I _think_ duplicating these two lines into the `existingType` check also worked but I can't be sure about it. 
```ts
      if (existingType) {
        const modelType = this.getModelTypeDeclaration(type);
        const properties = this.getModelProperties(modelType, genericTypes);
        return existingType;
      }
```

dgreene1:
@ryankeener this fix will go out in today's release. Please let me know if it works for you. Should be version 2.4.9.

dgreene1:
k @ryankeener it was released. Please let us know if the issue is still occurring. See our release notes for other changes: https://github.com/lukeautry/tsoa/releases

ryankeener:
@dgreene1 It works, thanks!

PR #419, opened 2019-08-16
Feature/support auto import controllers

### LINKED ISSUE #420: Missing support for dynamic controllers in the spec generator

joni7777:
Add support for the swagger spec generator as well, not only the routes generator
FIX #420


joni7777:
@dgreene1  There is some test on the spec generator that keep failing can you care to take a look to see why? unless we push this the glob path isn't really working

src/config.ts
```diff
@@ -37,6 +37,11 @@ export interface SwaggerConfig {
    */
   entryFile: string;
 
+  /**
+   * Controllers glob path, contains directories with wild card to file names.
+   */
+  controllerPathGlobs?: string[];
```
dgreene1: This is the addition that I'm confused by @joni7777. It doesn't make sense to have to have the user specific the same property in both the swagger config and the route config. It would be more ideal if both tsoa CLI commands looked in the same place. That's actually what I feel like must be happening since I am using the feature you wrote and I'm getting my routes and my swagger documents generated properly.

joni7777: But the swagger spec is running on the entry file only which don't include the controllers and the modules
Or does it have the reference using the import to the routes file?

I think we should also implement a way to consume tsoa on the app.ts file without pre generating routes file, by returning the routes class which being able to be used with
`app.use(TsoaRoutes)`

joni7777: Do you have slack/messenger or some IM app?

dgreene1: I don’t have slack or any IM app. If tsoa isn’t generating the swagger and the routes then please describe the reproduction steps in the issue. I’m able to produce both using the new glob config.

joni7777: I'm trying to understand why there isn't any issue
If the swagger spec gets only the app entry file which is not importing the controllers
How does the spec metadata generator find all the controllers and the models if the routes generator find them by glob

dgreene1: Maybe this is why it’s still working? Although I must say that I wasn’t aware that there was a “cache” until this other issue was  made: https://github.com/lukeautry/tsoa/issues/417

joni7777: I wrote an explanation in the issue #420 on why its working

dgreene1:
@joni7777 I looked at unit tests like you asked. I fixed the mistake and I also added information to the readme (plus some other enhancements) as part of https://github.com/lukeautry/tsoa/pull/430. So we'll close this PR in favor of https://github.com/lukeautry/tsoa/pull/430

PR #371, opened 2019-07-15
typo  parem => param

yanick:

PR #424, opened 2019-08-17
ParameterGenerator: Support DELETE body, path decorator

### LINKED ISSUE #362: DELETE body is not supported
Hi,

Http request Body used with Delete method throws the following error:
> Error: @Body('request') Can't support in DELETE method.

[https://stackoverflow.com/questions/299628/is-an-entity-body-allowed-for-an-http-delete-request](https://stackoverflow.com/questions/299628/is-an-entity-body-allowed-for-an-http-delete-request)
> The spec does not explicitly forbid or discourage it, so I would tend to say it is allowed.

As per Quote from SO: request body is not forbidden according to Spec.
Can you please allow it?

WoH:
Closes #362, #371

### DIFF
--- a/src/metadataGeneration/parameterGenerator.ts
+++ b/src/metadataGeneration/parameterGenerator.ts
@@ -178,11 +178,11 @@ export class ParameterGenerator {
   }
 
   private supportBodyMethod(method: string) {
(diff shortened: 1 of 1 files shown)


dgreene1:
needs a rebase

WoH:
rebased and ready

PR #425, opened 2019-08-17
Clear type resolver cache

### LINKED ISSUE #417: Internal typeResolver cache causes issues when running generateRoutes before generateSwaggerSpec

joni7777:
FIX #417


src/metadataGeneration/metadataGenerator.ts
```diff
@@ -15,6 +16,7 @@ export class MetadataGenerator {
   public IsExportedNode(node: ts.Node) { return true; }
 
   constructor(entryFile: string, private readonly compilerOptions?: ts.CompilerOptions, private readonly ignorePaths?: string[], controllers?: string[]) {
+    TypeResolver.clearCache();
```
dgreene1: @lukeautry this is the part I could use guidance on.

@joni7777 do we really want to clear the cache every time? Doesn’t that defeat the purpose of the cache? I feel like it would be best to only clear the cache if the routes generation was run before the swagger generation. So maybe we need to record the CLI command that was used when the cache is populated?

joni7777:
I think we should start new cache for every new program, the cache should be while exploring the same program

src/metadataGeneration/metadataGenerator.ts
joni7777: I think we should start new cache for every new program, the cache should be while exploring the same program

Issue #399, opened 2019-08-11
Nested type is being ignored


Asterix112:
When writing a request handler returning a nested type you will face a type name generation issue within the OpenAPI specification.

Example:
```
public testGet(): ResponseWrapper<Array<T>> {}
```

The name generated will be:
```
ResponseWrapperArray
```
But is expected to be:
```
ResponseWrapperArrayT
```

dgreene1:
Same as the other issue you opened:

Before I try to reproduce this, please include

the full controller file that uses this model
the interface model definition
the Swagger YAML that is produced
the Swagger YAML that you wish was produced

The reason why I’m asking this is because I don’t see how that would compile without giving testGet a generic parameter. Additionally, it doesn’t make sense to me to have a generic route because the model it returns should be a definite type.

dgreene1:
@Asterix112 it's been 5 days and we still haven't received any steps to reproduce this. Please see above. Due to that we'll close this for housekeeping purposes. However, if you are able to submit the information above we can re-open the issue and try to solve it. Thank you.

Asterix112:
@dgreene1 Sorry for my late response. Because of project protection purposes i cannot send the whole controller to you. I could send you everything that is directly in touch with the problem we are facing (e.g. controller function including annotations and the controller class itself, also including annotations and of course the interfaces)

Controller:
```typescript
@Route('')
@injectable()
export class DefaultController extends Controller {
    @inject(LoggingServiceImpl)
    private logging: LoggingServiceImpl;

    @inject(CachingServiceImpl)
    private caching: CachingServiceImpl;

    constructor() {
        super();
    }

    /***
     * Retrieve the current substitution plan blob by moduleId
     * @param request
     * @param moduleId Module ID obtained
     */
    @Get('subst/{moduleId}')
    @Security('sia', ['sia'])
    public async substGet(@Request() request: ExpressRequest, @Path() moduleId: string): Promise<GenericResponse<Array<Substitutionplan>>> {
        this.logging.info(`User: ${JSON.stringify(request.user)}`);

        this.logging.info("Retrieving substitution plan blob from caching service...");
        const data = <SubstitutionplanMap><any>(await this.caching.getBlob(CacheBlobType.SUBSTITUTION_PLAN, request.user["aclid"], moduleId));
        this.logging.debug(`Data:\n${JSON.stringify(data)}`);

        const result = new Array<Substitutionplan>();
        for (const e of Object.keys(data)) {
            result.push(data[e]);
[...]

PR #427, opened 2019-08-18
Correctly handle generic model name resolution

WoH:
#399

### DIFF
--- a/src/metadataGeneration/typeResolver.ts
+++ b/src/metadataGeneration/typeResolver.ts
@@ -309,7 +309,16 @@ export class TypeResolver {
 
   private getTypeName(typeName: string, genericTypes?: ts.NodeArray<ts.TypeNode>): string {
     if (!genericTypes || !genericTypes.length) { return typeName; }
-    return typeName + genericTypes.map((t) => this.getAnyTypeName(t)).join('');
+
+    const resolvedName = genericTypes.reduce((acc, generic) => {
+      if (ts.isTypeReferenceNode(generic) && generic.typeArguments && generic.typeArguments.length > 0) {
+        return [...acc, this.getTypeName(generic.typeName.getText(), generic.typeArguments)];
+      } else {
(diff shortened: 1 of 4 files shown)


src/metadataGeneration/typeResolver.ts
```diff
@@ -309,7 +309,16 @@ export class TypeResolver {
 
   private getTypeName(typeName: string, genericTypes?: ts.NodeArray<ts.TypeNode>): string {
     if (!genericTypes || !genericTypes.length) { return typeName; }
-    return typeName + genericTypes.map((t) => this.getAnyTypeName(t)).join('');
+
+    const resolvedName = genericTypes.reduce((acc, generic) => {
+      if (ts.isTypeReferenceNode(generic) && generic.typeArguments && generic.typeArguments.length > 0) {
+        return [...acc, this.getTypeName(generic.typeName.getText(), generic.typeArguments)];
```
dgreene1: @WoH I mentioned this in another recent PR, but there's a performance hit for using the spread operator since you are iterating through the `acc` array on every loop of the  `reduce`. The implementation of the `reduce` could still be a pure function if it mutates the seed since the seed is not considered to be part of the inputs. In fact, it's much more standard in what I've seen to just use `acc.push`. Now I'm aware that `genericTypes` probably has a very small length, so the performance advantages are probably minimal, but I also think that `.push` is more readable. Here's the JSPerf from last time to demonstrate the performance advantages: https://jsperf.com/reduce-with-new-set-vs-seed

dgreene1: Rather than have you make this `acc.push` update @WoH, I wanted to be considerate of the number of excellent contributions you've made. So I've made the update as part of https://github.com/lukeautry/tsoa/pull/428 which fixes the other issue that Asterix presented to us. So I'll merge this PR so I can fold it into mine. Sound good? :)

PR #428, opened 2019-08-19
fixes #398 and adds more tests to prove #399

### LINKED ISSUE #398: Method return type problem when dealing with arrays
If you define a request handler and want to return an array of some type you will face a type naming issue in the OpenAPI specification when writing
```
public testGet(): T[] {}
```
instead of
```
public testGet(): Array<T> {}
```

If you then try to generate an API client using the Swagger codegen executable you will face the issue of return types like this (Dart):
```
static List<T[]> listFromJson(List<dynamic> json) {}
```
Expected behavior:
```
static List<T> listFromJson(List<dynamic> json) {}
```

dgreene1:
fixes #398 and adds more tests to prove #399

### DIFF
--- a/src/metadataGeneration/typeResolver.ts
+++ b/src/metadataGeneration/typeResolver.ts
@@ -121,6 +121,16 @@ export class TypeResolver {
     }
 
     this.current.AddReferenceType(referenceType);
+
+    // We do a hard assert in the test mode so we can catch bad ref names (https://github.com/lukeautry/tsoa/issues/398).
+    // This regex allows underscore, hyphen, and period since those are valid in SwaggerEditor
+    const symbolsRegex = /[!$%^&*()+|~=`{}\[\]:";'<>?,\/]/;
+    if (symbolsRegex.test(referenceType.refName)) {
+      throw new Error(`Problem with creating refName ${referenceType.refName} since we should not allow symbols in ref names ` +
+        `because it would cause invalid swagger.yaml to be created. This is due to the swagger rule ` +
+        `"ref values must be RFC3986-compliant percent-encoded URIs."`);
+    }
+
     return referenceType;
   }
 
@@ -322,7 +332,9 @@ export class TypeResolver {
       }
     }, [] as string[]);
 
-    return typeName + resolvedName.join('');
+    const finalName = typeName + resolvedName.join('');
+
+    return finalName;
   }
 
   private getAnyTypeName(typeNode: ts.TypeNode): string {
--- a/src/metadataGeneration/typeResolver.ts
+++ b/src/metadataGeneration/typeResolver.ts
@@ -123,12 +123,16 @@ export class TypeResolver {
     this.current.AddReferenceType(referenceType);
 
     // We do a hard assert in the test mode so we can catch bad ref names (https://github.com/lukeautry/tsoa/issues/398).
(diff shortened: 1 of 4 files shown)


src/metadataGeneration/typeResolver.ts
```diff
@@ -329,7 +333,7 @@ export class TypeResolver {
 
     if (typeNode.kind === ts.SyntaxKind.ArrayType) {
       const arrayType = typeNode as ts.ArrayTypeNode;
-      return this.getAnyTypeName(arrayType.elementType) + '[]';
+      return this.getAnyTypeName(arrayType.elementType) + 'Array';
```
dgreene1: @WoH this is the change that I thought could be breaking for some users if they were relying on getting back `GenericRequestTypeAliasModel1[]` they would now get `GenericRequestTypeAliasModel1Array`

WoH: I see. But I have more of these, if you're not sure about breaking changes, I think we can make it worth the 3.0 release :wink:

tests/unit/swagger/schemaDetails3.spec.ts
```diff
@@ -359,11 +359,20 @@ describe('Definition generation for OpenAPI 3.0.0', () => {
...
                     expect(propertySchema.nullable).to.eq(true, `for property ${propertyName}.nullable`);
                 },
-                genericNested: (propertyName, propertySchema) => {
-                    expect(propertySchema.$ref).to.eq('#/components/schemas/GenericRequestArrayTypeAliasModel1');
+                genericMultiNested: (propertyName, propertySchema) => {
+                    expect(propertySchema.$ref).to.eq('#/components/schemas/GenericRequestGenericRequestTypeAliasModel1', `for property ${propertyName}.$ref`);
                 },
-                genericNested2: (propertyName, propertySchema) => {
-                    expect(propertySchema.$ref).to.eq('#/components/schemas/GenericRequestArrayTypeAliasModel2');
+                genericNestedArrayKeyword1: (propertyName, propertySchema) => {
```
dgreene1: @WoH I had to rename `genericNested2` to differentiate between the `[]` tests.  I also added `genericMultiNested` to see if we would have an issue with multiple levels of generics. Turns out that we don't because your function is recursive. Nice work! :)

src/metadataGeneration/typeResolver.ts
dgreene1: @WoH would I be overthinking it if I added a swagger.config parameter called `noSymbolsInRefNames` that would allow @Asterix112 (and anyone else who has stricter needs for their ref names) to opt into this?

And then when we get to version 3.0 we can remove this `noSymbolsInRefNames`.

The reason why I'm resistant to going to a 3.0 version right now is (a) Asterix appears to need this soon and (b) there are a lot of great features that we can release without going to 3.0 (like your union PR, and the PR that fixes auto controller generation).

I appreciate any and all thoughts. Thank you for this collaboration @WoH. It's a lot of fun! :)

WoH: [ ] in the $ref wasn't ever valid, was it?

dgreene1: Wow, you are correct @WoH. It gives this error in SwaggerEditor: `ref values must be RFC3986-compliant percent-encoded URIs`.

So we're safe to make this change since it's a true bug fix.

Issue #426, opened 2019-08-17
Model validation throws when key of a declared nested property is missing.


WoH:
Model used:
```typescript
export interface TestModel {
  typeAliasCase1?: MyInterface;
}

export interface MyInterface {
  value1: string;
}

```

Reproduction (express-server.spec.ts)

```typescript
  it('should reject invalid values', () => {
    const invalidValues = ['test', null, undefined, 123];

    return Promise.all(invalidValues.map((value: any) => {
      const data = getFakeModel();
      data.typeAliasCase1 = value;

      return verifyPostRequest(basePath + '/PostTest', data, (err: any, res: any) => null, 400);
    }));
  });
```

All of these will throw on the server and exit with a 500 because [`validateModel()`](https://github.com/lukeautry/tsoa/blob/master/src/routeGeneration/templateHelpers.ts#L365) assumes too much:

```ts
// Model is MyInterface
keysOnPropertiesModelDefinition.forEach((key: string) => {
  // key will be value1
  // property refers to the properties of the Server's schema of MyInterface
  const property = properties[key]; // => { "dataType": "string", "required": true }
  // here is the issue: value[key] will throw
  // because it's not an object, i.e. null['value1'], 123['value1']
  value[key] = this.ValidateParam(property, value[key], key, fieldErrors, parent, swaggerConfig);
});
```

dgreene1:
Would you like this merged before or after your intersection/union PR? In fact, if you could write down the order you want all of your PRs merged, that would be helpful. I don’t want you to have to do more rebasing/merge-resolving than necessary. :)

WoH:
This one should be first since I'd like to add integration tests for unions based on this behavior before the union pr is ready to be merged.

PR #432, opened 2019-08-20
Fix Model validation if value is not an Object

WoH:
#426 

@dgreene1 I believe recent commits to master broke typescript compilation because Array.includes() which requires es7 as lib in the tsconfig.json is used. I can add it to this PR if you want me to.

```
TSError: ⨯ Unable to compile TypeScript
src/utils/mutualConfigValidation.ts (21,42): Property 'includes' does not exist on type 'string[]'. (2339)
```

### DIFF
--- a/src/routeGeneration/templateHelpers.ts
+++ b/src/routeGeneration/templateHelpers.ts
@@ -381,6 +381,14 @@ export class ValidationService {
         return this.validateEnum(name, value, fieldErrors, enums, parent);
       }
 
+      if (!(value instanceof Object)) {
+        fieldErrors[parent + name] = {
+          message: `invalid object`,
+          value,
+        };
+        return;
+      }
+
       const properties = modelDefinition.properties || {};
       const keysOnPropertiesModelDefinition = new Set(Object.keys(properties));
       const allPropertiesOnData = new Set(Object.keys(value));
--- a/tests/fixtures/testModel.ts
+++ b/tests/fixtures/testModel.ts
@@ -258,6 +258,8 @@ export class ValidateModel {
    * @ignore
    */
   public ignoredProperty: string;
+
+  public model: TypeAliasModel1;
 }
 
 export interface ValidateMapStringToNumber {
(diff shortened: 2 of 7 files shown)


dgreene1:
> I believe recent commits to master broke typescript compilation because Array.includes() which requires es7 as lib in the tsconfig.json is used. I can add it to this PR if you want me to.

Array.includes is part of es6 I thought. And we have the es6 lib in tsconfig. Also, I don’t have any TypeScript errors when I compile locally. Also, CodeShip is passing. So I’m not sure why you’re seeing that.

src/routeGeneration/templateHelpers.ts
```diff
@@ -381,6 +381,14 @@ export class ValidationService {
         return this.validateEnum(name, value, fieldErrors, enums, parent);
       }
 
+      if (!(value instanceof Object)) {
+        fieldErrors[parent + name] = {
+          message: `Not an Object`,
```
dgreene1: The error message of `Not an object` is not very descriptive. If I was a consumer of an API that returned that error message, I would be quite confused. Maybe it could simply be `Expected an object for this property but received ${typeof value} instead`.

WoH: I agree. Since the other validations are rather brief as well, examples include "invalid boolean value", "invalid array" and "invalid string value", I think there is an opportunity to add some more convenience that may be shared.
For now, is it fine if I were to be closer to the other messages and return 'invalid object'?

dgreene1: Not a problem. I’m happy to get this moving to help you with your other PRs. :)

dgreene1: Not a problem. I’m happy to get this moving to help you with your other PRs. :)

WoH: I'll create an issue and revisit this, thanks for accepting a mediocre message for now.

Issue #311, opened 2018-11-18
Should typescript be a peer dependency?


lukewis:
My project is using Typescript 3.1.6.  My tsconfig.json includes the resolveJsonModule compiler option, however tsoa threw the following error:
```
Error: Unknown compiler option 'resolveJsonModule'.
```
It appears that tsoa lists typescript ^2.8.4 as a dependency.  Indeed, typescript 2.8.4 doesn't support the resolveJsonModule option, but it seems that tsoa should be picking up and using the typescript version my project is using.  I think this behavior would be accomplished if tsoa listed typescript as a peerDependency instead of a "real" dependency.

Thoughts?

lukeautry:
I think we've had it as a peer dependency in the past, but this project uses the TypeScript compiler API directly. The compiler API has breaking changes regularly, so the library would just not work a lot of the time if the peer dependency is the wrong version of TypeScript.

PR #434, opened 2019-08-21
Full dependency update

HoldYourWaffle:
Title says it all. This also fixes an out-of-sync `package-lock.json` file.


HoldYourWaffle:
As requested in #438 I did a full dependency update, which went pretty painless (just some confusion about the `hapi` → `@hapi/hapi` move for a non-`hapi` user).
I also pinned down the typescript version to an exact version for the reasons cited in #311.

I'd recommend installing [greenkeeper](https://github.com/apps/greenkeeper) to make this process automatic, which saves the trouble of having to manually test (major version) dependency updates.

The only thing still left to do is to prohibit/auto-delete a `yarn.lock` file (for my reasons in #438). What do you think would be an appropriate solution for this?

dgreene1:
>I also pinned down the typescript version to an exact version for the reasons cited in #311.
There's no reason that we need have a specific version of TypeScript in the package.json file. Once we're all using the same lockfile (which I will add in my PR), then we will get the version of TypeScript from the lockfile. I'll update to a semver pattern in my PR.

> I'd recommend installing greenkeeper to make this process automatic, which saves the trouble of having to manually test (major version) dependency updates.

Good idea. I'll message Luke and ask him to get that set up.

> The only thing still left to do is to prohibit/auto-delete a yarn.lock file (for my reasons in #438). What do you think would be an appropriate solution for this?

I'll take care of that.

PR #437, opened 2019-08-22
version bump

dgreene1:


tslint.json
```diff
@@ -4,6 +4,7 @@
   ],
   "rules": {
     "arrow-parens": false,
+    "object-literal-sort-keys": false,
```
dgreene1: @WoH did you skip the husky hook or something? When I went to deploy your changes I got a bunch of linter errors. Anyway, I'm going to disable this rule because it's a real pain. And it was much faster than changing these 11+ places where the linter was throwing.

package.json
```diff
@@ -34,7 +34,7 @@
     "preversion": "npm test",
     "postversion": "git push origin master && git push --follow-tags",
     "pretest": "cross-env NODE_ENV=tsoa_test ts-node ./tests/prepare.ts",
-    "test": "tsc -p ./tests/tsconfig.json --noEmit && cross-env NODE_ENV=tsoa_test mocha **/*.spec.ts --exit --require ts-node/register --timeout 3000",
+    "test": "tsc -p ./tests/tsconfig.json --noEmit && cross-env NODE_ENV=tsoa_test mocha **/*.spec.ts --exit --require ts-node/register --timeout 5000",
```
dgreene1: Maybe it's because my laptop is slow, but mocha keeps timing out, so I've set it to be 5000 so I could actually publish the package.

tslint.json
WoH: I won't pretend I haven't skipped it in my life ever. But I did run it before in this case because I know I miss way too many semicolons and brackets due to my own stylistic preferences.

Right now, `npm run lint` still passes on my feature branch (which has fb1f5231978893c73b3fcdd22118254f4bbf209c as the latest commit) :confused:

WoH: Are you sure you are running the packaged version of tslint and not your global installation or something?

Issue #438, opened 2019-08-22
Current master doesn't compile on TS 3.2.2


HoldYourWaffle:
fixes #438
fixes #434
While fixing up #361 I encountered the following compilation error:
```
src/routeGeneration/templateHelpers.ts:466:36 - error TS2322: Type 'ModelSchema' is not assignable to type 'boolean'.
```

This project also occurs on the `master` branch and is *probably* caused by https://github.com/lukeautry/tsoa/commit/8344f9f8c7f6cf90ff039e05f0a5f8bd8cafb6b9 (at least that's what `git blame` tells me).

The error occurs in this piece of code:
https://github.com/lukeautry/tsoa/blob/8344f9f8c7f6cf90ff039e05f0a5f8bd8cafb6b9/src/routeGeneration/templateHelpers.ts#L439-L442
`this.models[refName]` is a `ModelSchema` but a `boolean` was expected. I think this is supposed to be a 'truthy' comparison, which made me think that this is actually a bug with Typescript.

After investigating why I didn't get this error before and especially how it passed the pre-commit hook. I found out that it doesn't occur when using TS 3.5.1 (which was installed in my clone due to #434). This is supported by the yarn.lock which lists TS 3.5.1 as the "actually installed version":
https://github.com/lukeautry/tsoa/blob/ecdd050c6859112f080e55e88f141624a86e8c90/yarn.lock#L2224-L2225

There's a second error caused by the lack of ES7 definitions (`includes` is not found on `string[]`). I encountered this problem with #434, it's solved by adding `es7` to the `lib` section in `tsconfig.json`. This error still occurs even when using TS 3.5.1, and I have no idea how that passed the pre-commit hook.

As a side note, shouldn't the Typescript version in `package.json` be locked down to an exact version (or maybe allow only patch versions), for the reason cited in #311?

WoH:
I also experienced the issue with `Array.includes()` and believe I mentioned it, can't say where at the top of my head. But can confirm.
As the author of the lines in question, yes, that is a truthy comparison to ensure every subschema references a type tracked in the referenceMap :)

HoldYourWaffle:
The truthy comparison should be fine, that was probably just a bug in tsc.

The `includes` error is strange though. In the issue you linked @dgreene1 mentions that `Array.includes` should be in `es6` which I thought was the case too. Adding `es7` to the `tsconfig.json` did solve the issue for me though.
[...]

PR #439, opened 2019-08-23
enforces that contributors use yarn not npm

dgreene1:
This will be the true fix to #438 which recognized that some contributors were getting different errors based on if they were using package-lock.json or yarn.lock. This is what was changed:

* We will only use yarn.lock
* package-lock.json was removed and if it comes back then ensureYarn.js will throw and fail the tests (and the commit thanks to Husky)


dgreene1:
Check this PR out @HoldYourWaffle & @WoH. I'm going to merge it so you guys can try out how it behaves. Note that you will have to either install yarn or upgrade it to the version specified in the `"engines"` part of `package.json`

HoldYourWaffle:
Are you sure you want to enforce yarn instead of npm? Npm is a lot more widespread (and included with node.js), so I don't think it's contributor friendly to outright prohibit it.

Yarn can cooperate with npm's `package-lock.json` just fine (albeit with a warning), while npm doesn't understand the yarn format at all. In other words: as long as we make sure yarn never actually uses the `yarn.lock` file (for example by deleting it) we'll never get inconsistent results (since it'll use the repository-provided `package-lock.json`), meaning there's no reason to outright block using one or the other.

By the way I locked the Typescript version because even minor (or sometimes even patch) versions can (and have done so in the past) break the compiler api. Even with a functioning lockfile this could lead to issues by careless updating. It's more an insurance than an actual solution for a problem though, so if you want to leave it unlocked that's totally fine by me.

dgreene1:
> Are you sure you want to enforce yarn instead of npm? Npm is a lot more widespread (and included with node.js), so I don't think it's contributor friendly to outright prohibit it.

Yes. tsoa always did use yarn. That's what is run in CodeShip.

> Yarn can cooperate with npm's package-lock.json just fine (albeit with a warning), while npm doesn't understand the yarn format at all.

If we're using yarn (which tsoa always has been), then we should use the file that yarn expects.

> By the way I locked the Typescript version because even minor (or sometimes even patch) versions can (and have done so in the past) break the compiler api. Even with a functioning lockfile this could lead to issues by careless updating. It's more an insurance than an actual solution for a problem though, so if you want to leave it unlocked that's totally fine by me.

Unless I'm mistaken, I believe we will only have breaks in the compiler api if we go backwards. And the lockfile will ensure that that doesn't happen. As far as moving forward, tsoa should continue to use the latest typescript so that we're not out of sync with what our users are getting with VSCode (which automatically upgrades TypeScript).

HoldYourWaffle:
> If we're using yarn (which tsoa always has been), then we should use the file that yarn expects.

I disagree with your "hard logic" (English is not my first language...) here. Although yarn doesn't specifically "expect" `package-lock.json`, it's able to handle it perfectly fine. Unless yarn has some incredible feature that I don't know about (in which case I'd be very eager to know), I don't see how a little warning justifies banning the most popular (and bundled) package manager.
This doesn't apply if there's a safety (in the sense of dependency locking) benefit of course, but as far as I know there is none.
"It's what we've always used" doesn't sound like a good argument to me either, since you can continue using yarn just fine with `package-lock.json`.

> Unless I'm mistaken, I believe we will only have breaks in the compiler api if we go backwards

I have unfortunately seen it break even when moving forward :( Again if you want to leave it unlocked that's perfectly fine by me, it's just a suggestion on my end.

WoH:
>  I don't see how a little warning justifies banning the most popular (and bundled) package manager.

The project uses yarn and it should make sure the issues we both noticed do not impact anyone else? I don't see how that's really banning. It's like the linting, in order to make sure the code is consistent, you need to make decisions that every person working on it may not always agree on. I don't think you'd say linter rules are a ban even if they equally are.

HoldYourWaffle:
I don't see how enforcing a certain "visual style" for the code is the same as "you have to use this specific piece of software, that does exactly the same as the one you're familiar with (and is probably already installed on your system) for no reason other than a little warning for those that use our choice".

The issues we experienced shouldn't happen anymore because `yarn` **supports** `package-lock.json`. Unless there's a difference in how it handles this format (or some other consistency problem I don't know about), I honestly don't see a reason why we'd have to ban npm. If there is such a reason please tell me, because I'm really starting to feel like I'm missing something obvious here 😅

WoH:
> you have to use this specific piece of software, that does exactly the same as the one you're familiar with (and is probably already installed on your system) for no reason other than a little warning for those that use our choice

 -> you have to use this specific type of code format, that does exactly the same as the one you're familiar with (and is probably already what you understand more easily) for no reason other than a little warning for those that use our choice.

dgreene1:
>  in order to make sure the code is consistent, you need to make decisions that every person working on it may not always agree on.

Yes, thank you @WoH. I did not take this decision lightly. I did it for one pragmatic reason:

I have no ability to modify the CodeShip configuration since that is owned by Luke (who I am purposely not tagging). I've already asked him to give me access to modify the CodeShip build steps. But since I can not do that yet, we are have no choice but to use yarn.

> Although yarn doesn't specifically "expect" package-lock.json, it's able to handle it perfectly fine.

You're making an assumption that yarn works perfectly fine. I would wager that the majority of yarn's automated unit tests verify their interoperability with their first party item (yarn.lock). As the primary maintainer of this library and the npm module, I take a lot of responsibility with ensuring quality of the product. For me, I rest easier at night using yarn in it's intended manner. If I'm reading an instruction manual and it says, "Please only use the authorized batteries" then I do not put batteries off of ebay into the device. Some are comfortable with that-- I am not.

> I don't see how a little warning justifies banning the most popular (and bundled) package manager.
This doesn't apply if there's a safety (in the sense of dependency locking) benefit of course, but as far as I know there is none.

Benefits of yarn:
* yarn is much faster than npm
* yarn has workspace support
* yarn has always had a much more reliable lock file. I have been monitoring this even since the days of npm's "shrinkwrap" command back in 2013. Even with npm fighting back with package-lock.json, yarn has had more deterministic results

HoldYourWaffle:
That makes a lot of sense, and with that additional information in mind I agree that it would at least make sense to "ban" npm. I'd say it's little "over-safe", but that's not really my place to talk about. Thanks for clarifying your viewpoint.

As for your original request for me to try it out: seems to work fine, although you should probably add some kind of message if it fails because `package-lock.json` exists. It saying "you should use yarn" when I was in fact using yarn confused me a little.

There's also this weird thing where it shows the `echo` command and then the output, not sure if that's fixable but it's a little annoying (nothing important, just a little nitpick).
Double echo output

PR #441, opened 2019-08-23
Adding prettier so it's easier for contributors to contribute

dgreene1:
**Context:**

It's [rather frustrating](https://github.com/lukeautry/tsoa/pull/385#discussion_r311530902) to commit to the tsoa repository since it uses 2 space tabs and most editors default to 4 space tabs. But the internet has solved that "tabs vs. spaces" argument along time ago when [Prettier](https://prettier.io/) came to prominence in the JS/TS community. It's now used by 0.9+ million repositories.

**Solution:**

* Prettier now runs on every commit and fixes:
  * fixes any tab that isn't 2 spaces
  * adds semicolons wherever it can
  * adds dangling commas wherever it can
  * adds/removes whitespace to be consistent
* We no longer need those rules to run in tslint so they've been set to false
* I happened to also fix the tslint config since there were other issues

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
   },
   "lint-staged": {
     "*.{js,ts}": [
-        "prettier --print-width 200 --tab-width 2 --single-quote true --no-semi false --trailing-comma all --write",
+        "prettier --write",
         "git add"
     ]
   },
--- a/tests/unit/swagger/schemaDetails3.spec.ts
+++ b/tests/unit/swagger/schemaDetails3.spec.ts
@@ -1,148 +1,148 @@
-import { expect } from 'chai'
-import 'mocha'
-import { SwaggerConfig } from '../../../src/config'
-import { MetadataGenerator } from '../../../src/metadataGeneration/metadataGenerator'
-import { SpecGenerator3 } from '../../../src/swagger/specGenerator3'
-import { Swagger } from '../../../src/swagger/swagger'
-import { getDefaultOptions } from '../../fixtures/defaultOptions'
-import { TestModel } from '../../fixtures/duplicateTestModel'
+import { expect } from 'chai';
+import 'mocha';
+import { SwaggerConfig } from '../../../src/config';
+import { MetadataGenerator } from '../../../src/metadataGeneration/metadataGenerator';
+import { SpecGenerator3 } from '../../../src/swagger/specGenerator3';
+import { Swagger } from '../../../src/swagger/swagger';
+import { getDefaultOptions } from '../../fixtures/defaultOptions';
+import { TestModel } from '../../fixtures/duplicateTestModel';
 
 describe('Definition generation for OpenAPI 3.0.0', () => {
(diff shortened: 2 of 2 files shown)


HoldYourWaffle:
This is a very welcome improvement, but it's only a half-solution in my opinion. Most editors are able to detect the used indentation in a file and change their settings accordingly (the `.editorconfig` file helps with this).

The big frustration when working with TSOA for me, is that the indentation is *only* 2 spaces. It could be just me, but that's just not enough for me to (easily) spot where functions/conditionals/parameters/other kinds of blocks start and end. It makes navigating (especially while (re)learning the codebase) really hard and annoying. This is also the reason why I prefer tabs, since you can change the width of those to your personal preference, but that's definitely not a discussion for today.

I suggest we change the indentation from 2 to 4 spaces (tabs would be even better but I'm not going to go there right now 😉). This makes it **a lot** easier to (quickly) navigate the code, and to be honest I don't really see a reason why not (but I'm happy to hear your reasons for it, I've always been curious).

If you agree with me (and I really hope you do because it would make my life a lot easier) this would probably be the time to do the change. You just did https://github.com/lukeautry/tsoa/commit/efd88091aa669349ec98ccfd15ab00edee3f8ddc meaning **everyone** will have to solve conflicts anyway (I understand the need, I'm just a tiny bit salty because I *just* finished rebasing), if we wait too long people will have to solve conflicts twice.

Alternatively: is there a way I can get VS code to render 2-space indentation as 4-space? This would solve the issue without requiring a full re-format, but I haven't been able to find such a thing.

dgreene1:
>  suggest we change the indentation from 2 to 4 spaces (tabs would be even better but I'm not going to go there right now 😉)

I would be fine with 4 space sized tabs. @lukeautry is there any technical requirement for why we were using 2 spaces?

> You just did efd8809 meaning everyone will have to solve conflicts anyway

Since I've worked very hard over the last few weeks to merge or close PRs, there simply aren't that many open PRs. And most of those open PRs are multiple years old, so they have much worse conflicts than just tab sizes.

> Alternatively: is there a way I can get VS code to render 2-space indentation as 4-space? This would solve the issue without requiring a full re-format, but I haven't been able to find such a thing.

This is what makes Prettier so awesome. If you want to see the code as tabs, just convert the whole file that you're working on to 4 space tabs by pressing F1 then typing in `Format Documnent`. And then you have tabs. But when you save the file, Prettier will make it look like the rest of the code.

HoldYourWaffle:
> Since I've worked very hard over the last few weeks to merge or close PRs, there simply aren't that many open PRs

Your hard work is very much appreciated 😄 

> This is what makes Prettier so awesome. If you want to see the code as tabs, just convert the whole file that you're working on to 4 space tabs by pressing F1 then typing in Format Documnent. And then you have tabs. But when you save the file, Prettier will make it look like the rest of the code.

Interesting... I'll have to look into that, thanks!

Issue #429, opened 2019-08-19
Discussion: Type Aliases should be aliases, not references


WoH:
Tl;Dr: Treat type aliases like symlinks, not like references (models/classes).

From the [Typescript handbook](https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-aliases) on interfaces vs. type aliases:
> One difference is that interfaces create a new name that is used everywhere. Type aliases don’t create a new name — for instance, error messages won’t use the alias name. 

TSOA currently supports a very narrow set of type aliases: Interface aliasing and Intersections of Interfaces/Classes (where it collects the properties of all members).

Given #400, I would like to make aliases exactly what they are, convenient pointers to other types (unions, intersections, literals etc.) and treat them more or less like as a symlink.

However, this would break some swagger component/definition names (In my mind, the new representation is better, that's why I would prefer not to change it):

`#/components/schemas/Namespace_InnerNamespace_TestModel`
would become
`#/components/schemas/Namespace.InnerNamespace.TestModel`

(OpenAPI3, Swagger is similar)

Furthermore, there would no longer be a definition/component for the alias, the content would be inlined. 

Thoughts?

dgreene1:
This looks like a question for @lukeautry

lukeautry:
In the current framework, yeah, we don't *really* support type aliases, and the way that we do support them is as a pointer to an interface or a class, so I think this is reasonable.

I actually think full support for type aliases makes sense though, i.e. discriminated unions. It's possible that this has been implemented, but to my knowledge it has not.

If we have, say:

```
type OperationResult = IOperationSuccess | IOperationError;

interface IOperationSuccess {
  success: true;
  value: number;
}

interface IOperationError {
  success: false;
  message: string;
}
```

...the end user is probably thinking about `OperationResult` as a named data structure, not a simple pass through. But, in terms of what actually gets generated, the above would be a union type in Swagger, and in the routes would be handled similarly. The only place where one might expect to see `OperationResult` is in error messages, e.g. if we fail to resolve the type alias for some reason.

WoH:
Somewhat related: #204.
We could get rid of the issue by treating the alias like a pointer to a union, which can be properly resolved on #400 
For ppl that prefer to look at it: https://github.com/WoH/tsoa/commit/d5abad9fd1825eb073a26c9d4a7b104172424a8c

Note that some of the fixed tests refer to new properties on the TestModel and therefore would not be considered a breaking change in any case.

[...]

PR #444, opened 2019-08-23
Refactor type aliases

WoH:
**Historic context:**
As far as I can tell, resolving type aliases was introduced to support a narrow set of intersection types. 2 interfaces were scanned for properties and merged together.

**Why may this not make sense anymore?**
Now that we can properly represent the metadata of an intersection, we can use that to support the (narrowly defined) Swagger 2 representation of 2 interfaces.
However, we should eliminate that code path from type resolving and handle type aliases "properly": #429

**What's the cost?**
This would break some swagger component/definition names (In my mind, the new representation is better and consistent with nested interfaces, that's why I would prefer the new way):

`#/components/schemas/Namespace_InnerNamespace_TestModel`
would become
`#/components/schemas/Namespace.InnerNamespace.TestModel`

(OpenAPI3, Swagger is similar)

Furthermore, there would no longer be a definition/component for the alias, the content would be inlined.

**What's the benefit?**
We remove a legacy code path from the typeResolver that does not make sense anymore.
We get more consistent naming with fewer code.
We avoid issues with type aliases themselves (we forward the resolution responsibility to the thing referenced by the type alias)
We are closer to the TypeScript internals (See #429) and it would make it easier to support i.e. TS convenience types.
Why?: Pick, Omit, ..., all the convenience types are type aliases.
If we want to support them at a point in time, the way to do that would be to resolve the mapped type py passing the arguments to the type alias to the mapped type.
```
type Pick<T, K extends keyof T> = {
    [P in K]: T[P];
};
```

### DIFF
--- a/src/metadataGeneration/typeResolver.ts
+++ b/src/metadataGeneration/typeResolver.ts
@@ -139,9 +139,14 @@ export class TypeResolver {
       return literalType;
     }
 
-    const possibleTypeAlias = this.getModelTypeDeclaration(typeReference.typeName);
-    if (ts.isTypeAliasDeclaration(possibleTypeAlias)) {
-      return new TypeResolver(possibleTypeAlias.type, this.current, this.typeNode, this.extractEnum).resolve();
+    const typeAlias = this.getModelTypeDeclaration(typeReference.typeName);
+    if (ts.isTypeAliasDeclaration(typeAlias)) {
+      if (typeAlias.typeParameters && ts.isTypeReferenceNode(typeAlias.type)) {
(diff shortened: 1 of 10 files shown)


HoldYourWaffle:
Since I'm currently cleaning up the type resolver a bit (#445), we probably shouldn't merge this until that's merged to avoid conflicts.
(That is assuming the maintainers want my cleanup merged)

WoH:
Closing this in favor of another PR with a more advanced implementation based on https://github.com/lukeautry/tsoa/issues/429#issuecomment-532718400 (currently blocked by #470)

PR #446, opened 2019-08-23
Reformat the code from 2 to 4 spaces

HoldYourWaffle:
A PR that reformats the code to 4 space indentation as discussed in #441, for your convenience 😉
It felt really good to look around the code without having to concentrate on where a certain block ends, I think this is definitely a good change.

If @lukeautry decides that this isn't a good idea this can of course be closed.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -6,9 +6,9 @@ insert_final_newline = true
 
 [*.ts]
 indent_style = space
-indent_size = 2
+indent_size = 4
 trim_trailing_whitespace = true
 
 [*.hbs]
 indent_style = space
-indent_size = 2
+indent_size = 4
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,6 +1,6 @@
 {
     "trailingComma": "all",
-    "tabWidth": 2,
+    "tabWidth": 4,
     "semi": true,
     "singleQuote": true,
     "printWidth": 200
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -12,16 +12,16 @@ const workingDir: string = process.cwd();
 
 let packageJson: any;
 const getPackageJsonValue = async (key: string, defaultValue = ''): Promise<string> => {
-  if (!packageJson) {
-    try {
-      const packageJsonRaw = await fsReadFile(`${workingDir}/package.json`);
-      packageJson = JSON.parse(packageJsonRaw.toString('utf8'));
-    } catch (err) {
-      return defaultValue;
+    if (!packageJson) {
+        try {
+            const packageJsonRaw = await fsReadFile(`${workingDir}/package.json`);
+            packageJson = JSON.parse(packageJsonRaw.toString('utf8'));
+        } catch (err) {
+            return defaultValue;
+        }
     }
-  }
 
-  return packageJson[key] || '';
+    return packageJson[key] || '';
 };
 
 const nameDefault = () => getPackageJsonValue('name', 'TSOA');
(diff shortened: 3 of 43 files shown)


HoldYourWaffle:
I will rebase this once there's confirmation that this is a welcome change

lukeautry:
Shouldn't the conversion to prettier have solved this? Did we not run a mass --fix on the codebase at that time?

HoldYourWaffle:
We had a reformat but that was to 2 space indentation, [we're discussing changing to 4 spaces now](https://github.com/lukeautry/tsoa/pull/441#issuecomment-524348171).

lukeautry:
The one thing I would point out with 4 spaces is that we're going to get Prettier wrap much earlier than with two spaces. This means you won't be able to fit much at all one a single line, and files will become bigger vertically. This isn't so much a problem as it is a difference.

In terms of "ecosystem standard", two spaces or "soft tab" seems to be the most common: https://hackernoon.com/what-javascript-code-style-is-the-most-popular-5a3f5bec1f6f

The Airbnb JS style guide, which is as close to a gold standard as there is in this ecosystem, seems to prefer 2 spaces.

As bad as a justification as it is, I think I'd prefer sticking with 2 just because it's more common and it's what most people are used to.

HoldYourWaffle:
> The one thing I would point out with 4 spaces is that we're going to get Prettier wrap much earlier than with two spaces. This means you won't be able to fit much at all one a single line, and files will become bigger vertically. This isn't so much a problem as it is a difference.

This is true but we could always increase the horizontal limit to improve this if we wanted to. Not saying we *should*, but it's an option.

<br>

> As bad as a justification as it is, I think I'd prefer sticking with 2 just because it's more common and it's what most people are used to.

I'm honestly quite baffled that so many JS projects use 2 space indentation. It makes the code very hard to read for me, practically impossible if I'm working with bigger methods.
The only explanation I could find for this is that it helped in the days of callback hell, which I totally understand. Fortunately those days are mostly over now that we have async/await, so I simply don't see a reason for this anymore.

You could also see it this way: I haven't encountered *anyone* that has problems with 4 spaces. It might be a little strange at first if you're really not used to it, but you're probably going to be fine.
I'm willing to argue that 2 spaces is objectively harder to read (as long as you keep your nesting levels in check), based on my own experiences and from what I've read online.
In other words: with 2 spaces a lot of people (myself included) will have (a lot of) trouble, with 4 spaces most (if not all) people will be fine.

If there's a massive 2-space fan here that absolutely hates 4 spaces I'd love to hear their point of view. I'm just talking from personal experience and some online reading here, I'd love to learn some more about why people prefer 2 space indentation (other than "it's the standard" because as luke sad, that's a pretty bad justification).

HoldYourWaffle:
I'd still love to have this included. I have already given my reasons above, but my main reason is that having 2 space indentation makes it very hard to contribute for me.

Again if you really want to keep it as 2 spaces, that's perfectly fine too, I just won't be able to contribute as much.

WoH:
2 spaces are default config in literally all major guidelines for type/javascript.
- AirBnb
- standard
- google

dgreene1:
I feel like we've reached a consensus now.

>@HoldYourWaffle: "If @lukeautry decides that this isn't a good idea this can of course be closed."

and he did:

> @lukeautry: "The Airbnb JS style guide, which is as close to a gold standard as there is in this ecosystem, seems to prefer 2 spaces. As bad as a justification as it is, I think I'd prefer sticking with 2 just because it's more common and it's what most people are used to."

and then @WoH furthered the same point in comment https://github.com/lukeautry/tsoa/pull/446#issuecomment-535923352

So although I personally like 4 spaces, we need to keep personal preferences out of it. 

**Decision:** Consensus and standards points to using 2 spaces. So that's what we'll stick to.

(Side Note: there are plugins that allow you to see things how you like them in VSCode (but please look them up and don't ask for a recommendation since I don't know them. Here's an example of a similar but not exact approach: https://marketplace.visualstudio.com/items?itemName=chmln.better-whitespace))

HoldYourWaffle:
If that's your decision then I'll have to accept that, as unjustified as it might feel. I'll look into a whitespace rendering extension.

It might be worth noting that the default router generation templates use 4 spaces.

PR #448, opened 2019-08-24
Fixed a very stupid mistake that should've been caught by tests

HoldYourWaffle:
Whoops...

(Sorry @dgreene1)


WoH:
Is there a testing story for this?

HoldYourWaffle:
Don't use a GUI program that doesn't run commit hooks.
Additionally, actually pay attention to if the tests have ran, instead of assuming they did "because there is a commit hook".

WoH:
Ok, understood. The tests were not executed but would've caught the issue if they were.

HoldYourWaffle:
Yup. It was literally just me not paying enough attention.

PR #450, opened 2019-08-25
README: Add pronunciation and format examples

lukeautry:
- Add the pronunciation of `tsoa`. No one knows how to say it. I picked a way to say it.
- Format examples like we format internal code

### DIFF
--- a/README.MD
+++ b/README.MD
@@ -1,4 +1,5 @@
 # tsoa
+Pronounced so·uh
 
 [![Codeship Status for lukeautry/tsoa](https://codeship.com/projects/cdce38d0-1f6b-0134-258e-1ed679ae6c9d/status?branch=master)](https://codeship.com/projects/160322)
 [![npm version](https://badge.fury.io/js/tsoa.svg)](https://badge.fury.io/js/tsoa)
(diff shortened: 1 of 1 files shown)


README.MD
```diff
@@ -458,7 +456,7 @@ import "./controllers/fooController.ts"
 @DefaultResponse<ErrorResponse>('Unexpected error')
```
WoH: This should probably be updated

lukeautry: What's the latest and greatest way to do this?

WoH: I wish I could tell what the state of tsoa was when this was added.
When I got into it this was confusing though because this decorator [does not exist anymore](https://github.com/lukeautry/tsoa/blob/master/src/decorators/response.ts).

My best guess looking at the method is:

```suggestion
@SuccessResponse()
```

WoH: With 
```ts
@Response('400', 'Bad request')
```

becoming:

```ts
@Response<ErrorResponse>('400', 'Bad request')
```

README.MD
```diff
@@ -1,4 +1,5 @@
 # tsoa
+Pronounced so·uh
```
dgreene1: You’re welcome to make this change to whatever you’d like since you’re the founder, but may I put forward “tee-ess-so-uh”? I always like that it sounds like service oriented architecture (SOA) for TypeScript haha.

HoldYourWaffle:
Interesting to see how English-speaking people pronounce this. My version is completely different of course being a Dutch-native, but that's not really of any value to the rest of the world.

Issue #180, opened 2017-12-15
Is there a way to provide example of request body?


vlad-tkachenko:
I see that Example annotation can be used to provide response example, but is there a way to provide an example of request inside controller?

Something like: 
```
@Post() 
async fn(
  @Body()
  @Example<Model>({
    foo:'bar'
  }) body: Model
) { 
...
}
```

Recodify:
@lukeautry Just wondering if this is currently possible?

dgreene1:
Yes @vlad-tkachenko and @Recodify you can. An illustration of how to use the `@Example` decorator can be seen here: https://github.com/lukeautry/tsoa/blob/41dcf54ec2bfac4a301328f9a602718e4b2e3674/tests/fixtures/controllers/getController.ts#L31

vlad-tkachenko:
That link shows how to use example for response. My initial question was different - on how to provide example for request body.

dgreene1:
Yes the `@Example` decorator places the example in the response section of the success code area in Swagger. I’m actually rather new to tsoa so I apologize for not knowing that. So apparently there isn’t a way to provide an example for the request body. But since most APIs respond with a structure that is the same as request body, this is the first time we have had someone ask for a way to set an example for the body.

Are you asking for someone to write an enhancement to the `@Body` decorator to allow for an example to be added?

If so, we can reopen this issue and mark it as [help wanted].

vlad-tkachenko:
@dgreene1, check my original request, I'm not suggesting changing the `@Body` decorator, but to add support of additional extra decorator `@Example` specifically to provide payload example. 

I will be honest, original request was opened in 2017, I'm no longer using tsoa as my library of choice for REST API development, so currently I don't need any help with that. However other users might need that functionality.

dgreene1:
@vlad-tkachenko which library are you using? I’d love to know so we can be aware of our competition.

WoH:
In tsoa, you can't set an exmple for the specific Request, but for the interface you use there. It looks like this:

```ts
/**
 * @example
 * {
 *   "name": "string",
 *   "array": ['one', 'two', 'three']
 * }
 */
export interface Model {
  name: string;
  array: string[];
}

```

CristianoVotre:
Its strange to put the specific request to the model interface, if I use the same model on 2 or more APIs entry points the example will be the same.... and how to resolve @Path, @Query initial values?
@vlad-tkachenko I understand and I see your problem, check my PR.

[...]

Issue #224, opened 2018-05-24
Return HTML from controller


cpop68:
I have a couple endpoints where I would like to return HTML instead of JSON. Is this possible?

jamestharpe:
```typescript
@Get('html')
async html(
  @Request() request: express.Request
) {
  const response = (<any>request).res as express.Response;
  this.setStatus(200);
  response.contentType('text/html');
  response.send('<html><body><h1>Hello world!</h1></body></html>').end();
  return null; // Found via #44
}
```

karpoff:
I've extended built-in Controller class and now I can do the following:
```
    @Get('/login')
    async login(@Query() cid: string): Promise<void> {
        this.setContent('<html><body><h1>Hello world!</h1></body></html>');
    }
```
`tsoa.controller.ts:`
```
import {Controller} from 'tsoa';

export class TsoaController extends Controller {
    private content: string;

    setContent(c: string): TsoaController {
        this.content = c;
        return this;
    }

    getContent(): string {
        return this.content;
    }
}

```
`routes-express.template:`
```
function promiseHandler(controllerObj: any, promise: any, response: any, next: any) {
        return Promise.resolve(promise)
            .then((data: any) => {
                let statusCode;
                if (isController(controllerObj)) {
                    const headers = controllerObj.getHeaders();
                    Object.keys(headers).forEach((name: string) => {
                        response.set(name, headers[name]);
                    });

                    statusCode = controllerObj.getStatus();

                    if (controllerObj.getContent()) {
                        response.status(statusCode || 200).send(controllerObj.getContent()).end();
                        return;
                    }
                }

                if (data || data === false) { // === false allows boolean result
                    response.status(statusCode || 200).json(data);
                } else {
                    response.status(statusCode || 204).end();
                }
            })
            .catch((error: any) => next(error));
    }```

karpoff:
no, I don't)
maybe next week

HarelM:
Any updates on this? I'm also interested in returning stuff that are not JSON. I might be able to send a PR if I'd know what should be the required approach...

lukeautry:
I think this is a questionable direction. I could probably be convinced otherwise with a more detailed proposal.

`tsoa` is really oriented around JSON APIs. When we start getting into HTML, we're talking about returning content that isn't structured data. In my projects, I use tsoa for the core API, but I often make separate handlers when it doesn't fit well into the tsoa approach.

HarelM:
[...]

PR #459, opened 2019-09-01
Insertion of example in query/path/body

CristianoVotre:
Hi, I develop a form to put examples in the request, its was an old issue:
'Is there a way to provide example of request body? #180'

Now, we can put:
`processarDesconTurbo(
    @Path('idUsuario', 'ABCDEF125') idUsuario: string,
    @Path('idServico', '1') idServico: number,
    @Query('valorCompra', '19.90') valorCompra: number)
`
or in body tag:
`inserirPagamento(
    @Body(`{
        "usuario": {
          "codigo": "ABCDEF125"
        },
          "servico": {
          "codigo": 1
        },
        "desconTurbo": false,
        "parcela": 1,
        "valor": 19.90,
        "valorOnes": 19.90
      }`)
    pagamento: Pagamento)
`
And in the generated swagger will show these values on the fields.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
-  "name": "tsoa",
+  "name": "tsoa-zeroerro",
   "description": "Build swagger-compliant REST APIs using TypeScript and Node",
-  "version": "2.4.11",
+  "version": "2.5.0",
   "main": "./dist/index.js",
   "typings": "./dist/index.d.ts",
   "keywords": [
@@ -23,8 +23,8 @@
   },
   "lint-staged": {
     "*.{js,ts}": [
-        "prettier --write",
-        "git add"
+      "prettier --write",
+      "git add"
     ]
   },
   "scripts": {
@@ -41,7 +41,7 @@
     "postversion": "git push origin master && git push --follow-tags",
     "pretest": "cross-env NODE_ENV=tsoa_test ts-node ./tests/prepare.ts",
     "loading-msg-for-tests": "echo '... starting a test run to ensure quality. Note this takes some time as tests do not run in parallel (by design) ...' ",
-	"test": "node ensureYarn.js && tsc -p ./tests/tsconfig.json --noEmit && yarn loading-msg-for-tests && cross-env NODE_ENV=tsoa_test mocha **/*.spec.ts --exit --require ts-node/register --timeout 5000",
+    "test": "node ensureYarn.js && tsc -p ./tests/tsconfig.json --noEmit && yarn loading-msg-for-tests && cross-env NODE_ENV=tsoa_test mocha **/*.spec.ts --exit --require ts-node/register --timeout 5000",
     "pre-commit": "yarn lint && yarn test",
     "tsc": "tsc"
   },
@@ -59,7 +59,8 @@
     "typescript-formatter": "^7.2.2",
     "validator": "^11.1.0",
     "yamljs": "^0.3.0",
-    "yargs": "^14.0.0"
+    "yargs": "^14.0.0",
+    "yarn": "^1.17.3"
   },
   "devDependencies": {
     "@hapi/hapi": "^18.3.2",
@@ -101,6 +102,9 @@
     "supertest": "^4.0.2",
     "ts-node": "^8.3.0",
     "tslint": "^5.19.0"
+  },
+  "publishConfig": {
+    "registry": "https://devops.zeroerro.com.br/nexus/repository/zeroerro-private/"
   },
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
-  "name": "tsoa-zeroerro",
+  "name": "tsoa",
   "description": "Build swagger-compliant REST APIs using TypeScript and Node",
-  "version": "2.5.0",
+  "version": "2.4.11",
   "main": "./dist/index.js",
   "typings": "./dist/index.d.ts",
   "keywords": [
@@ -23,8 +23,8 @@
   },
   "lint-staged": {
     "*.{js,ts}": [
-      "prettier --write",
-      "git add"
+        "prettier --write",
+        "git add"
     ]
   },
   "scripts": {
@@ -41,7 +41,7 @@
     "postversion": "git push origin master && git push --follow-tags",
     "pretest": "cross-env NODE_ENV=tsoa_test ts-node ./tests/prepare.ts",
     "loading-msg-for-tests": "echo '... starting a test run to ensure quality. Note this takes some time as tests do not run in parallel (by design) ...' ",
-    "test": "node ensureYarn.js && tsc -p ./tests/tsconfig.json --noEmit && yarn loading-msg-for-tests && cross-env NODE_ENV=tsoa_test mocha **/*.spec.ts --exit --require ts-node/register --timeout 5000",
+	"test": "node ensureYarn.js && tsc -p ./tests/tsconfig.json --noEmit && yarn loading-msg-for-tests && cross-env NODE_ENV=tsoa_test mocha **/*.spec.ts --exit --require ts-node/register --timeout 5000",
     "pre-commit": "yarn lint && yarn test",
     "tsc": "tsc"
   },
@@ -59,8 +59,7 @@
     "typescript-formatter": "^7.2.2",
     "validator": "^11.1.0",
     "yamljs": "^0.3.0",
-    "yargs": "^14.0.0",
-    "yarn": "^1.17.3"
+    "yargs": "^14.0.0"
   },
   "devDependencies": {
     "@hapi/hapi": "^18.3.2",
@@ -102,9 +101,6 @@
     "supertest": "^4.0.2",
     "ts-node": "^8.3.0",
     "tslint": "^5.19.0"
-  },
-  "publishConfig": {
-    "registry": "https://devops.zeroerro.com.br/nexus/repository/zeroerro-private/"
   },
   "repository": {
     "type": "git",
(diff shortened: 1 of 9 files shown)


dgreene1:
Can you translate your text above? I apologize, but I can’t make sense if it.

package.json
```diff
@@ -102,6 +103,9 @@
     "ts-node": "^8.3.0",
     "tslint": "^5.19.0"
   },
+  "publishConfig": {
```
dgreene1: What is this and why did you need to add it?

I’m not comfortable with using another registry other than npm.

src/decorators/parameter.ts
```diff
@@ -2,7 +2,7 @@
  * Inject http Body
  *  @param {string} [name] properties name in body object
  */
-export function Body(): Function {
+export function Body(example?: any): Function {
```
dgreene1: Don’t use `any` it should be at least a generic with a requirement to extend an object.

src/decorators/parameter.ts
```diff
@@ -33,7 +33,7 @@ export function Request(): Function {
  *
  * @param {string} [name] The name of the path parameter
  */
-export function Path(name?: string): Function {
+export function Path(name?: string, example?: any): Function {
```
dgreene1: Same as above. Don’t use any.

src/metadataGeneration/tsoa.ts
```diff
@@ -30,12 +30,13 @@ export namespace Tsoa {
...
     parameterName: string;
     description?: string;
-    in: 'query' | 'header' | 'path' | 'formData' | 'body' | 'body-prop' | 'request';
+    in: 'query' | 'header' | 'path' | 'formData' | 'body' | 'body-prop' | 'request' | 'example';
     name: string;
     required?: boolean;
     type: Type;
     default?: any;
     validators: Validators;
+    example?: any;
```
dgreene1: Same as above. No any.

package.json
CristianoVotre: Sorry, this file is wrong..., I will remove it.

src/metadataGeneration/tsoa.ts
CristianoVotre: Ok

src/decorators/parameter.ts
CristianoVotre: Ok

src/decorators/parameter.ts
CristianoVotre: Ok

dgreene1:
@CristianoVotre is this PR still necessary after [someone clarified that there is an alternative approach](https://github.com/lukeautry/tsoa/issues/180#issuecomment-526950724) that already exists?

See here: https://github.com/lukeautry/tsoa/issues/180#issuecomment-526950724

WoH:
I believe this would still be a good addition since you can provide specific examples per path. (And would, therefore, be closer to the `@Request` decorator).
However, I'd like to note that OAPI3 does support multiple examples and I think, ideally, all the related decorators should accept multiple examples eventually (doesn't necessarily have to happen in this PR).

src/decorators/parameter.ts
```diff
@@ -33,7 +33,7 @@ export function Request(): Function {
  *
  * @param {string} [name] The name of the path parameter
  */
-export function Path(name?: string): Function {
+export function Path(name?: string, example?: string): Function {
```
dgreene1: `string` is not accurate either since path variables can be numbers, date, bool, etc (I believe)

src/decorators/parameter.ts
```diff
@@ -2,7 +2,7 @@
  * Inject http Body
  *  @param {string} [name] properties name in body object
  */
-export function Body(): Function {
+export function Body(example?: string): Function {
```
dgreene1: For all of these decorators, we need to make room for multiple configurations that could come down the line. So we need something extendable. That means an object. So instead of taking on primitive, instead have it be:

```ts

export interface BodyDecoratorConfig<T extends object> {
  swaggerOverride?: {
    description?: string,
    example?: T
  }
}

@Body(config?: BodyDecoratorConfig)
```

dgreene1: What are your thoughts @WoH / @LukeAutry / @endor on the notion of allowing for people to override any of the Swagger properties? I ask this because this PR only allows the override of the example and I describe [here](https://github.com/lukeautry/tsoa/pull/459/files#r319777012) how we need to be able future proof by structuring the function to allow an object, for extensibility purposes: https://github.com/lukeautry/tsoa/pull/459/files#r319777012

WoH: Good points. Easiest things first: Yes, every decorator should have a well-specified config interface. I am not against the idea of overriding specific high-level things on a per-request basis, but I'd have to take a look how far I'd go. For now I think, everything annotating the entire interface is reasonable (description and example). I've had cases where I'd have to duplicate an interface to adjust the example I'm giving (one that's valid in the context of the request).

WoH: Now the harder thing: The OpenAPI spec says `example` is of type `any` which would be `unknown` in typescript.
However, it explicitly mentions: 

>  The example SHOULD match the specified schema and encoding properties if present. The example field is mutually exclusive of the examples field. Furthermore, if referencing a schema which contains an example, the example value SHALL override the example provided by the schema. To represent examples of media types that cannot naturally be represented in JSON or YAML, a string value can contain the example with escaping where necessary.

In order to get this checked, the config should allow a type argument so you are able to get type checking, with the fallback to `unknown` (Assuming we can't get typescript to infer the type of a decorated interface).

If we specify an example, we allow to specify all the properties of an example object though: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#example-object

I'll have a look at this tomorrow, but these are the ideas at the top of my head.

CristianoVotre: Well @WoH , I think its a good idea, I put the any value because that, but I can pass all the properties of object like in specification link and I can implement the structure like @dgreene1  suggested... I really need this...

WoH: > I put the any value because that, but I can pass all the properties of object like in specification link

You can still accept everything if you ask for `unknown`, but TS will not allow you to do dangerous unchecked operations on the value.

However, before it makes sense for @dgreene1 to say yes here, it needs to generate correct Swagger 2 **and** OpenAPI3 specs. Also, Dan needs to think about the impact of this PR, regardless of the urgent requirement, he can't allow this PR to prevent future extensions of the features.

dgreene1: > To represent examples of media types that cannot naturally be represented in JSON or YAML, a string value can contain the example with escaping where necessary.

This isn’t a problem for us because tsoa has an [established stance of only handling JSON](https://github.com/lukeautry/tsoa/issues/224#issuecomment-524644680). So I've decided that we should start with strict typing and we can relax it later if someone requests that we make a change/enhancement. Let’s leave the type of `example` as unknown. However, we can't allow tsoa users to make silly mistakes.

So we need to run `swaggerOverrides.example` through `ValidationService.ValidateParam()` during the generation to ensure the developer provided the correct type.

Why am I asking us to run validation instead of this idea of @WoH’s? (keep reading)

> In order to get this checked, the config should allow a type argument so you are able to get type checking, with the fallback to unknown (Assuming we can't get typescript to infer the type of a decorated interface).

So I would prefer that the TypeScript compiler inform the user that the example doesn’t match, but there is a [known issue with inference of parameter decorators](https://github.com/Microsoft/TypeScript/issues/30102) as illustrated by this SO question: https://stackoverflow.com/questions/55144265/how-to-limit-the-type-of-parameter-that-the-given-parameter-decorator-may-be-app).

Reminder @CristianoVotre, I still want the input to be a config object like this:
```ts
interface PathDecoratorConfig {
  name?: string,
  swaggerOverride: {
    example: unknown
  }
}

const parameterDecorator = (decoratorConfig?: string | PathDecoratorConfig) ...
```

WoH: You can still check at compile time, but I think this would require a lot of work. If we'd want to have TS do the work, which is where I was looking at, I'd like to see something like `@Body<T>(config: Config<T>)` where `interface Config<T> = { description?: string, example: T }`. I disagree with the naming, since this is not overriding anything. These examples are explicit part of the OAPI3 spec.

dgreene1: > You can still check at compile time, but I think this would require a lot of work. If we'd want to have TS do the work, which is where I was looking at, I'd like to see something like @Body<T>

I wish we could, but the problem is that `T` can’t be inferred for decorators of parameters. So that’s why [the StackOverflow issue](https://stackoverflow.com/questions/55144265/how-to-limit-the-type-of-parameter-that-the-given-parameter-decorator-may-be-app) presents this possible problem:

```ts
function postSomething( @Body<Thing1>() aParemeterWhosTypeIsNotInferred: Thing2 ): Thing2
```

^ that’s a really problem that the T of Body (I.e `Thing1`) can be different than the variable’s type (I.e. `Thing2`) that’s blocked by https://github.com/Microsoft/TypeScript/issues/30102

WoH: > I wish we could, but the problem is that T can’t be inferred for decorators of parameters. 

It can, because we are in the fortunate situation to be able to parse the AST. However, since we both agreed we should not do it, there's no reason to discuss further imo.

> that’s a really unfortunate problem that the T of Body (I.e Thing1) can be different than the variable’s type (I.e. Thing2).

Yes. I'd say that's better than `unknown`, but I'm happy either way.

> No, swagger 2 supports example just not examples. 

This is incorrect in this context. This PR addresses query, path and body example(s). None of this is supported as is in Swagger 2 and therefore which the article you linked confirms by not mentioning any of these.

A[ Swagger 2] API specification can include examples for:
- response MIME types,
- schemas (data models),
- individual properties in schemas.

dgreene1: Thank you for clarifying @WoH. What version of the spec are you using @CristianoVotre?

PR #461, opened 2019-09-04
Propagate type arguments

### LINKED ISSUE #27: Support generic type parameter and return type

### LINKED ISSUE #391: Swagger not showing intersection model properly
I have the following model:

```
export interface IIQCommentRecord {
  id: string;
  comment: string;
}

export interface DomainCommand{
  aggregateId: string;
}

export type AddCommentCommand extends DomainCommand & {[key: string] : IIQCommentRecord};
```
Where key can be any key since we are working with a quite dynamic interface.

Build works great but the model shown in Swagger UI is the following:
image

Is there any way to show this intersecction properly? I would expect both `aggregateId` and the dictionary to show on the model.

WoH:
### All Submissions:

**Closing issues**

- closes #391
- closes #27 

Please comment if handling Models with multiple type parameters is not something you want to support and I'll `throw` on that :)

This PR assumes the scope of the type arguments has been checked by typescript (which is reasonable I think).

Edit: I don't like the implementation, but it's a lot less bad :/

**Test plan**

I'm sure there are millions of possible scenarios we want to spec here, I have no clue what to cover yet, I intend to keep this a draft until this section is less depressing to read.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,8 +23,8 @@
   },
   "lint-staged": {
     "*.{js,ts}": [
-        "prettier --write",
-        "git add"
+      "prettier --write",
+      "git add"
     ]
   },
   "scripts": {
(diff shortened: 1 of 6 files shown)


dgreene1:
Looks good to me @WoH. I think the only thing that's confusing is that I can't find the line where `this.context` is set for the first time.

As far as tests cases go, I'm fine if we just copy the examples from the two issues you referenced.

ryankeener:
@WoH this fixes the issue I was seeing, thanks!

@dgreene1 it's getting initialized in the constructor, no?

dgreene1:
I see that it's initialized in the constructor but I was wondering if the consuming module sends in the context. I don’t think it does. I think it's the TypeResolver itself when it constructs another TypeResolver.

PR #468, opened 2019-09-10
Close #460 empty string enum

### LINKED ISSUE #460: Empty string in Enum got converted to '0'
When using empty string `''` in string Enum used a input parameter, it got converted as `'0'`.

So imagine an enum like:

```ts
export enum MyThings {
    Undefined = '',
    Something = 'Something',
    Otherwise = 'Another Something',
```

## Sorting

* **I'm submitting a ...**
  - [X] bug report

* I confirm that I

## Expected Behavior

I would expect that the empty string is returned in the swagger doc.

```yaml
    "MyThings ": {
        "enum": [
            "",
            "Something",
            "Another Something"
        ],
        "type": "string"
    },
```

And the '' value should be accepted during data validation.

## Current Behavior
But I got that in swagger config and for validation:

```yaml
    "MyThings ": {
        "enum": [
            "0",
            "Something",
            "Another Something"
        ],
        "type": "string"
    },
```

Well it's 0 because it's in first position, but moving it elsewhere, it will take the index of the position:

```yaml
    "MyThings ": {
        "enum": [
            "Something",
            "Another Something",
            "2"
        ],
        "type": "string"
    },
```

Trying to pass empty string as the value to end point, I got the error "Should be one of [0, Something, Another Something]"

## Workaround
You can change the empty string for something else, like a single space, dash or the word 'Undefined'. But this break Boolean conversion and seems counter-intuitive.

## Context (Environment)
Version of the library: 2.3.8 & 2.4.11
Version of NodeJS: 10.14.1

## Breaking change?
This change the current behaviour, so it's a breaking change in principle. But this seems unlikely that anyone have used empty string in enum as it is.

fhewitt:
### All Submissions:

**Closing issues**

This closes #460 by removing the confusion between falsy and incorrect value. On validation we check that the value in enum is found instead of thruty, and we construct the accept list of values correctly.

Allowing empty enum is close to optional enum parameter, but it enforce the user of the API to knowingly pass the empty value instead of simply ignoring it. Moreover, the lack of support is a bug.

Don't see one, but I have very little knowledge of the code base. At least, this didn't break any unit test.

**Test plan**

None added. Probably should as the code base seems to confuse falsy/invalid, but don't know how.

1. `Adding VALUE_0 = '' as any` to testModel.ts > EnumStringValue
2. Then? I though about checking propertySchema.enum in definition.spec.ts line 263, but the value is undefined.

### DIFF
--- a/src/metadataGeneration/typeResolver.ts
+++ b/src/metadataGeneration/typeResolver.ts
@@ -283,7 +283,8 @@ export class TypeResolver {
 
     if (extractEnum) {
       const enums = enumDeclaration.members.map((member: any, index) => {
-        return getEnumValue(member) || String(index);
+        const enumValue = getEnumValue(member);
+        return !!enumValue || enumValue === '' ? enumValue : String(index);
       });
       return {
         dataType: 'refEnum',
--- a/src/routeGeneration/templateHelpers.ts
+++ b/src/routeGeneration/templateHelpers.ts
@@ -237,7 +237,7 @@ export class ValidationService {
     const enumValue = members.find(member => {
       return member === String(value);
     });
-    if (!enumValue) {
+    if (enumValue === undefined) {
       fieldErrors[parent + name] = {
         message: `should be one of the following; ['${members.join(`', '`)}']`,
         value,
--- a/tests/fixtures/testModel.ts
+++ b/tests/fixtures/testModel.ts
@@ -120,6 +120,7 @@ export enum EnumIndexValue {
  * EnumNumberValue.
  */
 export enum EnumNumberValue {
+  VALUE_0 = 0,
   VALUE_1 = 2,
   VALUE_2 = 5,
 }
@@ -128,6 +129,7 @@ export enum EnumNumberValue {
  * EnumStringValue.
  */
 export enum EnumStringValue {
+  EMPTY = '' as any,
   VALUE_1 = 'VALUE_1' as any,
   VALUE_2 = 'VALUE_2' as any,
 }
--- a/tests/unit/swagger/definitionsGeneration/definitions.spec.ts
+++ b/tests/unit/swagger/definitionsGeneration/definitions.spec.ts
@@ -249,6 +249,10 @@ describe('Definition generation', () => {
             expect(propertySchema.$ref).to.eq('#/definitions/EnumNumberValue', `for property ${propertyName}.$ref`);
             expect(propertySchema['x-nullable']).to.eq(true, `for property ${propertyName}[x-nullable]`);
             expect(propertySchema).to.not.haveOwnProperty('additionalProperties', `for property ${propertyName}`);
+
+            const validatedDefinition = getValidatedDefinition('EnumNumberValue', currentSpec);
+            const expectedEnumValues = ['0', '2', '5'];
+            expect(validatedDefinition.enum).to.eql(expectedEnumValues, `for property ${propertyName}[enum]`);
           },
           enumNumberArray: (propertyName, propertySchema) => {
             expect(propertySchema.type).to.eq('array', `for property ${propertyName}.type`);
@@ -265,6 +269,10 @@ describe('Definition generation', () => {
             expect(propertySchema).to.not.haveOwnProperty('additionalProperties', `for property ${propertyName}`);
             expect(propertySchema.$ref).to.eq('#/definitions/EnumStringValue', `for property ${propertyName}.$ref`);
             expect(propertySchema['x-nullable']).to.eq(true, `for property ${propertyName}[x-nullable]`);
+
+            const validatedDefinition = getValidatedDefinition('EnumStringValue', currentSpec);
+            const expectedEnumValues = ['', 'VALUE_1', 'VALUE_2'];
+            expect(validatedDefinition.enum).to.eql(expectedEnumValues, `for property ${propertyName}[enum]`);
           },
           enumStringArray: (propertyName, propertySchema) => {
             expect(propertySchema.type).to.eq('array', `for property ${propertyName}.type`);
--- a/tests/unit/swagger/schemaDetails3.spec.ts
+++ b/tests/unit/swagger/schemaDetails3.spec.ts
@@ -32,6 +32,20 @@ describe('Definition generation for OpenAPI 3.0.0', () => {
     specName: 'specWithNoImplicitExtras',
   };
 
+  const getComponentSchema = (name: string, chosenSpec: SpecAndName) => {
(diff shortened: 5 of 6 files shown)


src/routeGeneration/templateHelpers.ts
```diff
@@ -237,7 +237,7 @@ export class ValidationService {
     const enumValue = members.find(member => {
       return member === String(value);
     });
-    if (!enumValue) {
+    if (enumValue === undefined) {
```
dgreene1: shouldn't we also be checking for `null` too?

src/metadataGeneration/typeResolver.ts
```diff
@@ -298,7 +298,8 @@ export class TypeResolver {
 
     if (extractEnum) {
       const enums = enumDeclaration.members.map((member: any, index) => {
-        return getEnumValue(member) || String(index);
```
dgreene1: @fhewitt here's a step by step instruction on how to write a unit test for this:

1) add an enum member to https://github.com/lukeautry/tsoa/blob/13d856e68fd1349c3db98ff1160645f1ae8c27e6/tests/fixtures/testModel.ts#L131

So when you're done, it should look like:
```ts
export enum EnumStringValue {
  EMPTY = '',
  VALUE_1 = 'VALUE_1' as any,
  VALUE_2 = 'VALUE_2' as any,
}
```

2) Do the same for the number enums so we know that this PR doesn't create a regression bug. So add a 0 case for EnumNumberValue here: https://github.com/lukeautry/tsoa/blob/13d856e68fd1349c3db98ff1160645f1ae8c27e6/tests/fixtures/testModel.ts#L123

```ts
export enum EnumNumberValue {
  VALUE_0 = 0,
  VALUE_1 = 2,
  VALUE_2 = 5,
}
```

3) Update the unit tests to check the actual data that is produced from the schema. I would recommend a chai assertion like:
```ts
const expectedEnumValues = ['', 'VALUE_1', 'VALUE_2'];
expect(propertySchema.enum).to.eq(true, `for property ${propertyName}[enum]`);
```

3a) update the string enum tests: 
* for OpenAPI 3: https://github.com/lukeautry/tsoa/blob/e482d3c9325c4c11a376ca096628c388c64dd8ca/tests/unit/swagger/schemaDetails3.spec.ts#L303
* for Swagger 2: https://github.com/lukeautry/tsoa/blob/13d856e68fd1349c3db98ff1160645f1ae8c27e6/tests/unit/swagger/definitionsGeneration/definitions.spec.ts#L248

3b) update the nnumber enum tests:
* for OpenAPI 3: https://github.com/lukeautry/tsoa/blob/13d856e68fd1349c3db98ff1160645f1ae8c27e6/tests/unit/swagger/schemaDetails3.spec.ts#L288
* for Swagger 2: https://github.com/lukeautry/tsoa/blob/13d856e68fd1349c3db98ff1160645f1ae8c27e6/tests/unit/swagger/definitionsGeneration/definitions.spec.ts#L264

src/routeGeneration/templateHelpers.ts
fhewitt: I don't think so. enumValue is the result of a .find() it's either undefined, either a members, and I thing I should accept any value if it's a member.

src/metadataGeneration/typeResolver.ts
fhewitt: Ok, so I were close, but it fail:

```expect(propertySchema.enum).to.eq(expectedEnumValues, `for property ${propertyName}[enum]`);```

give

```
Definition generation for OpenAPI 3.0.0
       for specDefault
         should set additionalProperties to false if noImplicitAdditionalProperties is set to "throw-on-extras" (when there are no dictionary or any types)
           should produce a valid schema for the enumStringArray property on TestModel for the specDefault:
     **AssertionError: for property enumStringArray[enum]: expected undefined to equal [ '', 'VALUE_1', 'VALUE_2' ]**
      at Object.enumStringArray (tests\unit\swagger\schemaDetails3.spec.ts:315:44)
      at Context.<anonymous> (tests\unit\swagger\schemaDetails3.spec.ts:498:49)```

Notice that the added test fail also on the master branch.

The object receive:
```
{ items: { '$ref': '#/components/schemas/EnumStringValue' },
  type: 'array',
  description: undefined,
  format: undefined,
  default: undefined,
  nullable: true
}
```

Template helper use the `property.enums` value for the function validateEnum. Not sure why this isn't generated for the test...

WoH: Enums create a ref type.
@fhewitt you can use `getValidatedDefinition('EnumStringValue', currentSpec);` to get the definition and proceed to do the checks @dgreene1 mentioned on the definition

fhewitt: Mmmm, this method didn't exist in schemaDetails3.specs.ts.

I go with the hypothesis that the validation check didn't depends on the schema version, and therefore that schemaDetails3.spec.ts doesn't require any change. If I'm right, the latest commit should cover the unit test requirement.

Thank you both for you explanation and instructions.

dgreene1: You're still missing the tests for OpenAPI 3. It's unfortunate that there isn't a `getValidatedDefinition` in the OpenAPI 3 test file. But you can replicate it by checking the structure of `specDefault.spec.components.schemas.EnumStringValue`

Thank you for the additional tests and for submitting this PR. I'd really like to see coverage for OpenAPI 3 before we merge this PR since most people will probably be using OpenAPI 3 (since Swagger 2 doesn't support the majority of features).

fhewitt: At the same time, a *lot* of unit test over the validation are present in definitions.spec.ts but missing in OpenAPI 3. I strongly suspect that it's because the unit tests over validation are good for both cases / all OpenAPI 3 versions, and therefore not required to be duplicated. I can do it (but not today, I fear) if really required.

dgreene1: I don’t believe that the missing tests in the OpenAPI test file are intentional. I believe they are missing because OpenAPI 3 was only added to tsoa about 3 months ago. So the test coverage was minimal because it was new. You can even see that in the PR for that file where the original author was like “I hope this works?”

So yes it would be really appreciated if you add the tests. Thank you for your patients. I know that it seems unnecessary but it will really help us maintainers and contributes. :)

fhewitt: Sprint delivered, back here!

You seems right, I've remade an equivalent of `getValidatedDefinition` and it required enough changes that it's seems definitely to require distinct testing.

Let me know if you see something else.

PR #472, opened 2019-09-15
Resolve Generic type inside heritage clauses

### LINKED ISSUE #467: Generic type inside extends don't resolve
Following the addition of generic type support (awesome addition! 👍) I've made some tests with it with the hope of simplifying my project, and one of my scenario is failing when I use extends over a generic class where the type used in the controller, is not generic itself.

## Sorting

* **I'm submitting a ...**
  - [X] bug report
  - [X] feature request
  - [ ] support request

* I confirm that I

## Expected Behavior
Theses lines should works.

```ts
export interface ThingContainerWithTitle extends GenericContainer<TheThing> {
    title: string;
}

export interface GenericContainer<T> {
    id: string;
    list: T[];
}
```

While using `ThingContainerWithTitle` as a property/return type of a controller.

## Current Behavior
But I got

> There was a problem resolving type of 'T'.
> There was a problem resolving type of 'GenContainer'.
> There was a problem resolving type of 'ThingContainerWithTitle'.

## Investigation
It work if I create a whole chain of generic types:

```ts
export interface ThingContainerWithTitle<T> extends GenericContainer<TheThing> {
    title: string;
}

export interface GenericContainer<T> {
    id: string;
    list: T[];
}
```

Then using `ThingContainerWithTitle<TheThing>`. But it's just wrong.

## Context (Environment)
Version of the library: 2.5.1
Version of NodeJS: 10.14.1

## Breaking change?
No breaking change: it would add a new case where the generic can be used, but should not affect existing ones.

WoH:
### All Submissions:

**Closing issues**

Closes #467

### DIFF
--- a/tests/fixtures/testModel.ts
+++ b/tests/fixtures/testModel.ts
@@ -448,14 +448,29 @@ export interface GenericRequest<T> {
   value: T;
 }
 
-export interface ThingContainerWithTitle<T> extends GenericContainer<number> {
+interface ThingContainerWithTitle<T> extends GenericContainer<number, number> {
   // T is TestModel[] here
   t: T;
   title: string;
 }
 
-export interface GenericContainer<T> {
+interface GenericContainer<T, TSameNameDifferentValue> {
   id: string;
   // T is number here
   list: T[];
+  dangling: DanglingContext<T>;
+}
+
+/**
+ * This should only be used inside GenericContainer to check it's
+ * type argument T gets propagated while TSameNameDifferentValue does not
+ * and instead, the interface {@link TSameNameDifferentValue} is used.
+ */
+interface DanglingContext<T> {
+  number: T;
+  shouldBeString: TSameNameDifferentValue;
+}
+
+interface TSameNameDifferentValue {
+  str: string;
 }
--- a/tests/fixtures/testModel.ts
+++ b/tests/fixtures/testModel.ts
@@ -440,9 +440,22 @@ export interface GenericModel<T> {
   result: T;
   union?: T | string;
   nested?: GenericRequest<T>;
+  heritageCheck?: ThingContainerWithTitle<T>;
 }
 
 export interface GenericRequest<T> {
   name: string;
   value: T;
+}
+
+export interface ThingContainerWithTitle<T> extends GenericContainer<number> {
+  // T is TestModel[] here
+  t: T;
+  title: string;
+}
+
+export interface GenericContainer<T> {
+  id: string;
+  // T is number here
+  list: T[];
 }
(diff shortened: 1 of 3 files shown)


tests/fixtures/testModel.ts
```diff
@@ -440,9 +440,31 @@ export interface GenericModel<T> {
...
+  // T is number here
+  list: T[];
+  dangling: DanglingContext<T>;
+}
+
+interface DanglingContext<T> {
+  shouldBeString: U;
+}
+
+interface U {
```
dgreene1: @WoH why is `U`  a non-generic interface on line 468 and then it `U` is used before as a generic type on line 457? Are they the same U (and if so, how is that possible?)

tests/fixtures/testModel.ts
```diff
@@ -440,9 +440,31 @@ export interface GenericModel<T> {
   result: T;
   union?: T | string;
   nested?: GenericRequest<T>;
+  heritageCheck?: ThingContainerWithTitle<T>;
```
dgreene1: @WoH I don't necessarily understand what's happening in this PR, but I don't see anything dangerous. So I'm leaning towards "rubber stamping it" (as some people refer to PRs that don't get a lot of scrutiny). But can you clarify what you mean by "heritage?"

tests/fixtures/testModel.ts
WoH: That's part of what I'm trying to prove. They should not be.

tests/fixtures/testModel.ts
WoH: it's the TS naming. I'm scanning the heritage clauses (extends ...), which is a Typescript AST Node.
I'm ready to rename it though, not sure what a better name would be though.

tests/fixtures/testModel.ts
dgreene1: Okay that sounds good. I’m okay with the names, I just wanted t understand. Maybe there should be a comment in the code that explains this.

WoH: My intention was to indicate that by naming the property "shouldBeString". However, I realize that's not sufficient. What would you add there? I feel like any attempt by me would not be as helpful because of language barrier and the fact that I'm very familiar with the code I wrote recently which would potentially lead to bias.

dgreene1: Thank you for clarifying. :)

> What would you add there?

We could rename `U` to be `TSameNameDifferentValue` and then above the interface we could add a JSDoc comment with: "Although this has the same name as the generic type in GenericContainer, it is not the same since this type is concrete and the TSameNameDifferentValue within GenericContainer is a type parameter. We've created this scenario to clarify that tsoa should be able to tell there difference between the two."

WoH: I did not use that exact wording, but I added comments, used `TSameNameDifferentValue` and hope to clarify for everyone what I'm doing.

dgreene1: Cool yea, it was just a suggestion. I like what you wrote. Thank you for following up with that. :)

Btw, do you need published to npm? Or is it okay to sit in master until a new feature is released? (I can deploy this if you want people to start trying it)

PR #475, opened 2019-09-17
Generics: Use %-encoded escaped names to prevent clashes

### LINKED ISSUE #470: Improve Generics Naming
The current naming for supported generic types (aka interfaces) is not very helpful.
Tsoa does some parsing around the model and the generics like `MyModel<Item[]>` becomes `MyModelItemArray` in the schema. However, this use case is very narrowly defined and might lead to more name clashes down the line. Passing literals like `MyModel<'id'> is entirely unsupported only because there is no case to get the name. 

Any effort to fix #338 would also have to deal with this limitation.

## Sorting

* **I'm submitting a ...**
  - [x] bug report
  - [x] feature request
  - [x] support request
  - [x] discussion

* I confirm that I

## Expected Behavior
See possible solution

## Current Behavior
One example:  `MyModel<T | U>` becomes `MyModelobject` and therefore overrides any other `MyModel` declarations with an object or a union.

## Possible Solution
We do a lot of mostly unnecessary work when we could just use the name TypeScript gives the References (`typeReferenceWithGenerics.getText()`) and replace chars not allowed by RFC3986.

I.e. `MyType<T, A | B>` should be something like `MyType_T.A|B_`.

This is something that can be done in a few lines of code instead just blowing up the name finder for the type resolver.

Exemplary steps could be:
- Replace `<>` with `__`
- Replace `,` with `.`
- Replace `[]` with `Array`
- Replace `&` with And
- Strip escaped `"` or `'`

The amount of escapes here is properly specified and therefore, I'd argue, pretty reasonable to implement.

https://tools.ietf.org/html/rfc3986

Another option would be to %-encode (due to easy implementation), but it would hurt readability.

Exemplary types can be en/decoded here: https://www.url-encode-decode.com/

## Context (Environment)
Version of the library: any recent version
Version of NodeJS: irrelevant

## Breaking change?
Depends. Yes if someone depends on the naming tsoa uses for models. However, this does fix a lot of reproducible problems with schema creation. Maybe it makes sense to ship this improvement in 3.0, but it would require a timely major version bump since the current procedure should impede any effort of confidently releasing improvements for generic types.

WoH:
### All Submissions:

* [x] This PR is associated with an existing issue? #470

**Closing issues**

Closes #470 

This changes the names of the generated schemas and therefore may be a breaking change if you rely on the names of the schemas (if you do, please tell me why)

**Test plan**

I updated existing names. By additionally encoding potentially unescaped by illegal characters, it should be impossible to generate illegal names (although they will be harder to decipher)

I believe we could hold off on this for 3.0 and merge it together with type alias changes I'm currently working on.

### DIFF
--- a/src/metadataGeneration/typeResolver.ts
+++ b/src/metadataGeneration/typeResolver.ts
@@ -190,24 +190,9 @@ export class TypeResolver {
       this.typeArgumentsToContext(typeReference, typeReference.typeName, this.context);
     }
 
-    referenceType = this.getReferenceType(typeReference.typeName as ts.EntityName, this.extractEnum, typeReference.typeArguments);
+    referenceType = this.getReferenceType(typeReference);
 
     this.current.AddReferenceType(referenceType);
-
-    // We do a hard assert in the test mode so we can catch bad ref names (https://github.com/lukeautry/tsoa/issues/398).
-    //   The goal is to avoid producing these names before the code is ever merged to master (via extensive test coverage)
-    //   and therefore this validation does not have to run for the users
-    if (process.env.NODE_ENV === 'tsoa_test') {
-      // This regex allows underscore, hyphen, and period since those are valid in SwaggerEditor
-      const symbolsRegex = /[!$%^&*()+|~=`{}\[\]:";'<>?,\/]/;
-      if (symbolsRegex.test(referenceType.refName)) {
-        throw new Error(
-          `Problem with creating refName ${referenceType.refName} since we should not allow symbols in ref names ` +
-            `because it would cause invalid swagger.yaml to be created. This is due to the swagger rule ` +
-            `"ref values must be RFC3986-compliant percent-encoded URIs."`,
-        );
-      }
-    }
 
     return referenceType;
   }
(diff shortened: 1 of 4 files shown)


WoH:
**A note on [contextualized names](https://github.com/lukeautry/tsoa/pull/475/files#diff-04ed568b4a647b0bfd718a5afdb822fbR398)**

The node name TS returns is not always unique and can't be used on itself to determine equality:
```ts
interface Generic<T> {
  t: MyInterface<T>
}

let a: Generic<string>
let b: Generic<number>
```

Will both return the name `MyInterface<T>` when we parse property t on `Generic` and clash when we add MyInterface_T_ to the referenceMap / Schema. That's why I replace these with the name from the context.

WoH:
@dgreene1 would you be comfortable reviewing this?

dgreene1:
@WoH if I recall, this is the PR that will need to be a 3.0 release right?

If so, I’m hoping to close some of the minor features first so they can get out to the majority user. Maybe that’s unnecessary though?

WoH:
I'd personally prefer to merge this in a 3.x release alongside type aliases and some other things concerning type resolution.
Since this is a building block I still would like to make sure this is fine with you or I'm depending on something that's not going to make it which would be awkward.
So review yes, merge maybe? I'm not in a position to decide how tsoa should version itself / when to cut major releases :wink: But I still wanted to point out potential concerns you might have.

WoH:
@dgreene1 I'd like to build on top of this, should I just keep pushing on top of this PR so you can filter commits and review or is should I open more PRs? Does it make sense to target a new branch for the whole thing?

dgreene1:
For now it probably makes sense to keep adding to this PR. Unless the new code is for a completely different feature, then it’s okay to make a new PR that depends on this one.

But I have a note in my ToDo app to get back to this so it doesn’t have to languish out here.

Meanwhile I’m investigating how branches and tags work when you publish to NPM. It seems to be a bit of a headache. So my current plan is to just close and publish all of the small items that are in review and then head to v3 as latest.

dgreene1:
@WoH did you mean to close this? If so, why? (Is it because the other PR includes all of the other changes as it’s base?) Btw, I emailed Luke to make you a contributor (which is the only level available for non organizations).

WoH:
Yes, I closed both PRs and reopened the typealias PR based on version 2.5.5.
Since that new PR includes the nameing I felt like it'd be pretty useless to cherry pick the reworked naming commit onto 2.5.5 and open a second new PR.

dgreene1:
Okay cool. I just wanted to make sure I didn’t mess you up with that big change.

Issue #480, opened 2019-09-23
CLI generate both swagger and routes with one command


scttcper:
## Sorting

* **I'm submitting a ...**
  - [ ] bug report
  - [x] feature request
  - [ ] support request

* I confirm that I
  - [x] used the [search](https://github.com/lukeautry/tsoa/search?type=Issues) to make sure that a similar issue hasn't already been submit

## Expected Behavior
Generate both the swagger docs and the routes file with one command. This would be faster than running both commands individually because the metadata would be generated only once.

We always generate the routes and the swagger docs when starting our server in development mode (which might be overkill?) and the time could be reduced if TSOA generated both at the same time.

PR #481, opened 2019-09-24
feat: expose methods used in cli, so users could build their own cli

scttcper:
This exposes several functions that are used to generate routes and swagger in the cli. This allows these to be easier imported and used, like in a custom cli. Slightly related to #480


dgreene1:
@scttper I feel like it’s generally good practice to not export “private” methods. And I’m a bit concerned that it would become a maintainable headache for maintainers like myself who have to now own complicated custom scripts.

That being said, I would like you to be able to reach your goals.

So can you please share your custom script below so I can see how you use the methods that are newly extended?

scttcper:
@dgreene1 this is mostly related to #482 basically if I can trim down the `generateRoutes` and `generateSwaggerSpec` functions I can cut the restart time of our servers in half during development. I think implementing a custom cli is probably not something the average user will do, and that these do not need to be exposed.

dgreene1:
Okay good idea. Btw, if you’re willing to deep dive into the code, there’s a little known/understood feature of tsoa where it can cache the metadata. So instead of exposing the internals, we could expose a config option like “useCache”. I believe that we recently disabled it altogether due to issues described here: https://github.com/lukeautry/tsoa/issues/420#issuecomment-522265171

But it could be re-enabled if we better understood. If you see this discussion (https://github.com/lukeautry/tsoa/issues/420#issuecomment-522265171) you’ll see that it was an ordering issue.

scttcper:
@dgreene1 thanks! i'll look into that too

PR #482, opened 2019-09-24
feat: generate both swagger and routes with one metadata call

### LINKED ISSUE #480: CLI generate both swagger and routes with one command

scttcper:
### All Submissions:

* [ ] Have you written unit tests?

**Closing issues**

closes #480

This was an experiment I tried to improve the regeneration speed of swagger/routes during development. It reduced the time in our app from 7s to 5s. More time could be reduced by only doing `validateMutualConfigs` once. I'm unsure if this would add value to the project.

**Test plan**

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -4,9 +4,11 @@ import * as ts from 'typescript';
 import * as YAML from 'yamljs';
 import * as yargs from 'yargs';
 import { Config, RoutesConfig, SwaggerConfig } from './config';
+import { MetadataGenerator } from './metadataGeneration/metadataGenerator';
 import { generateRoutes } from './module/generate-routes';
 import { generateSwaggerSpec } from './module/generate-swagger-spec';
 import { fsExists, fsReadFile } from './utils/fs';
+// import { validateMutualConfigs } from './utils/mutualConfigValidation';
 
 const workingDir: string = process.cwd();
 
@@ -165,6 +167,18 @@ yargs
     },
     routeGenerator,
   )
+  .command(
+    'swagger-and-routes',
+    'Generate swagger and routes',
+    {
+      basePath: basePathArgs,
+      configuration: configurationArgs,
+      host: hostArgs,
+      json: jsonArgs,
+      yaml: yarmlArgs,
+    },
+    generateSwaggerAndRoutes,
+  )
   .help('help')
   .alias('help', 'h').argv;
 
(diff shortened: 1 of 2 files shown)


dgreene1:
What do you think @LukeAutry? Was there a reason why tsoa has always been split into two commands?

I know I always generate both and I suspect that the majority of users do too.

So is there some reason that we shouldn’t allow this PR?

scttcper:
I'm open to opinions. Went with `tsoa swagger-and-routes` because that was simple, but it could be a new command that does both by default like `tsoa generate` and takes flags to not generate.

```
tsoa generate --no-swagger
```

lukeautry:
I like this.

In my own projects, I don't use the CLI, I use the (poorly documented) Node.js API and have it tightly integrated with the build and dev setup. `ts-app` is a good example of how this might work.

Personally, I would like more people to consider this approach, or simply writing a Node script that uses the API, but I recognize that the CLI is super easy to wrap your head around. 

I'm fine with swagger-and-routes.

scttcper:
let me know if you think there's any way I could improve this cli method before merging

src/cli.ts
```diff
@@ -214,3 +227,36 @@ async function routeGenerator(args) {
     process.exit(1);
   }
 }
+
+async function generateSwaggerAndRoutes(args) {
+  try {
+    const config = await getConfig(args.configuration);
+    if (args.basePath) {
```
dgreene1: can you please give args a type so it's not implicitly any? (btw, there will be a PR soon that turns on noImplicitAny in tsconfig)

src/cli.ts
```diff
@@ -214,3 +227,36 @@ async function routeGenerator(args) {
     process.exit(1);
   }
 }
+
+async function generateSwaggerAndRoutes(args) {
```
dgreene1: Since `generateSwaggerAndRoutes` might someday become the "critical path" for tsoa, we should probably get it plugged into the integrated tests. So can you please replace a few of the `generateRoutes` inside of `prepare.ts` with `generateSwaggerAndRoutes` so that we will know when/if `generateSwaggerAndRoutes` regresses?

PR #488, opened 2019-09-27
Remove Request.user declarations (#486)

### LINKED ISSUE #486: What is the function of express/hapi/koa.d.ts?
There are currently 3 declaration files that add an optional `user` field to the `Request` interface of the corresponding framework:

https://github.com/lukeautry/tsoa/blob/0d4f17aa651418a1fec735e096cc0d907ff0449f/src/koa.d.ts#L1-L6

https://github.com/lukeautry/tsoa/blob/0d4f17aa651418a1fec735e096cc0d907ff0449f/src/express.d.ts#L1-L6

https://github.com/lukeautry/tsoa/blob/0d4f17aa651418a1fec735e096cc0d907ff0449f/src/hapi.d.ts#L1-L6

What is the function of these declaration files? Do the generated routes use them in some way, and if so, why? As far as I know TSOA doesn't have anything to do with authentication/user management.

HoldYourWaffle:
As discussed in #486. I actually moved them to the test file that uses this property to keep the typescript compiler happy.
I added the tests & docs folder to the .npmignore to avoid conflicts (and because they shouldn't be published as far as I know).
No router template modification was needed because `request` already has an `any`-type there.

### All Submissions:

* [ ] ~~Have you written unit tests?~~ *(not applicable)*
* [ ] ~~Have you written unit tests that cover the negative cases (i.e.: if bad data is submitted, does the library respond properly)?~~ *(not applicable)*

**Closing issues**
Closes #486

Users might be confused about not having a `user` property, I should probably add a note to the readme that mentions this and shows how to do it for those that don't use an external authentication library that already does this for them.
I have no idea how to properly write such a thing though.

### DIFF
--- a/src/hapi.d.ts
+++ b/src/hapi.d.ts
@@ -1,6 +1,0 @@
-import '@hapi/hapi';
-declare module '@hapi/hapi' {
-  interface Request {
-    user?: any;
-  }
-}
--- a/tests/fixtures/controllers/securityController.ts
+++ b/tests/fixtures/controllers/securityController.ts
@@ -1,25 +1,8 @@
-import * as hapi from '@hapi/hapi';
-import * as express from 'express';
-import * as koa from 'koa';
 import { Get, Request, Response, Route, Security } from '../../../src';
 import { ErrorResponseModel, UserResponseModel } from '../../fixtures/testModel';
 
-declare module 'express' {
-  interface Request {
-    user?: any;
-  }
-}
-
-declare module '@hapi/hapi' {
-  interface Request {
-    user?: any;
-  }
-}
-
-declare module 'koa' {
-  interface Request {
-    user?: any;
-  }
+interface RequestWithUser {
+  user?: any;
 }
 
 @Route('SecurityTest')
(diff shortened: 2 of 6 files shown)


tests/fixtures/controllers/securityController.ts
```diff
@@ -4,6 +4,24 @@ import * as koa from 'koa';
 import { Get, Request, Response, Route, Security } from '../../../src';
 import { ErrorResponseModel, UserResponseModel } from '../../fixtures/testModel';
 
+declare module 'express' {
```
dgreene1: If this is the only file that uses it, then do we even need to `declare module` at all?

Suggested change:
```ts
interface RequestWithUser {
  user?: any;
}

//...

public async GetWithApiForHapi(@Request() request: RequestWithUser): Promise<UserResponseModel> {
    return Promise.resolve(request.user);
}
```

src/hapi.d.ts
```diff
@@ -1,6 +0,0 @@
-import '@hapi/hapi';
```
dgreene1: Please delete (via `git rm`) these three `d.ts` files since they no longer have any other content are now blank.

HoldYourWaffle: They're already deleted on my machine and git refuses to `rm` them because it can't find them. Are you sure they still exist on your machine?

tests/fixtures/controllers/securityController.ts
HoldYourWaffle: Good idea. I already had a hunch that this could be done cleaner but this didn't occur to me (in my defense I was probably very tired ;)).

I have now implemented this change.

src/hapi.d.ts
dgreene1: That means you used `rm` and not `git rm` first. So in order to `git rm` you’d have to reset those three files via `git checkout — koa.d.ts`

HoldYourWaffle: I'll see if that works when I'm back at my desk.

In the meantime, are you really really really sure they're not deleted? The PR diff tells me the files have been deleted and they're not in [my fork's tree](https://github.com/HoldYourWaffle/tsoa-iso/tree/remove-user-property/src) either

dgreene1: yes I am sure that the hapi file has not been deleted. If it were deleted, then Github wouldn't show the diff like it currently does. It would show something like this:
image

dgreene1: This is strange. Now it's showing that the file was deleted. Github must have had a UI/UX bug. My apologies.

alfetch:
So now when you have removed the user from the request, how do all of us that do not use an external authentication as passport add this. We are only using tsoa authentication and used the user property before. Now it's removed :/ I have a hard time figuring out how to implement it again. 

Do any of you have a example of how to do this?

WoH:
This is a purely typescript specific question that is not related to tsoa in any way.
You can make use of declaration merging in your code the same way tsoa used to do it.
https://www.typescriptlang.org/docs/handbook/declaration-merging.html

```
declare module 'express' {
  export interface Request {
    myProp?: string;
  }
}
```

alfetch:
Thanks

PR #493, opened 2019-10-02
fixes issue 492 by documenting the correct type of the enum

### LINKED ISSUE #492: Numeric enums are being documented as string enums
The numeric enum PR (https://github.com/lukeautry/tsoa/issues/331) didn't have the correct type showing in Swagger.

## Sorting

* **I'm submitting a ...**
  - [X] bug report
  - [ ] feature request
  - [ ] support request

* I confirm that I

## Expected Behavior

```yaml
paths:
  /items:
    get:
      parameters:
        - in: query
          name: someNumber
          description: a number that I want the user to provide
          schema:
            type: number
            enum: [1, 2]
```

```ts
export enum someNumber {
  one = 1,
  two = 2
}
```

## Current Behavior

The `schema.type` parameter is wrong.
```yaml
paths:
  /items:
    get:
      parameters:
        - in: query
          name: someNumber
          description: a number that I want the user to provide
          schema:
            type: string
            enum: [1, 2]
```

```ts

```

## Possible Solution

Have the code check if the enum only has one type or the other and then use that type.

## Steps to Reproduce

1. Use the enum provided above and see what it outputs

## Context (Environment)
Version of the library: v2.5.4
Version of NodeJS: LTS

## Breaking change?
no

dgreene1:
### All Submissions:

* [X] Have you written unit tests that cover the negative cases (i.e.: if bad data is submitted, does the library respond properly)?

**Closing issues**

closes #492

* [X] Has the issue had a maintainer respond to the issue and clarify that the feature is something that aligns with the [goals](https://github.com/lukeautry/tsoa#goal) and [philosophy](https://github.com/lukeautry/tsoa#philosophy) of the project?

At first I thought I was just going to throw an error for when enums contains string && numbers, but I thought that it it was unwise to throw when that could be a breaking change. So I've left that as a nice little warning for people.

**Test plan**
I added test that cover the negative and positive cases. The most important tests are the one for the schema generation.

### DIFF
--- a/src/swagger/specGenerator.ts
+++ b/src/swagger/specGenerator.ts
@@ -194,6 +194,45 @@ export abstract class SpecGenerator {
     };
   }
 
+  private determineTypesUsedInEnum(anEnum: Array<string | number>) {
+    const typesUsedInEnum = anEnum.reduce((theSet, curr) => {
+      const typeUsed = typeof curr;
+      theSet.add(typeUsed);
+      return theSet;
+    }, new Set<'string' | 'number' | 'bigint' | 'boolean' | 'symbol' | 'undefined' | 'object' | 'function'>());
+
+    return typesUsedInEnum;
+  }
+
+  protected decideEnumType(anEnum: Array<string | number>, nameOfEnum: string): 'string' | 'integer' {
+    const typesUsedInEnum = this.determineTypesUsedInEnum(anEnum);
+
+    const badEnumErrorMessage = () => {
+      const valuesDelimited = Array.from(typesUsedInEnum).join(',');
+      return `Enums can only have string or number values, but enum ${nameOfEnum} had ${valuesDelimited}`;
+    };
+
+    let enumTypeForSwagger: 'string' | 'integer';
+    if (typesUsedInEnum.has('string') && typesUsedInEnum.has('number')) {
+      if (typesUsedInEnum.size !== 2) {
+        throw new Error(badEnumErrorMessage());
+      }
+      // TODO: It would be great to throw here, but without knowing how our users have been using tsoa, we can't make a breaking change until version 3.0
+      // tslint:disable-next-line: no-console
+      console.warn(`Swagger/OpenAPI does not support enums with both strings and numbers but found that for enum ${nameOfEnum}`);
+      const weHaveToPickOneSinceWeCantThrow = 'string';
+      enumTypeForSwagger = weHaveToPickOneSinceWeCantThrow;
+      // End of TODO
+    } else if (typesUsedInEnum.has('string') && typesUsedInEnum.size === 1) {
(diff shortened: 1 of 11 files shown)


dgreene1:
@WoH would you be willing to review this PR?

WoH:
I think this could be done in less code if you check in the type resolver using `Array.prototype.some()`, no?

dgreene1:
Yes I avoided .some because I’d have to do multiple iterations:
* const hasNumbers = theArray.some ...
* const hasStrings = theArray.some ...
* const hasANonStringNonNumberType = theArray.some ...

By creating a Set I only had to iterate once.

src/swagger/specGenerator.ts
```diff
@@ -194,6 +194,45 @@ export abstract class SpecGenerator {
...
+    };
+
+    let enumTypeForSwagger: 'string' | 'integer';
+    if (typesUsedInEnum.has('string') && typesUsedInEnum.has('number')) {
+      if (typesUsedInEnum.size !== 2) {
+        throw new Error(badEnumErrorMessage());
+      }
+      // TODO: It would be great to throw here, but without knowing how our users have been using tsoa, we can't make a breaking change until version 3.0
+      // tslint:disable-next-line: no-console
+      console.warn(`Swagger/OpenAPI does not support enums with both strings and numbers but found that for enum ${nameOfEnum}`);
```
WoH: This should be on the list for 3.0 then :wink:

WoH:
> Just from my initial look, without looking at the details, I would've assumed this could be done in less code if you check in the type resolver using Array.prototype.some().

You're right about using a set, I was confused as to where most of the code would come from, but that's mostly the lock file.

WoH:
However, the better OpenAPI 3 spec would probably be:

```yaml
    EnumMixedValue:
      description: A mixed enum.
      anyOf:
        - type: string
          enum:
          - "0"
          - "1"
        - type: number
          enum:
            - 2
```

I could see an argument to only have this be expressed as a typescript union of 2 enums, which we already support and should produce more or less this result (we'd ref the schemas, otherwise identical).
While the previous definition (https://github.com/lukeautry/tsoa/pull/493#pullrequestreview-296428489) would not error, at least the Swagger Editor did not show the enum values in the UI which leads me to believe enums will be ignored in combination with anyOf.

dgreene1:
Thank you yes, I couldn’t find any evidence that a mixed type enum was supported.

WoH:
I think the way you propose to handle this is the best way. Sorry it took me some time to come to the same conclusion :laughing: 

LGTM

src/swagger/specGenerator.ts
dgreene1: Okay, I added it: https://github.com/lukeautry/tsoa/issues/501

Issue #457, opened 2019-08-31
Clarifying what "refactors for the sake of refactoring" means


HoldYourWaffle:
## Sorting

* **I'm submitting a ...**
  - [ ] bug report
  - [ ] feature request
  - [ ] support request
  - [x] discussion point (@dgreene1 should this be added to the template?)

* I confirm that I
  - [x] used the [search](https://github.com/lukeautry/tsoa/search?type=Issues) to make sure that a similar issue hasn't already been submit

## Current Behavior
This is the section of the contributing guide I want to talk about:

> Note that it is unlikely the project will merge refactors for the sake of refactoring. These types of pull requests have a high cost to maintainers in reviewing and testing with little to no tangible benefit.

When I read this for the first time the first thing that came to my mind was: "When is a refactor a refactor for the sake of refactoring?", and after that: "What about actual *useful* cleanups?".
**I'm not aiming for another philosophical debate about software development here**, I think we've debated that issue enough in #445. However, I think (or at least hope) that we can all agree that completely closing the door for code quality/readability/accessibility improvements is neither necessary nor desirable, and that *significantly* improving code quality has an actual tangible benefit to the project (at least in the long-run).

## Expected Behavior
I hope that we can keep an open-mind about these kind of improvements that are actually useful. Perhaps that was already the intention, in which case this is just a clarification.

## Detailed Description
I think most of mine and @dgreene1's arguments on this matter are already listed in #445.
Some of my notable points were:
- The current code (at least in the areas I checked) can be really confusing or opaque to newcomers
- There's a lot of "low hanging fruit" when it comes to typesafety improvements, which might require some refactoring (I'd argue that this wouldn't be a "refactor for refactoring's sake" even with the current guide)
- Locking into "how things are now because it works" doesn't promote a healty and clean codebase or continuous quality improvements
- This code isn't big enough to *need* this kind of bureaucracy for sanity

Please look at #445 for a more detailed explanation of my reasoning and some additional minor points.

## Possible Solution
I think the contributing guide should be clarify what *isn't* a "refactor for the sake of refactoring". My take would be something like this:

[...]

Issue #452, opened 2019-08-27
Support for the unknown type


dgreene1:
This is a proposal for the `unknown` type.

Before anyone takes on this PR, we need to
* allow the object literal PR and the type alias PR to close
* we need to determine what input and expected outcomes should be for aggregates and unions that contain unknown

**Value that the `unknown` type offers:**

Sometimes an API might allow the consumer to pass arbitrary data down to the service that will then be persisted in a schema-less DB like Mongo or PostGres’s JSONB. So as where the `any` type has risks associate, the `unknown` type more clearly communicates “I wouldn’t recommend trying to unwrap or utilize any subset of this data since it could have any shape or type.”

WoH:
What is the expected representation of the unknown type on its own?
Imagine a controller fn returning `unknown`. What's that intended to be documented as?

WoH:
Basically binary data? From the OpenAPI3 Spec:

```yaml
requestBody:
  content:
    application/octet-stream:
      # any media type is accepted, functionally equivalent to `*/*`
      schema:
        # a binary file of any type
        type: string
        format: binary
```

WoH:
> But maybe we already support returning any which is essentially the same.

We document any like an object with unspecified properties, which is IMHO wrong and maybe should not be repeated.
On the validation side: We just don't do any checks, which is the most reasonable behavior for any.
`unknown` should check for existence at least I think.

HoldYourWaffle:
> I’m creating this thread @WoH / @HoldYourWaffle due to the excellent guidance from the ElasticSearch contribution guidelines.

Do I sense some shade 😉 
Just to be clear, I agree that the guidelines you cited make sense for this project.

> we need to determine what input and expected outcomes should be for aggregates and unions that contain unknown

As an actual input to the discussion: what do you mean by aggregates in this context?

WoH:
> No it’s not a mistake, it’s actually how swagger and OpenAPI 3 document the any type. See here: https://swagger.io/docs/specification/data-models/data-types/#any

I did not know that and it's really valuable information, thank's a lot for pointing it out.
I retract my initial assessment.

> Since an unknown value can also be undefined or null in TypeScript, we can’t validate for existence. So unknown would have all of the same validation qualities (or lack-thereof) as any.

I should have clarified I did not mean JS existence. But since unknown may be `undefined` I'd agree.

HoldYourWaffle:
> I would prefer to stop talking about the philosophical aspects of refactoring.

[...]

PR #485, opened 2019-09-25
Referenced TypeAliases

WoH:
### All Submissions:

* [x] Have you followed the guidelines in our [Contributing](https://github.com/lukeautry/tsoa/tree/master/docs/CONTRIBUTING.md) document?
* [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/lukeautry/tsoa/pulls) for the same update/change?
* [ ] Have you written unit tests?
* [ ] Have you written unit tests that cover the negative cases (i.e.: if bad data is submitted, does the library respond properly)?
* [ ] This PR is associated with an existing issue?

**Closing issues**

Closes #204 

### If this is a new feature submission:

* [ ] Has the issue had a maintainer respond to the issue and clarify that the feature is something that aligns with the [goals](https://github.com/lukeautry/tsoa#goal) and [philosophy](https://github.com/lukeautry/tsoa#philosophy) of the project?

**Potential Problems With The Approach**

**Test plan**

Generated validation models look good to me, I added some props to the validation model to indicate what we can do now.

Ideal test plan would be:
- alias a string with validation
- alias a number with validation
- maybe alias a date with validation
- maybe alias union/intersection
- alias a union of a model and an alias with validation on the alias
- alias nestedObjectLiteral with validation on a property within the nOL
- generic type alias

For each of these:
- Definition: Description
- Definition: Example
- Definition: type is $ref
- Definition: Model is fine
- Validation: Validates
- Validates: invalidates

This is why this will probably take me a long time and I'd be super grateful if someone could contribute parts of that.

Opened a second pull request based on #475 .
To view only this PR without dependencies, use GitHub's filter commits feature. (Similar to [this](https://help.github.com/en/articles/filtering-files-in-a-pull-request), but filter commits (left) instead of files (middle)).

**Todo:**
- [x] Test plan
- [ ] Merge #475
- [x] Refactor legacy reference checking based on properties to check on kind instead.
- [ ] Test implementations

PR #495, opened 2019-10-03
Internal refactor: "tagged union" allows us to remove almost all type assertions in TypeResolver

dgreene1:
This implements the [tagged union](https://github.com/FrontlineEducation/func-alexa-1/pull/210/files) solution recommended by @lukeautry as a solution for reducing the number of type assertions (i.e the `as` operator).

This should resolve the refactoring that was presented in https://github.com/lukeautry/tsoa/issues/457 and that was designed to make tickets such as https://github.com/lukeautry/tsoa/issues/452 easier to implement.

As recommended by Luke and presented by @HoldYourWaffle, there are no new features or bug fixes. It's just a refactor and therefore I tried to make as minimal changes as possible to the test code so that it was clear that no functionality should have changed with the refactor.


dgreene1:
@WoH I can not figure out why this refactor broke two of the integration tests. One of them is the `"should call out any additionalProperties on Unions"` test that you wrote, so would you be willing to give a second set of eyes to help me figure out what I did wrong?

Also, since this code touches a lot of the v3 work you're doing, please review the changes and let me know if/when you think that it should get into master (i.e. as part of v2 or as part of v3)

WoH:
I'm doing something similar with the refTypes in the TypeAlias PR:
https://github.com/lukeautry/tsoa/pull/485/commits/a91dae8509ebed5cdae62acb14e824d1566d4db8

So generally, I like the approach. But I have strong feelings concerning the (new) names.

README.MD
```diff
@@ -223,6 +223,16 @@ export interface MyModel {
 ```
 
 ### Overriding route template
+
```
WoH: Is this related to the goal of the PR?

src/metadataGeneration/initializer-value.ts
```diff
@@ -1,7 +1,7 @@
 import * as ts from 'typescript';
 import { Tsoa } from './tsoa';
 
-export const getInitializerValue = (initializer?: ts.Expression, type?: Tsoa.Type) => {
```
WoH: I would strongly prefer to keep the name as it is and only changing the underlying type declaration.

src/metadataGeneration/metadataGenerator.ts
```diff
@@ -87,4 +87,4 @@ export class MetadataGenerator {
       .filter(generator => generator.IsValid())
       .map(generator => generator.Generate());
   }
-}
```
WoH: Formatting

src/metadataGeneration/typeResolver.ts
```diff
@@ -127,7 +130,7 @@ export class TypeResolver {
         additionalType = new TypeResolver(indexSignatureDeclaration.type as ts.TypeNode, this.current, this.parentNode, this.extractEnum, this.context).resolve();
       }
 
-      const objLiteral: Tsoa.ObjectLiteralType = {
+      const objLiteral: Tsoa.NestedObjectMetaType = {
```
WoH:
```suggestion
      const objLiteral: Tsoa.NestedObjectLiteralType = {
```

Please don't change the naming without reason. I agreed to add nested so new contributors can read this more easily. However, TS calls these ObjectLiterals, so everyone familiar with the TS terminology should be respected aswell.

src/metadataGeneration/typeResolver.ts
```diff
@@ -185,40 +191,56 @@ export class TypeResolver {
       return literalType;
     }
 
-    let referenceType: Tsoa.ReferenceType;
     if (typeReference.typeArguments && typeReference.typeArguments.length > 0) {
       this.typeArgumentsToContext(typeReference, typeReference.typeName, this.context);
     }
 
-    referenceType = this.getReferenceType(typeReference.typeName as ts.EntityName, this.extractEnum, typeReference.typeArguments);
```
WoH: I don't really see why this has any benefits over the current implementation and it does not warrant all this new logic to avoid type casts in my opinion.
Honestly, I am already struggling with the scope of the type aliases. This feels like an additional punishment.

src/routeGeneration/routeGenerator.ts
```diff
@@ -238,3 +263,35 @@ export class RouteGenerator {
     return schema;
   }
 }
+
+/**
+ * This will help us do exhaustive matching against only reference types
+ */
+export function isRefType(metaType: Tsoa.MetaType): metaType is Tsoa.ReferenceType {
```
WoH: should be util.
We may want to add more of these.

tests/fixtures/custom/custom-tsoa-template.ts.hbs
```diff
@@ -18,9 +18,11 @@ const models: TsoaRoute.Models = {
     {{#each models}}
     "{{@key}}": {
         {{#if enums}}
```
WoH: Can we use the new check here?

tests/unit/templating/templateHelpers.spec.ts
```diff
@@ -4,12 +4,20 @@ import { FieldErrors, TsoaRoute, ValidationService } from '../../../src';
 import { SwaggerConfigRelatedToRoutes } from '../../../src/routeGeneration/routeGenerator';
 import { TypeAliasModel1, TypeAliasModel2 } from '../../fixtures/testModel';
 
+/**
+ * This is a convenience type so you can check .properties on the items in the Record without having TypeScript throw a compiler error. That's because this Record can't have enums in it. If you want that, then just use the base interface
+ */
+interface ModelsButOnlyRefObjects extends TsoaRoute.Models {
```
WoH: Should be defined next to Models as RefObjectModels.

src/routeGeneration/templates/express.hbs
```diff
@@ -24,16 +24,20 @@ const models: TsoaRoute.Models = {
     {{#each models}}
     "{{@key}}": {
         {{#if enums}}
```
WoH: The way you are setting them up would mean I should change the TAs to look similar. ("dataType": "refType")

src/metadataGeneration/metadataGenerator.ts
WoH: Seems like you avoided all checks to be able to push this.
You can still get the formatting to not be off by running `yarn run prettier **/*.ts --write`

src/metadataGeneration/typeResolver.ts
dgreene1: Good point

src/metadataGeneration/typeResolver.ts
dgreene1: The only advantage it provides is that refObjects definitely don’t have an `enums` property and refEnums definitely don’t have a `properties` property.

In the current code, the interface has all properties as nullable.

But yes, I hear you. The ramifications of the tagged union approach were much more than I expected. So if you don’t see an advantage in the approach, then we can close this PR.

README.MD
dgreene1: It’s only related because I had to make changes that would break everyone who has a custom template. Those people using custom templates should probably be told that they’re at risk. If not, then we have to consider every data change to the templates as a breaking change. I don’t want to do that many major versions because people will miss out on cool features then.

src/routeGeneration/templates/express.hbs
```diff
@@ -24,16 +24,20 @@ const models: TsoaRoute.Models = {
...
         {{#if properties}}
+        "dataType": "refObject",
         "properties": {
             {{#each properties}}
             "{{@key}}": {{{json this}}},
             {{/each}}
         },
         {{/if}}
-        "additionalProperties": {{additionalPropsHelper additionalProperties}},
+        {{#if additionalProperties}}
```
WoH: validateModel grabs this [here](https://github.com/lukeautry/tsoa/blob/master/src/routeGeneration/templateHelpers.ts#L632).

Afterwards it checks if it `=== false`.
We should make it either check falsy of put false in the template by default. This causes the bug in validateModel and therefore breaks validate Union which calls validateModel.

src/metadataGeneration/initializer-value.ts
dgreene1: Oh yea, I changed the name in the first place so that I could see (via compiler error messages) where I needed to update the code. Essentially I was using the compiler as my pair-programming buddy.

I fully intended to change the name back.

But honestly, I have no intention of merging a PR that makes development on the project harder. After all, refactor is supposed to:

* improve speed of development
  * it didn’t in this case because the new strictness caused some weirdness
* improve code quality
  * well, two tests are broken that I can’t fix so I can’t say that quality was improved

And most importantly, there’s the subjective quality of the work. If you would rather not work on top of this PR, then we shouldn’t merge it.

Either way, the name `Type` stays the same.

src/metadataGeneration/tsoa.ts
```diff
@@ -95,48 +95,121 @@ export namespace Tsoa {
...
+
+  export type ReferenceType = RefEnumMetaType | RefObjectMetaType;
+
+  export interface ReferenceTypeMinimal extends MetaTypeMinimal {
     description?: string;
     dataType: RefTypeLiteral;
     refName: string;
-    properties?: Property[];
-    additionalProperties?: Type;
-    enums?: string[] | number[];
```
dgreene1: @WoH these three lines above are probably the only part of this PR that can clarify the possible value that this PR provides. Right now it looks like a `ReferenceType` can have all of these properties or none. Which means that the compiler can’t help us enforce that we give the right properties to the right type.

dgreene1:
Long story short, we don’t have to merge this PR at all. I wanted to try the challenge of doing the refactor that HoldYourWaffle proposed. But if this change is perceived as being “more harm than good” then it shouldn’t be merged.

src/metadataGeneration/typeResolver.ts
WoH: @dgreene1 I think I didn't clearly state what I meant by "this".

I am specifically referring to the change of logic below the line I commented on.
I approve of the type changes (minus the naming). I would've liked to see a narrower scope basically.

After checking it out locally now, I think I am more ok with the logic here, maybe I am too biased because of all the rebasing I'll have to do, keep that in mind.

src/metadataGeneration/tsoa.ts
WoH: I certainly [agree](https://github.com/lukeautry/tsoa/pull/485/commits/a91dae8509ebed5cdae62acb14e824d1566d4db8) that this provides value ;)

WoH:
> But if this change is perceived as being “more harm than good” then it shouldn’t be merged.

Let me be clear. It's very good but it certainly creates more work, especially the change of control flow in the type resolver.

If you can address:
- the failing tests
- formatting
- type naming

I think it's worth the effort. Which was my position with the PR that inspired this one aswell.

src/metadataGeneration/tsoa.ts
dgreene1: Okay yea, so 50% (or more) of the value here has already been achieved in your other PR. That’s more reason to close this.

WoH: I disagree. The value does not only lie in fewer bugs/better checking of the existing code but a better DX for contributors. Let's not underestimate that

WoH:
I already got comfortable with rebasing and had a plan in mind ;)

If you can address [these issues](https://github.com/lukeautry/tsoa/pull/495#issuecomment-537934417) I'm leaning very much towards B.

src/routeGeneration/templates/express.hbs
dgreene1: I am checking falsy. It's just unclear because I was being fancy/tricky with it because I included that check inside of the `isDefaultForAdditionalPropertiesAllowed` function. What I was trying to do was to use "code as comments" to clarify that an undefined value for `additionalProperties` should be construed as "yes, additional properties are allowed."

So if I make your suggested change, then that  would kind of subvert the goal of that method.

I opted instead to make sure that the `"additionalProperties"` property is always there in the template, but that it's value might be literally `undefined`. That makes this fix a handlebars only fix.

Thank you so much for pointing me in the right direction! :)

tests/fixtures/custom/custom-tsoa-template.ts.hbs
dgreene1: I understand handlebars so poorly that I'm hoping to not have to change it again after the the fix (commit https://github.com/lukeautry/tsoa/pull/495/commits/457407cc8d5840026879471f7d32fb63169d694c). But as you can probably tell by my commit message haha (`"fixes the broken unit test. We really should refactor the handlebars templates at some point so these mistakes are caught in the typescript compiler"`)... I think it would be awesome if none of the JSON transforms for models lived in handlebars. Maybe that's a future refactor PR?

Like it would be awesome if that whole section of the template was:
```hbs
const models: TsoaRoute.Models =  {{{modelsTemplateFunction models}}};
};
```
* where `modelsTemplateFunction` is a TypeScript transformer function that does all of the messy bits that are currently in the template.

> Can we use the new check here?

In summary, "yes" but "not as part of this PR please."

src/routeGeneration/tsoa-route.ts
```diff
@@ -13,20 +13,34 @@ export namespace TsoaRoute {
...
+
+  export interface RefEnumModelSchema {
+    dataType: 'refEnum';
+    enums: string[] | number[];
+  }
+
+  export interface RefObjectModelSchema {
+    dataType: 'refObject';
+    properties: { [name: string]: PropertySchema };
     additionalProperties?: boolean | PropertySchema;
```
WoH:
```suggestion
     additionalProperties: boolean | PropertySchema;
```

I know this will cause TS to complain, but fixing the complaints might be worth the additional safety, what do you think?.

src/routeGeneration/routeGenerator.ts
```diff
@@ -137,25 +138,37 @@ export class RouteGenerator {
...
-        properties: Object.keys(properties).length === 0 ? undefined : properties,
-      } as TsoaRoute.ModelSchema;
-      if (referenceType.additionalProperties) {
-        modelSchema.additionalProperties = this.buildProperty(referenceType.additionalProperties);
-      } else if (this.minimalSwaggerConfig.noImplicitAdditionalProperties) {
-        modelSchema.additionalProperties = false;
+
+        const refObjModel: TsoaRoute.RefObjectModelSchema = {
+          dataType: 'refObject',
+          properties: propertySchemaDictionary,
```
WoH:
```suggestion
          properties: propertySchemaDictionary,
          additionalProperties: false,
```

src/utils/internalTypeGuards.ts
```diff
@@ -0,0 +1,56 @@
+// This file is designed to contain functions that narrow the input type to a type within src\metadataGeneration\tsoa.ts
+import { Tsoa } from '../metadataGeneration/tsoa';
+import { assertNever } from './assertNever';
+
+/**
+ * This will help us do exhaustive matching against only reference types. For example, once you have narrowed the input, you don't then have to check the case where it's a `integer` because it never will be.
+ */
+export function isRefType(metaType: Tsoa.Type): metaType is Tsoa.ReferenceType {
```
WoH:
```suggestion
export function isRefType(type: Tsoa.Type): type is Tsoa.ReferenceType {
```

src/routeGeneration/tsoa-route.ts
dgreene1: I understand where you’re coming from, but Swagger doesn’t require that field. So I personally feel like that makes debugging things harder.

WoH:
@dgreene1 This is the wrong place, but can you open the 3.x branch so I can rebase my PRs? Luke hasn't responded to my request yet.

dgreene1:
Btw, @WoH I'd really like to take this further since the biggest DX problems (for me) are in TypeResolver. But when I tried to use exhaustiveness checking there, it uncovered that there are quite a few types that we don't specifically validate and instead we just "fallthrough" to `this.validateModel`.

image

What do you think?

New PR? Leave it alone?

Personally, I'd love to see function get upgraded to use more narrow types. Like once you know that `dataType` is `"enum"` then you know that it's `.enums` property will be there. So the `members?` parameter of `validateEnum` can become `members` (i.e. non-nullable)
So, instead of:
```ts
public validateEnum(name: string, value: any, fieldErrors: FieldErrors, members?: Array<string | number>, parent = ''): any {
```
```ts
public validateEnum(name: string, value: any, fieldErrors: FieldErrors, members: Array<string | number>, parent = ''): any {
```

That's just one example. See, this is why I don't like to refactor for refactoring sake... cause once I start I can't stop myself! haha

WoH:
What you are pointing out is an issue that should be considered a bug discovered by refactoring.
I'd suggest submitting a PR that addresses these bugs.

Fyi, the check for `.ref` makes sure we only validate the model, so they don't fall through to validate model, they just don't get checked.

A simple refactoring PR that makes this explicit and creates a branch for each of the unvalidated types should be pretty minimal.

PR #498, opened 2019-10-03
Add support for using @Hidden() decorator on controllers 

ryankeener:
Adds support for the `@Hidden()` decorator to controllers which will hide all of its endpoints.

### All Submissions:

> I followed the pattern of disallowing duplicate definitions of decorators but I don't see any existing tests for them. I can add them if desired.
* [x] This PR is associated with an existing issue? #499 

None that I can think of.

### DIFF
--- a/tests/unit/swagger/definitionsGeneration/metadata.spec.ts
+++ b/tests/unit/swagger/definitionsGeneration/metadata.spec.ts
@@ -503,6 +503,22 @@ describe('Metadata generation', () => {
     });
   });
 
+  describe('HiddenControllerGenerator', () => {
+    const parameterMetadata = new MetadataGenerator('./tests/fixtures/controllers/hiddenController.ts').Generate();
+    const controller = parameterMetadata.controllers[0];
+
+    it('should generate hidden methods', () => {
+      const method = controller.methods.find(m => m.name === 'hiddenGetMethod');
+      if (!method) {
+        throw new Error('Method hiddenGetMethod not defined!');
+      }
+
+      expect(method.method).to.equal('get');
+      expect(method.path).to.equal('hiddenGetMethod');
+      expect(method.isHidden).to.equal(true);
+    });
+  });
+
   describe('DeprecatedMethodGenerator', () => {
     const parameterMetadata = new MetadataGenerator('./tests/fixtures/controllers/deprecatedController.ts').Generate();
     const controller = parameterMetadata.controllers[0];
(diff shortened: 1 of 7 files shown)


tests/unit/swagger/definitionsGeneration/metadata.spec.ts
```diff
@@ -503,6 +503,22 @@ describe('Metadata generation', () => {
     });
   });
 
+  describe('HiddenControllerGenerator', () => {
+    const parameterMetadata = new MetadataGenerator('./tests/fixtures/controllers/hiddenController.ts').Generate();
+    const controller = parameterMetadata.controllers[0];
+
+    it('should generate hidden methods', () => {
```
dgreene1: I'm a little confused by the message `'should generate hidden methods'`. Do you mean `'should generate metadata for the hidden methods even though it won't generate swagger definitions'`?

Speaking of which, can you add a unit test that confirms that the swagger output does not contain the methods of the hidden controller?

I'm asking because I believe that is the primary goal of the `@Hidden` decorator, right?

You'll need to do those in two places:
* For Swagger (i.e. OpenAPI v2) `tests\unit\swagger\definitionsGeneration\definitions.spec.ts`
* For OpenAPI 3: `tests\unit\swagger\schemaDetails3.spec.ts`

Other than that, this is a great PR. Thank you for adding to the readme. Well done! :)

ryankeener: Sure thing, I can make those changes.  The test message was just a victim of copy & paste from the existing `@Hidden` tests.

ryankeener: > For Swagger (i.e. OpenAPI v2) `tests\unit\swagger\definitionsGeneration\definitions.spec.ts`

Can you provide a little more guidance on this? I don't see any `paths` tests in that file.

dgreene1: You're correct. I just checked and there are not any path tests in that file. So I'm not sure where it's best to put those tests, but we do need to make sure that this decorator works for both Swagger 2.0 and Open API 3.0

tests/unit/swagger/schemaDetails3.spec.ts
```diff
@@ -126,6 +126,21 @@ describe('Definition generation for OpenAPI 3.0.0', () => {
...
     });
+    describe('hidden paths', () => {
+      it('should not contain hidden paths', () => {
+        const metadataHiddenMethod = new MetadataGenerator('./tests/fixtures/controllers/hiddenMethodController.ts').Generate();
+        const specHiddenMethod = new SpecGenerator3(metadataHiddenMethod, getDefaultOptions()).GetSpec();
+
+        expect(specHiddenMethod.paths).to.have.keys(['/Controller/normalGetMethod']);
+      });
+
+      it('should not contain paths for hidden controller', () => {
```
dgreene1: So basically the same thing you did here but with SpecGenerator2 instead. And just put it anywhere other than this file.

ryankeener:
@dgreene1 should be all set now

PR #508, opened 2019-10-08
Single literal type

### LINKED ISSUE #320: Single literal type does not work correctly
```
interface WorksCorrectly {
  myLiteral: 'hello' | 'world';
}

interface DoesNotWork {
  myLiteral: 'hello';
}
```

I believe that this should produce an enum that allows a single item.

A workaround is to declare it as a typescript enum, or a string literal with two redundant values
```
enum SingleValueEnum {
  HELLO = 'hello',
}

interface WorkAround {
  myLiteral: SingleValueEnum;
}

interface WorkAroundUglier {
  myLiteral: 'hello' | 'hello';
}
```

WoH:
### All Submissions:

**Closing issues**

Closes #320 

### DIFF
--- a/src/metadataGeneration/tsoa.ts
+++ b/src/metadataGeneration/tsoa.ts
@@ -157,7 +157,7 @@ export namespace Tsoa {
    */
   export interface EnumType extends TypeBase {
     dataType: 'enum';
-    enums: Array<string | number>;
+    enums: Array<string | number | boolean>;
   }
 
   export interface ArrayType extends TypeBase {
--- a/src/metadataGeneration/tsoa.ts
+++ b/src/metadataGeneration/tsoa.ts
@@ -157,7 +157,7 @@ export namespace Tsoa {
    */
   export interface EnumType extends TypeBase {
     dataType: 'enum';
-    enums: Array<string | number | boolean>;
+    enums: Array<string | number>;
   }
 
   export interface ArrayType extends TypeBase {
--- a/src/metadataGeneration/typeResolver.ts
+++ b/src/metadataGeneration/typeResolver.ts
@@ -213,20 +213,20 @@ export class TypeResolver {
     return enumOrReferenceType;
   }
 
-  private resolveLiteralValue(typeNode: ts.LiteralTypeNode): string | number | boolean {
+  private resolveLiteralValue(typeNode: ts.LiteralTypeNode): string {
     let value: boolean | number | string;
     switch (typeNode.literal.kind) {
       case ts.SyntaxKind.TrueKeyword:
-        value = true;
+        value = 'true';
         break;
       case ts.SyntaxKind.FalseKeyword:
-        value = false;
+        value = 'false';
         break;
       case ts.SyntaxKind.StringLiteral:
         value = typeNode.literal.text;
         break;
       case ts.SyntaxKind.NumericLiteral:
-        value = parseFloat(typeNode.literal.text);
+        value = typeNode.literal.text;
         break;
       default:
         if (typeNode.literal.hasOwnProperty('text')) {
--- a/src/metadataGeneration/typeResolver.ts
+++ b/src/metadataGeneration/typeResolver.ts
@@ -54,16 +54,7 @@ export class TypeResolver {
       if (supportType) {
         const enumMetaType: Tsoa.EnumType = {
           dataType: 'enum',
-          enums: (this.typeNode.types as ts.NodeArray<ts.LiteralTypeNode>).map(type => {
-            switch (type.literal.kind) {
-              case ts.SyntaxKind.TrueKeyword:
-                return 'true';
-              case ts.SyntaxKind.FalseKeyword:
-                return 'false';
-              default:
-                return String((type.literal as ts.LiteralExpression).text);
-            }
-          }),
+          enums: (this.typeNode.types as ts.NodeArray<ts.LiteralTypeNode>).map(type => this.resolveLiteralValue(type)),
         };
         return enumMetaType;
       } else {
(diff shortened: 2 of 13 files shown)


dgreene1:
Do you still want that this to be a v3 thing? I only ask because this PR is targeting master, but you previous comment said you wanted it to be in the v3 branch: https://github.com/lukeautry/tsoa/issues/320#issuecomment-537502957

src/metadataGeneration/tsoa.ts
```diff
@@ -157,7 +157,7 @@ export namespace Tsoa {
    */
   export interface EnumType extends TypeBase {
     dataType: 'enum';
-    enums: Array<string | number>;
+    enums: Array<string | number | boolean>;
```
dgreene1: Please keep the type the same as it was because enums can not have `boolean` as a value type. See [this playground](http://www.typescriptlang.org/play/index.html#code/KYDwDg9gTgLgBMAdgVwLZwJIGcvOAZgCYAGAURAENUwAbYOAbwFgAoOduCgZRigEtEAczgBeOACJuvAYPEAaVh04A5NACNgUUXEIK2HNRAh0KiLAEEowAEIUAJtt55WAXyA) which shows that you would get `"Computed values are not permitted in an enum with string valued members."` if you tried.

WoH:
> Do you still want that this to be a v3 thing? I only ask because this PR is targeting master, but you previous comment said you wanted it to be in the v3 branch: #320 (comment)

I am aware, this is the bare minimum, fix the version thing. However, this allows me, in combination with aliases, to remove a lot of code for the 3.x branch on top. I'll port this to 3.0 and move from there.

> Please keep the type the same as it was because enums can not have boolean as a value type. See this playground which shows that you would get "Computed values are not permitted in an enum with string valued members." if you tried.

This is not relevant here, since we are not talking about ts enums, which should be refEnums, but swagger/openapi enums, which allow booleans. Or am I missing something here?

src/metadataGeneration/tsoa.ts
WoH: See https://github.com/lukeautry/tsoa/pull/508#issuecomment-539665332

dgreene1: I think it will be confusing for maintainers.

> This is not relevant here I believe, since we are not talking about ts enums, which should be refEnums, but swagger/openapi enums, which allow booleans. Or am I missing something here?

So I’m okay with the enum in Swagger.d.ts being described as having a Boolean in there. But it doesn’t make sense for the places where a TypeScript enum is parsed into tsoa metadata.

I have a feeling that you’re using the swagger enum field as a place to store aliases. But I find that incredibly confusing.

I have to think about this some more.

WoH: At least I'm not alone :P
The way we (ab)use the extractEnum flag to [check if we support certain constructs](https://github.com/lukeautry/tsoa/blob/master/src/metadataGeneration/parameterGenerator.ts#L193) is incredibly confusing. (See the references to the linked method)

WoH: I can only speculate the reasoning here is to see if we can transform the parameter try into a string, which are the only valid param types in Header, Query and Path.
There, it makes sense to only allow strings and not even numbers.

WoH: I think since we allow numbers, I don't really see why booleans wouldn't be allowed either, but I am fine with casting keeping the casts fo String for now as well, but we should refactor this for 3.0, it's confusing and buggy.

dgreene1: Agreed. I don’t think we should cast a type to string.

dgreene1: Also, I’d like to get rid of `extractEnum` so if you can refactor that out, that would be amazing! :)

WoH: I already did, I am at a point where my cross dependencies are growing though.

Turns out determining the correct properties for 
```typescript
export interface TestSubModel2 extends TestSubModel, Partial<Omit<UserResponseModel, 'id'>> {
  testSubModel2: boolean;
}
```
requires quite a few steps.

WoH: https://github.com/lukeautry/tsoa/pull/508/commits/5bded384ea75b0af32d5c59629e4f8fddef01a89

I'm not gonna even pretend I like this, but I guess this is the most compatible way to address this.

dgreene1: @WoH in light of how busy I will be in the following weeks (possibly 2-3 months), I’m just gonna let you do it how you want. So please revert the last commit and just add a JSDoc comment above `enums` that explains what you said about how Swagger supports enums with Booleans.

WoH: @dgreene1 I think it's fine to keep the implementation as compatible as possible. Maybe I voiced my concerns a bit too much. I fully understood your reasoning, I just don't think we should cherry-pick 5bded38 into 3.0 where we have an opportunity to do things better.

dgreene1: Okay, but in your 3.0 branch, can you add a JSDoc comment that explains that OpenAPI supports enums with booleans? I'd ask you to link to the OpenAPI spec that explains that but I can't actually find it. As for this PR, I'll merge it.

WoH:
```yaml
components:
  schemas:
    EnumIndexValue:
      type: boolean
      enum:
        - false
```

Is valid as per JSON Schema draft 0, which Swagger & OpenAPI build on top, so they probably just refer you to this spec.
https://json-schema.org/specification-links.html#draft-0
https://tools.ietf.org/html/draft-zyp-json-schema-00#section-5.16

dgreene1: I've made a separate ticket for just this: https://github.com/lukeautry/tsoa/issues/510

PR #509, opened 2019-10-10
TypeResolver: Add defaults for type parameters to context

WoH:
This is a backport from #505 because it's possible to have and review this on it's own.

```ts
interface MyModel<T = number> {
  numberOrT?: T
}

interface ReturnedFromController {
  model: MyModel
}
```
Now works, assuming your default is something we can resolve.

### All Submissions:

* [ ] This PR is associated with an existing issue?

**Closing issues**

Haven't seen any issues related to this specifically.

I haven't written negative cases because that would mean there's a model without a default that doesn't get any arguments and TypeScript ensures that code wouldn't even compile.

### DIFF
--- a/src/metadataGeneration/typeResolver.ts
+++ b/src/metadataGeneration/typeResolver.ts
@@ -191,9 +191,7 @@ export class TypeResolver {
       return literalType;
     }
 
-    if (typeReference.typeArguments && typeReference.typeArguments.length > 0) {
-      this.typeArgumentsToContext(typeReference, typeReference.typeName, this.context);
-    }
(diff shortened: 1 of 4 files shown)


WoH:
Basically, this just removes the check if we have type arguments before populating the new context (since they might be the default on the parameter).
In case we don't have values for the parameters (from context or from type arguments), this introduces a new check for defaults on the type parameter. 

Since this code only affects metadata generation (which impacts spec generation) and not validation (we still validate models, no new types etc.) the specs just make sure the default gets applied within the model and propagated to child models.

PR #524, opened 2019-11-05
fix to generate OpenAPI3 requestBody output for @BodyProps

SmerkyG:
Small change to support requestBody in OpenAPI3. Inserts schema directly into requestBody.content.application/json.schema rather than using a ref. There didn't appear to be any test coverage for OpenAPI3 output whatsoever that I could find, unfortunately. 

@dimitor115 feel free to look at this code and use it for reference for your pull request, if you think it needs more work or you want to add test coverage, etc.

### DIFF
--- a/src/swagger/specGenerator3.ts
+++ b/src/swagger/specGenerator3.ts
@@ -178,6 +178,7 @@ export class SpecGenerator3 extends SpecGenerator {
     }
 
     const bodyParams = method.parameters.filter(p => p.in === 'body');
+    const bodyPropParams = method.parameters.filter(p => p.in === 'body-prop');
 
     pathMethod.parameters = method.parameters
       .filter(p => {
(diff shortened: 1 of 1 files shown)


dgreene1:
@SmerkyG can you please add the unit tests so we can merge this?

joppino:
Can you ropen this so I can add the unit tests as requested in #596 ?

dgreene1:
@joppino I reopened it for you

SmerkyG:
Thanks for doing that. I would have originally, but there were literally no
openapi3 tests to use for reference and at the same time another person had
said they were working separately on the same feature.

joppino:
@dgreene1 I see there are no swagger unit tests with specGenerator3 (all the tests are using specGenerator2). Do you prefer a single test case for this patch or a new harness for all specGenerator3? And for the latter can you suggest a preferred directory/file structure?

dgreene1:
@joppino there are OpenApi 3 tests. I think the file is called defintions3 or something. Sorry I can’t find the name right now but I’m on mobile.

There are also integration tests too that have real tsoa controllers which is where you might need to add it.

Please reach out to @WoH if you can’t find them.

joppino:
Yep I think I found them. Letting you know soon

PR #526, opened 2019-11-09
fix(model-validation): only set properties in tsoa request body if provided or required

### LINKED ISSUE #525: about undefined properties in tsoa request body
## Sorting

* **I'm submitting a ...**
  - [X] bug report
  - [ ] feature request
  - [X] support request

* I confirm that I

## Issue / Question

I have a simple input post body e.g.: 
```ts
interface InputDTO {
	firstName?: string;
	lastName?: string;
	email?: string;
	phone?: string;
	birthday?: Date;
	gender?: Gender;
	nationality?: string;
}
 enum Gender {
	Male = 'male',
	Female = 'female'
}

     @Post('/test')
	async test(
		@Request() req,
		@Body() body: InputDTO
	): Promise<{
		userUpdate: { [key: string]: any };
	}> {
		console.log('body', body);
       }
```

and if I call it with "firstName" only, all other properties are stil there, but undefined.
```js
{
  firstName: 'safd',
  lastName: undefined,
  email: undefined,
  phone: undefined,
  birthday: undefined,
  gender: undefined,
  nationality: undefined
}
```

the question is, is this expected? Therefore I cannot distinguish between calling "firstName: undefined" and firstName not provided at all.
The check 
```js 
if ('firstName' in body) {
    console.log('firstName provided');
}
``` 
is always true.

## Possible Solution
Only set values if they are provided and set undefined as fallback if the property exists in the request but has no value.

## Context (Environment)
Version of the library: tsoa@2.5.8
Version of NodeJS: v12.13.0

simllll:
### All Submissions:

* [X] Have you written unit tests that cover the negative cases (i.e.: if bad data is submitted, does the library respond properly)?

**Closing issues**

closes #525 

it changes non defined properties in body (in models) so that they do not exist in there anymore, before that they were undefined even if they were not provided. So far as I understand this is a more correct approach and shouldn't break anything.

**Test plan**

I instanciate the validator and check if non provided values are still show up, besides that I check if require fields are still valiated even if they are not provided.

### DIFF
--- a/tests/unit/swagger/templateHelpers.spec.ts
+++ b/tests/unit/swagger/templateHelpers.spec.ts
@@ -207,7 +207,7 @@ describe('ValidationService', () => {
       };
       const result = v.validateModel({ name: '', value: {}, refName: 'ExampleModel', fieldErrors: error, minimalSwaggerConfig });
       expect(Object.keys(error)).to.be.empty;
-      expect(result).to.eql({ a: undefined });
+      expect(result).to.eql({});
     });
 
     it('should validate a model with additional properties', () => {
@@ -246,7 +246,7 @@ describe('ValidationService', () => {
       };
       const result = v.validateModel({ name: '', value: {}, refName: 'ExampleModel', fieldErrors: error, minimalSwaggerConfig });
       expect(Object.keys(error)).to.be.empty;
-      expect(result).to.eql({ a: undefined });
+      expect(result).to.eql({});
     });
 
     it('should validate additional properties only against non-explicitly stated properties', () => {
@@ -271,6 +271,71 @@ describe('ValidationService', () => {
       const result = v.validateModel({ name: '', value: { a: 9 }, refName: 'ExampleModel', fieldErrors: error, minimalSwaggerConfig });
       expect(Object.keys(error)).to.be.empty;
       expect(result).to.eql({ a: 9 });
+    });
+
+    it('non provided parameters should not result in undefined', () => {
+      const v = new ValidationService({
+        BEnum: {
+          dataType: 'refEnum',
+          enums: ['X', 'Y'],
+        },
+        General: {
+          dataType: 'refObject',
+          properties: {
+            a: { dataType: 'string' },
+            b: { ref: 'BEnum' },
+            c: { dataType: 'string' },
+          },
+          additionalProperties: false,
+        },
+      });
+
+      const error = {};
+
+      const result = v.ValidateParam(
+        { required: true, ref: 'General' },
+        {
+          a: 'value',
+          b: undefined,
+        },
+        'body',
+        error,
+        undefined,
+        {},
+      );
+
+      expect(result).to.deep.equal({ a: 'value', b: undefined });
+      expect(Object.keys(error)).to.be.empty;
+      expect('a' in result).to.be.true; // provided
+      expect('b' in result).to.be.true; // provided, but empty
+      expect('c' in result).to.be.false; // not provided
+    });
+
+    it('required provided parameters should result in required errors', () => {
+      const v = new ValidationService({
+        General: {
+          dataType: 'refObject',
+          properties: {
+            a: { dataType: 'string', required: true },
+          },
+          additionalProperties: false,
+        },
+      });
+
+      const error: any = {};
+
+      v.ValidateParam(
+        { required: true, ref: 'General' },
+        {
+          c: 'value',
+        },
+        'body',
+        error,
+        undefined,
+        {},
+      );
+
+      expect(error.a.message).to.equal(`'a' is required`);
     });
   });
 
--- a/src/routeGeneration/templateHelpers.ts
+++ b/src/routeGeneration/templateHelpers.ts
@@ -622,7 +622,11 @@ export class ValidationService {
 
       keysOnPropertiesModelDefinition.forEach((key: string) => {
         const property = properties[key];
-        value[key] = this.ValidateParam(property, value[key], key, fieldErrors, parent, swaggerConfig);
+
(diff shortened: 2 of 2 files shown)


simllll:
Attention: I was not able to run all the test locally due to two issues:
* first I use npm
* but more importantely, I got typescript errors in tests/prepare.ts that scopes is not valid in tsoa_auth (it is part of this commit). The error is valid, according to the defintion oauth2 implicit has no scopes.

tests/unit/swagger/templateHelpers.spec.ts
```diff
@@ -246,7 +246,7 @@ describe('ValidationService', () => {
       };
       const result = v.validateModel({ name: '', value: {}, refName: 'ExampleModel', fieldErrors: error, minimalSwaggerConfig });
       expect(Object.keys(error)).to.be.empty;
-      expect(result).to.eql({ a: undefined });
```
dgreene1: I was nervous about this part. It looks to me like this is a breaking change. But I think it's an acceptable change (especially since it's associated with a big fix). But I'd like @WoH to confirm if he agrees that this is an acceptable change to fall within a minor version release.

tests/prepare.ts
```diff
@@ -20,10 +20,10 @@ const defaultOptions: SwaggerConfig = {
     tsoa_auth: {
       authorizationUrl: 'http://swagger.io/api/oauth/dialog',
       flow: 'implicit',
-      scopes: {
+      /*scopes: {
         'read:pets': 'read things',
         'write:pets': 'modify things',
-      },
+      },*/
```
dgreene1: I've never has this issue when compiling. So it's really surprising to me that you're having this issue. Perhaps you have the wrong stuff in your node_modules?

Can you please:
1) delete your node modules folder
2) install yarn globally
3) yarn install
4) revert this change (since I'm not sure what this breaks)
5) rerun the tests

simllll: I believe this error is absolutely valid, look into: swagger/swagger.ts
  export interface OAuth2ImplicitSecurity {
    type: 'oauth2';
    description?: string;
    flow: 'implicit';
    authorizationUrl: string;
  }
OAuth2PasswordSecurity, OAuth2ApplicationSecurity and OAuth2AccessCodeSecurity have a scopes parameter, but OAuth2ImplicitSecurity simply does not have it... so quite unsure why this I'm the only one who gets this error? Maybe a newer typescript version shows this error (I'm on 3.7.2)

tests/unit/swagger/templateHelpers.spec.ts
simllll: Yeah just wanted to update the PR description about it, please double check if this does not break anything. It shouldn't and I this is actually the "new" expected behaviour if this PR will be approved ;)

tests/prepare.ts
simllll: I have reverted this change on my side because I think this should probably not be part of this PR, but the issue still exists on my side:
Error:(23, 7) TS2322: Type '{ api_key: { in: "query"; name: string; type: "apiKey"; }; tsoa_auth: { authorizationUrl: string; flow: "implicit"; scopes: { 'read:pets': string; 'write:pets': string; }; type: "oauth2"; }; }' is not assignable to type '{ [name: string]: Security; }'.
  Property 'tsoa_auth' is incompatible with index signature.
    Type '{ authorizationUrl: string; flow: "implicit"; scopes: { 'read:pets': string; 'write:pets': string; }; type: "oauth2"; }' is not assignable to type 'Security'.
      Object literal may only specify known properties, and 'scopes' does not exist in type 'OAuth2ImplicitSecurity'.

dgreene1: so I can't explain the error or lack of error (because I'm on not on my laptop), but I can clarify the OAuth stuff.

Basically our security types are wrong. Check out the canonical swagger types that live in definitely typed. They show that scopes is optional is on all flows. https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/swagger-schema-official/index.d.ts#L216

So basically, I'd leave prepare as it was before (I.e. Revert the change), but update our internal swagger type definitions to allow scopes to be passed for implicit type like it the [linked types](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/swagger-schema-official/index.d.ts#L216) have it.

tests/unit/swagger/templateHelpers.spec.ts
dgreene1: Yea, the only way to know if it breaks consumers is to solicit more input from other users. So besides @WoH, I'll also tag @lukeautry, @endor, @rgripper, @atomicus, @egrando.

tests/prepare.ts
simllll: fixed it over here: https://github.com/lukeautry/tsoa/pull/527/files

tests/unit/swagger/templateHelpers.spec.ts
fhewitt: Viewing this PR, I double checked our code and found effectively a small issue with a check on a boolean value in a partial patch endpoint. Easily fixed in our case with a `=== undefined` instead of `'isActive' in` because we don't need to do the distinction. But yet, the current behavior is completely unexpected and I see no case where you would use the current behavior.

simllll: sounds good to me, basically means in best case we can even fix some unknown bugs for some.. ;)

simllll:
as we are kinda relying on this, and therefore really looking forward to get this released / merged,.... I allow myself to ping you all again @WoH, @lukeautry, @endor, @rgripper, @atomicus, @egrando!
Any feedback? :) Thanks

PR #547, opened 2019-12-18
Allow relative locations for OpenApi server URLs

### LINKED ISSUE #540: Support for relative `host` in openapi 3
## Sorting

* **I'm submitting a ...**
  - [ ] bug report
  - [x] feature request
  - [ ] support request

* I confirm that I

## Expected Behavior

Have in the output something like
```json
	"servers": [
		{
			"url": "/v1"
		}
	]
```

I'd like to be able to specify relative `host`, which is allowed by specification https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#fixed-fields-4.

## Current Behavior
Currently if I omit `host` field in the `swagger` config or use empty `"host": ""`, it gets replaced with `http://localhost:3000`. Using `"host": "/"` also doesn't work.

## Context (Environment)
Version of the library: 2.5.9
Version of NodeJS: 12
* Confirm you were using yarn not npm: [ ]

jreusch:
### All Submissions:

**Closing issues**
closes #540 

This introduces the possibility of having relative URLs by setting the host explicitly to null.
Doing it that way will not introduce a breaking change.
Although as a downside i do not currently see a possibility to specify this value as a argument to the _--host_ cli parameter

It was suggested in the issue that in the next major version this behavior could be adapted that _all_ values of host which evaluate to _false_ could behave that way, could there be anything more done to not forget about this adaption for this next version? i.e. is there a list of ideas for breaking changes for the next version?

**Test plan**

Added test which explicitly checks if the URL is only "/v1/" which is in accordance to the OA 3.0 specs (see issue)

### DIFF
--- a/src/swagger/specGenerator3.ts
+++ b/src/swagger/specGenerator3.ts
@@ -94,10 +94,18 @@ export class SpecGenerator3 extends SpecGenerator {
   private buildServers() {
     const basePath = normalisePath(this.config.basePath as string, '/', undefined, false);
     const scheme = this.config.schemes ? this.config.schemes[0] : 'https';
-    const host = this.config.host || 'localhost:3000';
+
+    let url;
+    if (this.config.host === null) {
+      url = basePath;
+    } else {
+      const host = this.config.host || 'localhost:3000';
+      url = `${scheme}://${host}${basePath}`;
+    }
+
     return [
       {
-        url: `${scheme}://${host}${basePath}`,
+        url,
       } as Swagger.Server,
     ];
   }
--- a/tests/unit/swagger/schemaDetails3.spec.ts
+++ b/tests/unit/swagger/schemaDetails3.spec.ts
@@ -18,16 +18,13 @@ describe('Definition generation for OpenAPI 3.0.0', () => {
   const optionsWithNoAdditional = Object.assign<{}, SwaggerConfig, Partial<SwaggerConfig>>({}, defaultOptions, {
     noImplicitAdditionalProperties: 'silently-remove-extras',
   });
-  const optionsWithHostIsNull = Object.assign<{}, SwaggerConfig, Partial<SwaggerConfig>>({}, defaultOptions, {
-    host: null,
-  });
 
   interface SpecAndName {
     spec: Swagger.Spec3;
     /**
      * If you want to add another spec here go for it. The reason why we use a string literal is so that tests below won't have "magic string" errors when expected test results differ based on the name of the spec you're testing.
      */
-    specName: 'specDefault' | 'specWithNoImplicitExtras' | 'specWithHostIsNull';
+    specName: 'specDefault' | 'specWithNoImplicitExtras';
   }
 
   const specDefault: SpecAndName = {
@@ -37,10 +34,6 @@ describe('Definition generation for OpenAPI 3.0.0', () => {
   const specWithNoImplicitExtras: SpecAndName = {
     spec: new SpecGenerator3(metadata, optionsWithNoAdditional).GetSpec(),
     specName: 'specWithNoImplicitExtras',
-  };
-  const specWithHostIsNull: SpecAndName = {
-    spec: new SpecGenerator3(metadata, optionsWithHostIsNull).GetSpec(),
-    specName: 'specWithHostIsNull',
   };
 
   const getComponentSchema = (name: string, chosenSpec: SpecAndName) => {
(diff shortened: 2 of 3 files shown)


dgreene1:
>  i.e. is there a list of ideas for breaking changes for the next version?

Yes, if you or someone else wants to open a new PR, I can label it so it shows up in our v3 milestone so we don’t forget to do the “better way.”

Thanks again for submitting this PR.

jreusch:
> Yes, if you or someone else wants to open a new PR

"new PR" means one with exactly the wanted (breaking) behavior?

dgreene1:
> "new PR" means one with exactly the wanted (breaking) behavior?

Yes. Please create a new PR that we will use to track the breaking change so that we complete that in a few weeks/months when we get to v3 of tsoa.

dgreene1:
Sorry, I meant a new issue.

src/swagger/specGenerator3.ts
```diff
@@ -94,10 +94,18 @@ export class SpecGenerator3 extends SpecGenerator {
...
-    const host = this.config.host || 'localhost:3000';
+
+    let url;
+    if (this.config.host === null) {
+      url = basePath;
+    } else {
+      const host = this.config.host || 'localhost:3000';
+      url = `${scheme}://${host}${basePath}`;
+    }
+
```
dgreene1: @jreusch please forgive the delay. This looks good. Please do it for version 2 of the spec too. We have to continue to support Swagger v2 as well as OpenAPI (i.e. Swagger v3 which you made the changes for).

tests/unit/swagger/schemaDetails3.spec.ts
```diff
@@ -18,13 +18,16 @@ describe('Definition generation for OpenAPI 3.0.0', () => {
...
+    host: null,
+  });
 
   interface SpecAndName {
     spec: Swagger.Spec3;
     /**
      * If you want to add another spec here go for it. The reason why we use a string literal is so that tests below won't have "magic string" errors when expected test results differ based on the name of the spec you're testing.
      */
-    specName: 'specDefault' | 'specWithNoImplicitExtras';
+    specName: 'specDefault' | 'specWithNoImplicitExtras' | 'specWithHostIsNull';
```
dgreene1:
```suggestion
    specName: 'specDefault' | 'specWithNoImplicitExtras';
```
**Short version:** please initialize `optionsWithHostIsNull` inside of your `'setting host to null should result in relative URL'` test. Do not add it to the `SpecAndName` interface since that's designed to for the purpose of iterating over schema properties.

**Longer version:** Is it necessary to add a new spec type here? The reason why I ask is that this causes another iteration over the `assertionsPerProperty` record (if it were added to [that array](https://github.com/lukeautry/tsoa/pull/547/files#diff-19811e20ee8ebdc5405ac00d6cd74d94L56))... which is a lot of tests. i.e. it takes a long time. I'm talking about this part:
```ts
const assertionsPerProperty: Record<keyof TestModel, (propertyName: string, schema: Swagger.Spec) => void> 
```

And in the case of your change, there's no reason why the host would change the property schemas. So it doesn't need to be tested in that manner. So `optionsWithHostIsNull` doesn't need to live in the outer scope. It only is used within the `'setting host to null should result in relative URL'` test and that's fine. Let's keep it private to there.

jreusch: ok, i will adapt the PR.
i kind of missed the usecase of specName, i thought this was to collect all variants for different schemas and to have them always checked, just to be sure.

src/swagger/specGenerator3.ts
jreusch: hey hey, i checked this back then but totally forgot to mention:
actually version 2 already seems to behave how we'd like it to have for tsoa v3.

e.g.
defining only basePath in tsoa.json and generating  specs results in
for  v2:
`swagger.json 
"basePath": "/v1",
(host missing)
`
for v3:
`swagger.json
"url": "https://localhost:3000/v1"
`

defining the host as example.com and generating the spec:
v2:
`
"basePath": "/v1",
"host": "example.com"
`
v3
`
 "url": "https://example.com/v1"
`

so currently you already get _different_ results depending if you generate a v2 or v3 spec...

dgreene1: That’s really unfortunate, but I don’t think it changes our plan if releasing this change, then the breaking version later.

jreusch:
all needed and requested changes are now in so far, i hope everything is as expected now.
if still something needs to be done just say it. :)

jreusch:
awesome, thanks! :)

PR #552, opened 2020-01-03
Add ability to handle streams returned from controller

### LINKED ISSUE #551: Ability to handle streams returned from controller
I have a use-case where I want my controller to pipe streams through to the response object (eg. proxying from s3). This isn't possible to be encoded currently, but the changes required seem pretty minimal. 

## Sorting

* **I'm submitting a ...**
  - [x] feature request

* I confirm that I

## Expected Behavior

- Controllers can return a `Readable` stream object instead of a `Buffer`
- The swagger spec should be the same as if you returned a buffer directly

## Possible Solution

- Add check to `promiseHandler` in route templates that pipes the response rather than sending it via `json`
- Swagger generator should be aware of the `Readable` type and output the same return type as a `Buffer`

## Context (Environment)
Version of the library: 2.5.10
Version of NodeJS: 12

## Breaking change?

Not a breaking change.

jtfell:
I have a use-case where I want my controller to pipe streams through to the response object (eg. proxying from s3). This isn't possible to be encoded currently, but the changes required seem pretty minimal. 

My assumptions:

- The swagger spec should be the same as if you returned a buffer directly.
- The route template should pipe the response rather than sending it via `json`

### All Submissions:

Resolves #551

- I couldn't find any relevant tests for controllers using buffers, as the structure here should be quite similar I imagine
- I'm not experienced with Koa or Hapi so wasn't sure how to integrate the `data.pipe(res)` pattern that I used in the express routes template

**Test plan**

Some guidance here would be great.

### DIFF
--- a/src/metadataGeneration/tsoa.ts
+++ b/src/metadataGeneration/tsoa.ts
@@ -81,7 +81,6 @@ export namespace Tsoa {
     | 'date'
     | 'binary'
     | 'buffer'
-    | 'stream'
     | 'byte'
     | 'void'
     | 'object'
@@ -114,7 +113,6 @@ export namespace Tsoa {
     | DateType
     | BinaryType
     | BufferType
-    | StreamType
     | ByteType
     | AnyType
     | RefEnumType
@@ -182,10 +180,6 @@ export namespace Tsoa {
 
   export interface BufferType extends TypeBase {
     dataType: 'buffer';
-  }
-
-  export interface StreamType extends TypeBase {
-    dataType: 'stream';
   }
 
   export interface ByteType extends TypeBase {
--- a/src/metadataGeneration/tsoa.ts
+++ b/src/metadataGeneration/tsoa.ts
@@ -81,6 +81,7 @@ export namespace Tsoa {
     | 'date'
     | 'binary'
     | 'buffer'
+    | 'stream'
     | 'byte'
     | 'void'
     | 'object'
@@ -113,6 +114,7 @@ export namespace Tsoa {
     | DateType
     | BinaryType
     | BufferType
+    | StreamType
     | ByteType
     | AnyType
     | RefEnumType
@@ -180,6 +182,10 @@ export namespace Tsoa {
 
   export interface BufferType extends TypeBase {
     dataType: 'buffer';
+  }
+
+  export interface StreamType extends TypeBase {
+    dataType: 'stream';
   }
 
   export interface ByteType extends TypeBase {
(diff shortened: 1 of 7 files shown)


src/routeGeneration/templates/express.hbs
```diff
@@ -161,7 +161,9 @@ export function RegisterRoutes(app: express.Express) {
 
                 // WARNING: This file was auto-generated with tsoa. Please do not modify it. Re-run tsoa to re-generate this file: https://github.com/lukeautry/tsoa
 
-                if (data || data === false) { // === false allows boolean result
+                if (data && data.readable) {
```
WoH: This is a dangerous check, please make sure this only runs when data is an instance of `Readable`.

Also, please copy this to the other templates.

src/metadataGeneration/tsoa.ts
```diff
@@ -182,6 +184,10 @@ export namespace Tsoa {
     dataType: 'buffer';
   }
 
+  export interface StreamType extends TypeBase {
```
WoH: You can avoid adding a new type (and therefore, a lot of code) if you reuse `BufferType`.

src/routeGeneration/templates/express.hbs
jtfell: Are you suggesting a guard like: `data && data instanceof Readable`? This is not the recommended way according to this SO (from 2014) https://stackoverflow.com/questions/23885095/nodejs-check-if-variable-is-readable-stream. They suggest using duck-typing so something more like:

`data && data.pipe && data._read`

Alternatively [this package](https://github.com/sindresorhus/is-stream) is an option.

I'm open to whichever, what do you prefer? @WoH

src/metadataGeneration/tsoa.ts
jtfell: Good idea. I've pushed a new commit to reuse `BufferType`.

WoH:
I'm not sure we want to handle legacy streams or arbitrarily call pipe on anything that responds to that method, but there's no good solution here that I'm aware of.
I'd accept a check that checks for a method called `pipe` and makes sure the stream is readable.

src/routeGeneration/templates/express.hbs
WoH: Good points. I didn't consider handling anything except the Node ReadableStream. I'm not sure we want to handle legacy streams or arbitrarily call pipe on anything that responds to that method, but there's no perfect solution here that I'm aware of.
I'd accept a check that checks for a function called pipe and makes sure the stream is readable.

jtfell: I've gone with this check: `data && typeof data.pipe === 'function' && data.readable && typeof data._read === 'function'`.

It looks like both Hapi and Koa will automatically handle streams so no further changes should be needed for the other templates.

WoH:
Is there a way to make this work across all supported frameworks?

jtfell:
It should already @WoH - Koa and Hapi both support streams without changes to the templates.

PR #558, opened 2020-01-10
Deprecate extractEnum

WoH:
### All Submissions:

* [ ] This PR is associated with an existing issue?

The enum gets also rendered as a definition which is not used. I think this is fine, however, we may remove it.

### What? Why?

Swagger2 does not support $ref to link to enum definitions in header, query or path parameters.
Body does apparently work.
In order to generate proper definitions, at the time, a new argument was introduced for the type resolver.
However, this causes a lot of additional complexity. Instead of handling this issue in the resolver, I moved it to the Spec2 Generator, where it's a lot easier to inline the referenced TS Enum

### DIFF
--- a/tests/unit/swagger/definitionsGeneration/metadata.spec.ts
+++ b/tests/unit/swagger/definitionsGeneration/metadata.spec.ts
@@ -2,6 +2,8 @@ import { expect } from 'chai';
 import 'mocha';
 import { MetadataGenerator } from '../../../../src/metadataGeneration/metadataGenerator';
 import { Tsoa } from '../../../../src/metadataGeneration/tsoa';
+import { SpecGenerator2 } from '../../../../src/swagger/specGenerator2';
+import { getDefaultOptions } from '../../../fixtures/defaultOptions';
 
 describe('Metadata generation', () => {
   const metadata = new MetadataGenerator('./tests/fixtures/controllers/getController.ts').Generate();
@@ -293,7 +295,7 @@ describe('Metadata generation', () => {
       expect(genderParam.parameterName).to.equal('gender');
       expect(genderParam.description).to.equal('Gender description');
       expect(genderParam.required).to.be.true;
-      expect(genderParam.type.dataType).to.equal('enum');
+      expect(genderParam.type.dataType).to.equal('refEnum');
 
       const nicknamesParam = method.parameters[6] as Tsoa.ArrayParameter;
       expect(nicknamesParam.in).to.equal('query');
@@ -360,7 +362,7 @@ describe('Metadata generation', () => {
       expect(genderParam.parameterName).to.equal('gender');
       expect(genderParam.description).to.equal('Gender description');
       expect(genderParam.required).to.be.true;
-      expect(genderParam.type.dataType).to.equal('enum');
+      expect(genderParam.type.dataType).to.equal('refEnum');
     });
 
     it('should generate an header parameter', () => {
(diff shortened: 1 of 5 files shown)


tests/unit/swagger/definitionsGeneration/metadata.spec.ts
```diff
@@ -474,6 +476,27 @@ describe('Metadata generation', () => {
       expect(parameter.parameterName).to.equal('firstname');
       expect(parameter.required).to.be.true;
     });
+
+    it('Should inline enums for TS Enums in path, query and header when using Swagger', () => {
+      const spec = new SpecGenerator2(parameterMetadata, getDefaultOptions()).GetSpec();
+      const method = spec.paths['/ParameterTest/Path/{firstname}/{last_name}/{age}/{weight}/{human}/{gender}']['get'];
```
dgreene1: I'm a bit confused by the use of curly braces. Is this string interpolation and you forgot to include a dollar sign? Like {x} instead of ${x}? Sorry for not understanding what's happening here.

tests/unit/swagger/definitionsGeneration/metadata.spec.ts
```diff
@@ -474,6 +476,27 @@ describe('Metadata generation', () => {
       expect(parameter.parameterName).to.equal('firstname');
       expect(parameter.required).to.be.true;
     });
+
+    it('Should inline enums for TS Enums in path, query and header when using Swagger', () => {
```
dgreene1: Does this test still pass without the refactoring that you did? I'm trying to make sense of if this is a change in behavior or a functionality change. I don't care which one it is, I'm just trying to understand so I can better serve the community if we get questions later.

tests/unit/swagger/definitionsGeneration/metadata.spec.ts
WoH: It's the wat to get to the part of the spec for [this](https://github.com/lukeautry/tsoa/blob/master/tests/fixtures/controllers/parameterController.ts#L54) controller method

tests/unit/swagger/definitionsGeneration/metadata.spec.ts
WoH: Well there are 2 things, I shoulda explained earlier.
While the end result should be equal in terms of swagger docs, the intermediate representation does change (the metadata).

I'm trying to defer the special case extractEnum tried to solve from the resolver that builds the metadata to the specGenerator for the Swagger2 specification that takes the metadata and produces the swagger.json.

WoH: That means, all the tests that check the Swagger output (and OpenAPI aswell) stay unchanged.
However, I had to adapt the tests that perform assertions on the metadata.

WoH:
Please note that this is just the prequel for many other issues both TypeScript and swagger or OpenAPI enums cause, I'm working on a few ideas for that aswell.
See https://github.com/WoH/tsoa/commit/81b1340a9fd94e4792ea69125e750ec095057cd4 and https://github.com/WoH/tsoa/commit/76217805f7183522a421bbe55d148cf62720b0d5

tests/unit/swagger/definitionsGeneration/metadata.spec.ts
dgreene1: Oh wow, yea. I really am "out of it." Hopefully I can convince my day job to five me a TypeScript project again soon.

PR #568, opened 2020-01-22
Change default behavior when no host is defined

### LINKED ISSUE #555: Omitting the host property should make the host relative
This is the v3 version for the compromise we made in https://github.com/lukeautry/tsoa/pull/547

## Sorting

* **I'm submitting a ...**
  - [ ] bug report
  - [X] feature request
  - [ ] support request

* I confirm that I

## Expected Behavior

Assuming that the `host` property was not provided and the `basePath` is "v1" (correct me on this if I'm wrong @Eoksni / @jreusch-barco) but the desired output would be:
```yaml
"servers": [
		{
			"url": "/v1"
		}
	]
```
I'd like to be able to specify relative host, which is allowed by specification https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#fixed-fields-4.

## Current Behavior
Currently if I omit host field in the swagger config or use empty "host": "", it gets replaced with http://localhost:3000. Using "host": "/" also doesn't work.

## Possible Solution
The solutions were discussed in the v2 version of this change: https://github.com/lukeautry/tsoa/issues/540

## Detailed Description

We will accept that if someone doesn't specify a host, then it should not default (since tsoa couldn't possibly know which host to default to). It should be relative.

## Breaking change?

Yes, this should be part of the v3 branch.

jreusch:
In OpenAPI 3.0 definitions of an url can be absolute and contain an host, or
relative. This commit changes the bahavior of tsoa to generate a
relative url when no host is defined.

In version 2 of this lib when no host was defined it was as default set
to localhost:3000. The possibility to also generate relative urls
without breaking backwards compatability it was chosen to only generate
such urls when host is set tu 'null'.

For spec versoin 2 the urls where always relative, so this commit
realignes the behavior again.

Fixes #555

### All Submissions:

**Closing issues**

closes #555

Since this is changing the default behavior should we maybe mention it somewhere? (e.g. release notes?) and if so where?

### DIFF
--- a/src/swagger/specGenerator3.ts
+++ b/src/swagger/specGenerator3.ts
@@ -117,10 +117,10 @@ export class SpecGenerator3 extends SpecGenerator {
   private buildServers() {
     const basePath = normalisePath(this.config.basePath as string, '/', undefined, false);
     const scheme = this.config.schemes ? this.config.schemes[0] : 'https';
-    const host = this.config.host || 'localhost:3000';
+    const url = this.config.host ? `${scheme}://${this.config.host}${basePath}` : basePath;
     return [
       {
-        url: `${scheme}://${host}${basePath}`,
+        url,
       } as Swagger.Server,
     ];
   }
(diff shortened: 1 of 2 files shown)


dgreene1:
I didn't get a change to mention this but @jreusch thank you for finding the unit tests and adding them. Welcome to the tsoa contributors club! Excellent work! :)

PR #564, opened 2020-01-18
Fix OpenAPI3 QueryArrays

WoH:
### All Submissions:

* [x] Have you followed the guidelines in our [Contributing](https://github.com/lukeautry/tsoa/tree/master/docs/CONTRIBUTING.md) document?
* [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/lukeautry/tsoa/pulls) for the same update/change?
* [ ] Have you written unit tests?
* [ ] Have you written unit tests that cover the negative cases (i.e.: if bad data is submitted, does the library respond properly)?
* [ ] This PR is associated with an existing issue?

OpenAPI 3 uses the equivalent of `collectionFormat: 'multi'` for arrays in querystrings by default (`{style: 'form', explode: true }`).
However, the key `collectionFormat` is no longer allowed and should therefore not be set.

Allowed fields: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#style-values
Defaults: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#fixed-fields-10

PR #575, opened 2020-01-25
Check required nested properties for existence

### LINKED ISSUE #574: required Nested Properties are not validated for existence
## Sorting

* **I'm submitting a ...**
  - [x] bug report
  - [ ] feature request
  - [ ] support request

* I confirm that I

## Expected Behavior
Types containing nested properties which are required should be validated to exist.

```ts
export interface UserCreateRequest {
    firstName: string;
    lastName: string;
    email: string;
    address: {
        country: string;
        state: string;
    };
    phoneNumber?:  string;
}
```

## Current Behavior
Validation is bypassed on fields which are not included in the input.  For example, this is allowed through as a valid request:
```
{
	"firstName": "fanta",
	"lastName": "pop",
	"email": "blah@blah.com",
	"address": {}
}
```

## Possible Solution
It looks like the issue is here: https://github.com/lukeautry/tsoa/blob/282905a260f35ab743f5cdf20b71373f6090f070/src/routeGeneration/templateHelpers.ts#L131-L141

The incoming request object is being looped over to find keys to validate but if the field is missing, there is nothing checking that.

I think the fix is to do something similar to what is being done here:
https://github.com/lukeautry/tsoa/blob/282905a260f35ab743f5cdf20b71373f6090f070/src/routeGeneration/templateHelpers.ts#L626-L633 in which all the propertes in the data model are looped over and then checked for required or not.

## Context (Environment)
Version of the library: 3.0.1
Version of NodeJS:  12.12.0

## Breaking change?
I think technically this is a breaking change because certain api calls which are now being allowed would now fail validation.

WoH:
### All Submissions:

**Closing issues**

Closes #574

### DIFF
--- a/src/routeGeneration/templateHelpers.ts
+++ b/src/routeGeneration/templateHelpers.ts
@@ -128,17 +128,16 @@ export class ValidationService {
       }
     }
 
-    Object.keys(value).forEach(key => {
-      if (!nestedProperties[key]) {
-        if (additionalProperties && additionalProperties !== true) {
-          return this.ValidateParam(additionalProperties, value[key], key, fieldErrors, parent + name + '.', swaggerConfig);
-        } else {
-          return key;
-        }
+    Object.keys(nestedProperties).forEach(key => {
+      const validatedProp = this.ValidateParam(nestedProperties[key], value[key], key, fieldErrors, parent + name + '.', swaggerConfig);
+      if (validatedProp) {
+        value[key] = validatedProp;
       }
-
-      return this.ValidateParam(nestedProperties[key], value[key], key, fieldErrors, parent + name + '.', swaggerConfig);
     });
+
+    if (Object.keys(fieldErrors).length > 0) {
+      return;
+    }
 
     return value;
   }
--- a/tests/integration/dynamic-controllers-express-server.spec.ts
+++ b/tests/integration/dynamic-controllers-express-server.spec.ts
@@ -752,7 +752,7 @@ describe('Express Server', () => {
           expect(body.fields['body.typeAliases.fourtyTwo'].message).to.equal('min 42');
           expect(body.fields['body.typeAliases.unionAlias'].message).to.contain('Could not match the union against any of the items');
           expect(body.fields['body.typeAliases.intersectionAlias'].message).to.equal(
-            `Could not match the intersection against every type. Issues: [{"body.typeAliases.value1":{"message":"'value1' is required"}}]`,
+            `Could not match the intersection against every type. Issues: [{"body.typeAliases.intersectionAlias.value1":{"message":"'value1' is required"}},{"body.typeAliases.value1":{"message":"'value1' is required"}}]`,
           );
           expect(body.fields['body.typeAliases.nOLAlias'].message).to.equal('invalid object');
           expect(body.fields['body.typeAliases.genericAlias'].message).to.equal('invalid string value');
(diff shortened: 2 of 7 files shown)


WoH:
@dgreene1 This looks pretty solid to me, any objections?

In terms of iterating on user input, that was a bad idea anyway, no idea what happened to my brain there. Iterating on the property definition is the fix and the proper way to do this.
It's mostly tests that changed since some of the intersections already break when we can't pass the nOL validation any more.

Targeting 3.x since it might be considered breaking.

If it's not too much to ask, it'd be helpful if you could push a new 3.x, I drafted some of the notes already:


# 3.0.0 Release 3:
## Breaking Changes:

- Improve nested object validation: See #574 


- Change default behavior when no host is defined:

When no host is defined, urls are assumed to be relative. Explicitly set your host in case you want to have absolute urls.

- Remove .. in fieldErrors (145222af5fb044d0de2afb44b88fedefa4cb6124):

When detecting illegal additional properties (if you are using tsoa setting `additionalProperties: 'throw-on-extras'`), the key on the error would contain an additional dot. 

```js
{
  "TestModel..additionalProp: : {
    ...
  }
}
```

This is now fixed and the key is `TestModel.additionalProp`.


- Change naming (| from %7C -> \~OR\~, & from \~AND to \~AND\~)


SwaggerUI has troubles parsing percent encoded URLs that confirm to the OpenAPI standard.
In order to mitigate issues, we changed some of the names. 

Example: `Pick<Model, 'one' | 'two'>` not generates this name in the Spec: `Pick_Model.one~OR~two_`

We hope this avoids issues and conveniently improves legibility.

## Features:
- Unknown type, ported to 3.x: #559

## Bugfix:
- Fix OpenAPI3 QueryArrays, ported to 3.x #564

dgreene1:
My whole household is sick so I’m both recovering and on nurse duties. I probably won’t be able to push a branch until next week. But I set a reminder on both my work and home computers. So I will get to it.

src/routeGeneration/templateHelpers.ts
```diff
@@ -128,18 +128,17 @@ export class ValidationService {
...
+      const validatedProp = this.ValidateParam(nestedProperties[key], value[key], key, fieldErrors, parent + name + '.', swaggerConfig);
+      if (validatedProp) {
+        value[key] = validatedProp;
       }
-
-      return this.ValidateParam(nestedProperties[key], value[key], key, fieldErrors, parent + name + '.', swaggerConfig);
     });
 
+    if (Object.keys(fieldErrors).length > 0) {
+      return;
```
dgreene1: @WoH I approved this PR because you wanted to get it out for people to review, but if you don't mind, please submit a new PR with a comment between lines 138 and 139 that explains why we're not doing anything with the errors. I'll take any code comment at all. When you're scanning the code it's just rather strange.

dgreene1:
@WoH I added your release notes (thank you btw), but I could not say that it was "3.0.0 release 3" because that yarn bug had prevented me from making a `3.0.0.betaX` release. So I just wanted to remind you that the releases so far have been `3.0.0`, `3.0.1`, and `3.0.2` but they're still beta in my mind since no one can get them without explicitly installing them (because the latest tag is still set to the master branch which is v2.5.13 at the time of this writing).

Anyway here you go: https://github.com/lukeautry/tsoa/releases/tag/v3.0.2

Issue #571, opened 2020-01-23
RFC: Rework resolving TypeScript enums


WoH:
# Summary
[summary]: #summary

Currently, enum generation has some room for improvement. It supports constructs that should probably not be supported, in a fashion that leads to problems down the road.

# Motivation
[motivation]: #motivation

Currently, tsoa supports enums of 3 shapes of enum members:

- No initializer
- Literals
- Expressions

```ts
enum MyEnum {
  ONE //EnumMember without initializer
}

enum MyEnum2 {
  TWO = 2, //EnumMember with literal (number or string)
  THREE = 'THREE' as any //EnumMember with Expression (something as any), legal
  FOUR = 4 as any //EnumMember with initializer of type Expression and expression of type NumericLiteral, will silently casted to '4'
  FIVE = {} as any //EnumMember with initializer of type Expression, will silently be casted to '3'(!!!)
}
```

Metadata for MyEnum2:

```js
  "EnumStringValue": {
    "dataType": "refEnum",
    "enums": [2, "THREE", "4", "3"],
  },
```

While MyEnum2.THREE will work, since we are just casting a string enum member as any and tsoa will treat that properly since the current code only supports AsExpression with expression of type StringLiteral. However, Numeric literals would already be silently casted to strings. If the expression does not have a `text` property, the current fallback is `String(index)`, we end up with some pretty unreasonable behavior. (See MyEnum.FIVE)

I would suggest we should simply error if a user provides an Expression. We can only ensure we don't have to deal with arbitrary expressions that will be hard to validate if we cut this feature and address enum members with expressions at a later stage.

# Approach
[approach]: #approach

Whenever we encounter anything but a Numeric|StringLiteral as the initializer of an enum member, we throw an Error. Falling back to index should only be valid if there is explicitly no initializer.

NOTE: We can allow mixing strings and enums here, since we can split the enum into a numeric enum and a string enum combined using `anyOf` in OpenAPI.
Since swagger does not support this, we can save the metadata with the correct format and specify type string enum once we rework Swagger/OpenAPI validation, which will be another RFC at a later point, OR throw when we try to render a mixed enum using the SpecGenerator for Swagger.

# Drawbacks
[drawbacks]: #drawbacks

This is a breaking change.

dgreene1:
I agree, the enum stuff is problematic. I think this is a fine proposal for OpenApi 3 and in our v3 branch. I'd leave the Swagger v2 generation alone (I assume that's what you would have done anyway).

[...]

PR #593, opened 2020-02-22
Fix TS Enum generation

### LINKED ISSUE #501: Enums with strings and numbers should not be supported
The goal is to be more strict in v3 so we can clarify to users why they're getting a bad swagger document when they used mixed enums (i.e. enums with strings and numbers).

See this comment for more information: https://github.com/lukeautry/tsoa/pull/493/files#r330696346

## Sorting

* **I'm submitting a ...**
  - [ ] bug report
  - [X] feature request
  - [ ] support request

* I confirm that I

## Expected Behavior

in the v2 branch, if you pass in an enum with strings and numbers then it tries to build a swagger object, but it shouldn't
```ts
export enum MixedStringAndNumberEnum {
  one = 1,
  two = 'two',
  three = 3,
  four = 'four',
}
```
Due to requirements to not create breaking changes within v2, we had no choice but to build some schema, but it's wrong:
```yaml
paths:
  /items:
    get:
      parameters:
        - in: query
          name: sort
          description: Sort order
          schema:
            type: string // <-- If swagger were to support it, this should be string | number
            enum: [1, two, 3, four]
```
Ideally it would throw an error.

## Possible Solution

* Throw an error that clarifies that swagger/ OpenAPI does not support enums with strings and numbers

## Steps to Reproduce

1. see the unit test labeled `"should warn if an enum is mixed with numbers and strings"`
2. that ^ is the unit test that should be changed to be `"should throw an error if an enum is mixed with numbers and strings"`

## Breaking change?

yes, that's why I'm adding it the the `V3` github project kanban board

WoH:
- Support mixed Enums in OpenAPI 3
- Throw when using mixed Enums in Swagger
- Deprecate enum initializer Expressions, will be ignored
- Change numeric enums to type 'number' instead of 'integer' !

### All Submissions:

* [x] Have you adapted unit tests?
* [x] Have you adapted unit tests that cover the negative cases (i.e.: if bad data is submitted, does the library respond properly)?

**Closing issues**

Closes #501, #571

Breaking change, needs to be documented

### DIFF
--- a/tests/unit/swagger/schemaDetails3.spec.ts
+++ b/tests/unit/swagger/schemaDetails3.spec.ts
@@ -6,6 +6,7 @@ import { SpecGenerator3 } from '../../../src/swagger/specGenerator3';
 import { Swagger } from '../../../src/swagger/swagger';
 import { getDefaultOptions } from '../../fixtures/defaultOptions';
 import { TestModel } from '../../fixtures/duplicateTestModel';
+import { Tsoa } from '../../../src/metadataGeneration/tsoa';
 
 describe('Definition generation for OpenAPI 3.0.0', () => {
   const metadata = new MetadataGenerator('./tests/fixtures/controllers/getController.ts').Generate();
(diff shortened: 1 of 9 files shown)


tests/unit/swagger/schemaDetails3.spec.ts
```diff
@@ -1058,48 +1054,4 @@ describe('Definition generation for OpenAPI 3.0.0', () => {
       });
     });
   });
-
-  describe('illegal input values', () => {
```
dgreene1: Nice work @WoH. Can we replace this deleted test with a positive test so people can see how the enum of strings and numbers becomes a union of two schemas?

WoH: Sure, good point!

WoH: Added: a9e7b50

dgreene1:
@WoH did you see that Luke Autry offered you deployment rights now that I’m looking to take a backseat?

If not, check his email out. He needs your NPM account name.

WoH:
@dgreene1 I forwarded my reply to Luke, apparently it didn't cc you. Haven't heard anything after.

PR #595, opened 2020-02-23
Heller/bugfix/hosts not parsing correctly

givethemheller:
If you enter multiple hosts, the current implimentation will simply stringify the array. This prevents you from using multiple "servers" in version 3.0. This fix allows a user to have multiple servers by enforcing host<>scheme pairing when there are multiple hosts specified. This is non breaking as anyone upgrading will see the same behavior. The behavior only requires pairing when adding multiple hosts. So, if you have http and https options for a api.com when you add api2.com you would need to have ["http", "https", "http", "https"] in your schemes. On this basis, I would categorize this as a feature release.

Test Plan
1. Build a spec with an open api 2.0 config with two schemes and 1 host
2. confim that it builds and makes two servers available in swagger preview for the one host.
3. add a second host to the spec and confirm that the swagger spec build fails with an error stating that multiple hosts are not allowed in the 2.0 spec
4. switch to the 3.0 spec and build, confirm that the build is valid and previews correctly.
5. Add a 3rd record to the host list and confim that build fials, stating that host lists should match the length of the scheme list
6. Add a 3rd scheme and confirm that it builds correctly and previews

### All Submissions:

* [Ish... ] Have you followed the guidelines in our [Contributing](https://github.com/lukeautry/tsoa/tree/master/docs/CONTRIBUTING.md) document?
* [yes ] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/lukeautry/tsoa/pulls) for the same update/change?
* [yes ] Have you written unit tests?
* [yes ] Have you written unit tests that cover the negative cases (i.e.: if bad data is submitted, does the library respond properly)?
* [no ] This PR is associated with an existing issue?

**Closing issues**

**Test plan**

i wrote the test plan into the unit tests.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -65,6 +65,12 @@ export const validateSwaggerConfig = async (config: SwaggerConfig): Promise<Swag
   if (!config.outputDirectory) {
     throw new Error('Missing outputDirectory: configuration must contain output directory.');
   }
+  if (config.specVersion === 2 && config.host && typeof config.host !== 'string' && config.host.length > 1) {
+    throw new Error('Multiple hosts are only allowed in the 3.0 spec.');
+  }
+  if (config.specVersion === 3 && config.host && typeof config.host !== 'string' && config.schemes && config.host.length !== config.schemes.length) {
+    throw new Error('When hosts are an array,  they must match the length of the protocols. They are paired together');
+  }
   if (!config.entryFile) {
     throw new Error('Missing entryFile: Configuration must contain an entry point file.');
   }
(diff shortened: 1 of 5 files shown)


WoH:
I think I would be a lot more comfortable with pushing this to 3.0 without having to specify `["http", "https", "http", "https"] in your schemes`. I think it would be a lot more understandable if you could either 
1. Specify hosts as an array of strings where we pull out the schemes for swagger and just add all hosts for OAPI3 While removing host/scheme properties)
2. Consider host/scheme/basePath if specVersion is 2 or hosts of specVersion is 3 (probably the best option)

givethemheller:
Opened up a new PR on the 3.x branch. Didn't want to rebase across a major, so a new branch it was.

Issue #592, opened 2020-02-20
Improve Docs for monorepo setup


givethemheller:
In the process of using this package in a Lerna MonoRepo I believe I have a good use case for an external dependency mode. 

Note - I did read the MD file on why you made this choice. In the detailed description, I present a proposal to justify this use case. 

Side bar: Happy to do the work and contribute this. Love your package. Would like to treat this as a discussion.

## Sorting

* **I'm submitting a ...**
  - [ ] bug report
  - [X] feature request
  - [ ] support request

* I confirm that I
  - [x] used the [search](https://github.com/lukeautry/tsoa/search?type=Issues) to make sure that a similar issue hasn't already been submit

## Expected Behavior
When I set an "externalDependencies" CLI flag, TSOA crawls my node modules to build a complete type file for the composition that is generated through the dependency tree. As the developer/consumer, I manage the compile time cost of this independently.

## Current Behavior
I have to duplicate typings, by hand, through all levels of the types dependency tree. Further, I have to maintain those during upgrades.

## Possible Solution
Started looking at the TS API - but its new. 

### Use Case
I have recently started building a monorepo out to host a project that has grown to deserve it. I'm also breaking up my express servers into multiple smaller services. They still share the same data structure and to support that, I've created a shared resources library. It includes my models and a few things for authentication and various other tasks. 

### Discussion
You make two arguments against external dependencies. 

1. You don't want to slow down the compile time.
2. You don't want API interfaces changing on you unexpectedly. 

I propose that creating a flag and a corresponding spelunking method to fetch those types would make this a much more useful package. Creating Common libraries shared across packages is frequently done and in large enterprise applications. 

1: If I have control over it, I can be expeditious with how I use typings. I can actively look at the types I am importing and control this outcome manually. 

2. I think this runs somewhat counter intuitively to semantic versioning. These are all typescript projects being compiled, so if my dependency changes its interfaces without a breaking change version bump, someone has dropped the ball. More over, even if they fail to document a breaking change, anywhere that I am using that imported type, I am going to get compile time errors. 

I actually want all of my dependent packages to dictate my dependent interfaces and I want my builds to break when interfaces change on me unexpectedly.

A Corner Case: There is a use case where this is still problematic. If I pull an interface into a server and never operate on that interface programmatically, I wouldn't see a improperly versioned breaking change. What I would see though, is my swagger spec changing on generation before committing it to master.

[...]

PR #600, opened 2020-02-29
Updates Read me for mono repo usage with shared configs.

givethemheller:
Its just a docs update.

Bug
https://github.com/lukeautry/tsoa/issues/592#issuecomment-592961431

### DIFF
--- a/README.MD
+++ b/README.MD
@@ -1,4 +1,5 @@
 # tsoa
+
 Pronounced so·uh
 
 [![Codeship Status for lukeautry/tsoa](https://codeship.com/projects/cdce38d0-1f6b-0134-258e-1ed679ae6c9d/status?branch=master)](https://codeship.com/projects/160322)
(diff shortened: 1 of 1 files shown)


README.MD
```diff
@@ -1,57 +1,58 @@
 # tsoa
+
 Pronounced so·uh
 
 [![Codeship Status for lukeautry/tsoa](https://codeship.com/projects/cdce38d0-1f6b-0134-258e-1ed679ae6c9d/status?branch=master)](https://codeship.com/projects/160322)
 [![npm version](https://badge.fury.io/js/tsoa.svg)](https://badge.fury.io/js/tsoa)
 
 ## Table of Contents
 
-  * [Goal](#goal)
```
WoH: It seems like you have a formatter running on save, would you mind resubmitting this PR without the noise?

README.MD
```diff
@@ -103,9 +104,22 @@ export class UsersController extends Controller {
   }
 }
 ```
+
 Note: tsoa can not create swagger documents from interfaces that are defined in external dependencies. This is by design. Full explanation available in [ExternalInterfacesExplanation.MD](https://github.com/lukeautry/tsoa/blob/master/docs/ExternalInterfacesExplanation.MD)
```
WoH:
```suggestion
Note: tsoa can not create swagger documents from interfaces that are not part of your program defined in your tsConfig. This is by design. Full explanation available in [ExternalInterfacesExplanation.MD](https://github.com/lukeautry/tsoa/blob/master/docs/ExternalInterfacesExplanation.MD)
```

README.MD
```diff
@@ -103,9 +104,22 @@ export class UsersController extends Controller {
...
 
+### Mono Repos
```
WoH: This should be part of the External Interfaces Explanation.

README.MD
givethemheller: @WoH I'll clear the formatting and resubmit.

README.MD
givethemheller: @WoH should I update the external interfaces only or do you want information about this in the base readme as well?

WoH: Mostly the external interfaces, I added a small change to the base readme as a suggestion/inspiration because I think the language was not very clear.

Issue #510, opened 2019-10-11
Better handling for `true` or `false` literals in unions


dgreene1:
## Sorting

* **I'm submitting a ...**
  - [X] bug report
  - [ ] feature request
  - [ ] support request

* I confirm that I
  - [X] used the [search](https://github.com/lukeautry/tsoa/search?type=Issues) to make sure that a similar issue hasn't already been submit

## Expected Behavior
```ts
export type ComplicatedLiteralUnion = "hello" | false;
```

I would expect to get OpenAPI schema like:
```yaml
      oneOf:
        - type: string
          enum:
            - hello
        - type: boolean
          enum:
            - false
```

## Current Behavior
In the 2.0 branch it pretends that the boolean is a string
```yaml
      oneOf:
        - type: string
          enum:
            - hello
            - false
```

## Possible Solution
Have the code produce the expected result above.

## Steps to Reproduce
1. Create a tsoa route that returns this object (note: in the 2.0 branch you can't use an exported type so it has to be inline)
```ts
export interface ObjectThatHasComplicatedLiteral {
  complicatedLiteral: 'hello' | false
}
```

## Context (Environment)
Version of the library: 2.5.5
Version of NodeJS: LTS
* Confirm you were using yarn not npm: [X]

## Breaking change?
Yes, so I've put it in the 3.0 project

dgreene1:
Btw, the reason why I think that it should be two separate enums is due to this explanation in JSON Schema:
```
However, in most cases, the elements in the enum array should also be valid against the enclosing schema:

{
  "type": "string",
  "enum": ["red", "amber", "green", null]
}
This is in the enum, but it’s invalid against { "type": "string" }, so it’s ultimately invalid:
```
And yes, I recognize that OpenAPI is not JSONSchema, but it's a great example of the principle that the values in the enum must satisfy the type of the enum.

Read more here: https://json-schema.org/understanding-json-schema/reference/generic.html#enumerated-values

WoH:
My plan is to produce a type aware enum for each literal type and treat that like any other union for the next major.

dgreene1:
Cool, then you can use this issue number to track that.

dgreene1:
And more importantly, now you have acceptance criteria. Because neither of the following should happen:

```yaml
      oneOf:
        - type: string
          enum:
            - "hello"
            - false
```

```yaml
      oneOf:
        - type: string
          enum:
            - "hello"
            - "false"
```

Issue #497, opened 2019-10-03
generate swagger v2 without x-nullable


marknelissen:
#257 has been closed, stating that it is possible to mark a property as nullable by making use of the v3 spec. However, the original problem of #257 is the presence of x-nullable in v2 generated files.

This still causes the same problems for people having to use v2. As such, this addition should be removed or at least the option should be added to disable this behaviour through config.

## Sorting

* **I'm submitting a ...**
  - [ ] bug report
  - [x] feature request
  - [ ] support request

* I confirm that I
  - [x] used the [search](https://github.com/lukeautry/tsoa/search?type=Issues) to make sure that a similar issue hasn't already been submit

## Expected Behavior
No x-nullable in generated swagger.json.

## Current Behavior
Optional parameters are marked with x-nullable.

## Possible Solution
Remove the x-nullable generation or add config parameter to disable the generation.

## Steps to Reproduce

1. Use optional type in interface (eg. name?: string)
2. generate swagger.json

## Context (Environment)
Version of the library: 2.5.3
Version of NodeJS: v10.16.3
* Confirm you were using yarn not npm: [x]

dgreene1:
@marknelissen can you clarify how this is a bug? Swagger v2 allows the use of custom properties. See here: https://swagger.io/docs/specification/2-0/swagger-extensions/

What I'm saying is that (without me knowing more about the situation) I believe that `x-nullable` is valid schema. However, if you're encountering a problem due to the presence of a custom property, please describe at which point in the process it's causing an issue for you.

The more information we have the better we can assist. Thank you.

marknelissen:
Hi @dgreene1 , in general, this does not pose any problem. However, for my current project, the swagger.json needs to be submitted in order to publish the api in an api store. In order to do this, the API specification needs to meet the store standards. There are several rules (eg. no 200 response on POST), and among these is also a rule that trips regarding the presence of x-nullable in the generated swagger.json.

I opened this ticket, since it corresponds to the problem the OP of #257 was experiencing on Amazon, and the ticket never came to any real solution, since the OP ended up writing a module that reworked the generated json into something they could word with.

Since my problem could be fixed by not outputting x-nullable, I was interested in the proposed solution of #257 to disable the generation of x-nullable through a flag in the tsoa.json configuration. This would avoid me having to go through the file to remove all occurences, and be able to submit the generated swagger.json as is.

dgreene1:
[...]

Issue #479, opened 2019-09-22
Union with null is not supported


atomicus:
TSOA fails on generating routes when interface includes 'null' type.

* **I'm submitting a ...**
  - [x] bug report
  - [ ] feature request
  - [ ] support request

* I confirm that I
  - [x] used the [search](https://github.com/lukeautry/tsoa/search?type=Issues) to make sure that a similar issue hasn't already been submit

## Expected Behavior
Generate propper swagger for empty-valued patams

## Current Behavior
Error is presented:
```
Generate routes error.
 Error: Unknown type: NullKeyword
    at new GenerateMetadataError (/Users/atom/Projekty/atom/107/dcs-api/node_modules/tsoa/dist/metadataGeneration/exceptions.js:20:28)
    at TypeResolver.resolve (/Users/atom/Projekty/atom/107/dcs-api/node_modules/tsoa/dist/metadataGeneration/typeResolver.js:153:19)
    at /Users/atom/Projekty/atom/107/dcs-api/node_modules/tsoa/dist/metadataGeneration/typeResolver.js:567:152
    at Array.map (<anonymous>)
    at TypeResolver.getModelProperties (/Users/atom/Projekty/atom/107/dcs-api/node_modules/tsoa/dist/metadataGeneration/typeResolver.js:555:18)
    at TypeResolver.getReferenceType (/Users/atom/Projekty/atom/107/dcs-api/node_modules/tsoa/dist/metadataGeneration/typeResolver.js:352:35)
    at TypeResolver.resolve (/Users/atom/Projekty/atom/107/dcs-api/node_modules/tsoa/dist/metadataGeneration/typeResolver.js:202:30)
    at /Users/atom/Projekty/atom/107/dcs-api/node_modules/tsoa/dist/metadataGeneration/typeResolver.js:567:152
    at Array.map (<anonymous>)
    at TypeResolver.getModelProperties (/Users/atom/Projekty/atom/107/dcs-api/node_modules/tsoa/dist/metadataGeneration/typeResolver.js:555:18)
```

## Steps to Reproduce

1. Create interface in eg:
```
interface IServerSessionDetails {
   id: number,
   name: string,
   emptyVal: string | null
}
```
2.run routes generation

## Context (Environment)
Version of the library: 2.5.2
Version of NodeJS: 8.15 & 10.15
* Confirm you were using yarn not npm: [x]

dgreene1:
[...]

Issue #257, opened 2018-08-14
x-nullable is no Swagger


jpramondon:
Hey there,

We noticed a small change sneaked into the code recently:

https://github.com/lukeautry/tsoa/blob/f97c738b88c9af82dd09a83c1fb5399e40d5ca40/src/swagger/specGenerator.ts#L247

Is this an attempt to mark a model's property as required or mandatory ?
If so, then unfortunately this is a breaking change to anyone using AWS and its ApiGateway. Since _x-nullable_ is not a Swagger-valid property, the swagger.json file becomes invalid to anything trying to match it against the Swagger specification.

Additional input : The Swagger way to deal with required properties is described [here](https://swagger.io/docs/specification/data-models/data-types/).

For now, we'll just filter the _x-nullable_ properties out of the generated swagger file, but this point is worth a second look and discussion (before leaping to OpenAPI 3), which I'd like to have with you all if you're ok with this.

Cheers !

Jerome

lukeautry:
Hey Jerome. That is indeed a change I put in. I couldn't find anything official in OpenAPI docs pertaining to nullable properties, but it is indeed something that needs to be expressed in some way. It could be turned into a configurable option.

endor:
It could be worth removing this from the code and trying to push the OpenAPI 3 changes because there a valid `nullable` property exists.

jpramondon:
Thanks for your replies, guys.

Indeed the swagger specification has something for _required_ fields, but seeing you used the word _nullable_, I wasn't sure this was exactly what you meant initially.

This below is the swagger way of expressing a required field:
```
type: object
properties:
  id:
    type: integer
  username:
    type: string
  name:
    type: string
required:
  - id
  - username
```

Again, _nullable_ might mean something else.

Just like @endor states, OpenAPI 3 offers a _nullable_ qualifer, so it's really worth keeping it for later use (I've seen he's working on the OpenAPI3 compatible version).
Which brings up another question for this coming version: will we be able to control which specification we want the file to follow (Swagger or OpenAPI3) ?

And another one: is it still worth working on the swagger file generation now the OpenAPI3 version is on sight ? I'm asking so to know if I'm taking some time to implement that or if I just let it like it is right now (without support for required fields).
On one hand, we've been living without this feature until now, on the other that would improve the swagger specification coverage.

Tell me what you think.

endor:
[...]

PR #601, opened 2020-03-03
Add Null Keyword

### LINKED ISSUE #557: Optional properties should not be nullable by default
## Sorting

* **I'm submitting a ...**
  - [x] bug report
  - [ ] feature request
  - [ ] support request

* I confirm that I

## Expected Behavior
```json
"OptionalNotNullable": {
  "properties": {
      "optional": {
        "type": "string"
      },
    },
    "required": [],
    "type": "object",
    "additionalProperties": false
},
```

```ts
interface OptionalNotNullable {
  optional?: string
}
```

## Current Behavior
```json
"OptionalNotNullable": {
  "properties": {
      "optional": {
        "type": "string",
        "nullable": true
      },
    },
    "required": [],
    "type": "object",
    "additionalProperties": false
}
```

## Possible Solution
Don't mark [optional properties](https://github.com/lukeautry/tsoa/blob/master/src/swagger/specGenerator3.ts#L338) as `nullable: true`.

## Context (Environment)
Version of the library: 2.x
Version of NodeJS: any

- [x] Confirm you were using yarn not npm

## Detailed Description
If we have a representation of the concept (see propsal here: https://github.com/lukeautry/tsoa/issues/479#issuecomment-571134853) of `null` in OpenAPI spec, we should not mark any optional properties as `nullable: true` and instead use the explicit TS `null`.

## Breaking change?
Yes, should be 3.x

WoH:
### Goals

1. Support TypeScript's null keyword
2. Represent null as a "nullable" enum with type `[null]`
3. Distinguish between nullable (`null`) and required/optional (`undefined`)
4. Support Swagger whenever possible*

\* OpenAPI can use anyOf, to combine schemas. In case of Swagger, we need to check if we are combining a non-$ref type with `null`. In this case we can set `x-nullable: true`. Otherwise, we need to bail out since {$ref: '', ['x-nullable']: true} does not work/will be ignored (sibling values along $ref are ignored).

### All Submissions:

**Closing issues**

Closes #557, #479, #497 (presumably), #257 (presumably), #510

Breaks compat

### DIFF
--- a/src/metadataGeneration/tsoa.ts
+++ b/src/metadataGeneration/tsoa.ts
@@ -159,7 +159,7 @@ export namespace Tsoa {
    */
   export interface EnumType extends TypeBase {
     dataType: 'enum';
-    enums: Array<string | number | boolean>;
+    enums: Array<string | number | boolean | null>;
   }
 
   export interface ArrayType extends TypeBase {
--- a/src/metadataGeneration/typeResolver.ts
+++ b/src/metadataGeneration/typeResolver.ts
@@ -41,6 +41,14 @@ export class TypeResolver {
       return primitiveType;
     }
 
+    if (this.typeNode.kind === ts.SyntaxKind.NullKeyword) {
+      const enumType: Tsoa.EnumType = {
+        dataType: 'enum',
+        enums: [null],
+      };
+      return enumType;
+    }
+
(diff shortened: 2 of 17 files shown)


dgreene1:
Looks good to me! Great work! :)

WoH:
I'll try to add more tests around validation and integration, but I wanted to make sure we are all on board with the idea.

PR #602, opened 2020-03-03
convert path params with colon delimiter to bracelet delimiter during spec generation

itamarco:
### All Submissions:

**Closing issues**

**Test plan**

* I have wrote unit tests for the `convertColonPathParams` method in `pathUtils.ts`
* I have added route with colon delimiter path params to `paramaeterController.ts` and tested the resulted spec

### DIFF
--- a/src/utils/pathUtils.ts
+++ b/src/utils/pathUtils.ts
@@ -2,6 +2,7 @@
  * Removes all '/', '\', and spaces from the beginning and end of the path
  * Replaces all '/', '\', and spaces between sections of the path
  * Adds prefix and suffix if supplied
+ * Replace ':pathParam' with '{pathParam}'
  */
 export function normalisePath(path: string, withPrefix?: string, withSuffix?: string, skipPrefixAndSuffixIfEmpty: boolean = true) {
   if ((!path || path === '/') && skipPrefixAndSuffixIfEmpty) {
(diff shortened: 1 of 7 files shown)


src/utils/pathUtils.ts
```diff
@@ -18,3 +19,12 @@ export function normalisePath(path: string, withPrefix?: string, withSuffix?: st
   normalised = normalised.replace(/[/\\\s]+/g, '/');
   return normalised;
 }
+
+export function convertColonPathParams(path: string) {
```
WoH: minor nitpick: I think it's reasonable to do this inside `normalizePath`

itamarco: I agree and I put it there at first, but it made route generation tests fail.

src/utils/pathUtils.ts
```diff
@@ -18,3 +19,12 @@ export function normalisePath(path: string, withPrefix?: string, withSuffix?: st
...
+  if (!path || typeof path !== 'string') {
+    return path;
+  }
+
+  const normalised = path.replace(/:([^/]+)/g, '{$1}');
```
WoH:
```suggestion
  const normalised = path.replace(/\/:([^/]+)/g, '{$1}');
```

src/utils/pathUtils.ts
WoH: Yes, we are calling this again while building the templates. I assume this is the better approach then.

src/utils/pathUtils.ts
WoH:
```suggestion
  const normalised = path.replace(/:([^\/]+)/g, '{$1}');
```

PR #607, opened 2020-03-08
correctly propagate field name in validateModel

### LINKED ISSUE #605: fields in ValidateError incorrect for arrays containing refObjects
## Sorting

* **I'm submitting a ...**
  - [x] bug report
  - [ ] feature request
  - [ ] support request

* I confirm that I

## Expected Behavior
Field names returned inside a ValidateError object should be sufficient to locate field inside a type.

## Current Behavior
The errors are missing the array index.  

## Possible Solution
It looks like the validateModel function is different caller contexts which should dictate how the parent is sent down to the child.  I was able to fix the exact problem I was running into by checking if the current element name looked like an array index. I don't know if this is a good general solution however.  I was going to submit a PR and then saw the guidelines to create an issue first.  https://github.com/fantapop/tsoa/commit/e6a5727f3aaff090ab17b9174c7c113b89ec6d39

## Context (Environment)
Version of the library: 3.0.2
Version of NodeJS:  12.12.0

## Breaking change?
If people are relying on the incorrect paths then it would be broken for them after this fix.

fantapop:
Additionally stop adding a parent for body and body-prop at the getValidatedArgs entry point since it's now being added for each model ref.

### All Submissions:

**Closing issues**
closes #605 

None that I'm aware.  I do want to discuss some other changes but we can do that over in the issue.

**Test plan**

I added two tests to get a better understanding of how errors look when they're returned from `header` and `body-prop` type requests.  These were essentially just used to make sure i preserved the existing behavior.

I added one test to demonstrate that returning nested objects within array's did not give the expected field names within the error response.  I used this test provide faith that the new solution is working as intended.

I fixed numerous tests to use the correct field names for targeting the fields which are actually in error.

### DIFF
--- a/src/routeGeneration/templateHelpers.ts
+++ b/src/routeGeneration/templateHelpers.ts
@@ -641,12 +641,13 @@ export class ValidationService {
       }
 
       if (modelDefinition.dataType === 'refAlias') {
-        const parentName = modelDefinition.type.ref ? parent + name + '.' : parent;
-        return this.ValidateParam(modelDefinition.type, value, name, fieldErrors, parentName, swaggerConfig);
+        return this.ValidateParam(modelDefinition.type, value, name, fieldErrors, parent, swaggerConfig);
       }
 
+      const fieldPath = parent + name;
+
       if (!(value instanceof Object)) {
-        fieldErrors[parent + name] = {
+        fieldErrors[fieldPath] = {
           message: `invalid object`,
           value,
         };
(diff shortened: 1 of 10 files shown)


src/routeGeneration/templateHelpers.ts
```diff
@@ -641,12 +641,13 @@ export class ValidationService {
       }
 
       if (modelDefinition.dataType === 'refAlias') {
-        const parentName = modelDefinition.type.ref ? parent + name + '.' : parent;
-        return this.ValidateParam(modelDefinition.type, value, name, fieldErrors, parentName, swaggerConfig);
+        return this.ValidateParam(modelDefinition.type, value, name, fieldErrors, parent, swaggerConfig);
       }
 
+      const fieldPath = parent + name;
```
WoH:
```suggestion
      const modelDefinitionPath = parent + name;
```

Issue #372, opened 2019-07-24
Support for mapped types (including Omit)


manast-apsis:
Currently TSOA is not generating correct models for MappedTypes, and for some of them, like Omit it just giving an error:
```
There was a problem resolving type of 'Omit'.
There was a problem resolving type of 'MyModel'.
Generate swagger error.
 Error: Multiple matching models found for referenced type Omit; please make model names unique.
```

In particular Omit is really important when implementing Rest Apis. You may have a Model class, that represents all the fields part of the model, some may be mandatory other optional, however, in the REST api you will not support all these fields, for example, your model may have a "id" field, which is mandatory but it is autogenerated by the database when the model is stored. This is solved perfectly with the Omit type, unfortunately is not supported.
Would also be grateful for a workaround if you know any.

dgreene1:
@manast-apsis we'd be open to a PR for this if you'd be interested in enhancing tsoa to support this.

dgreene1:
I'll add this to the roadmap/project for version 3 of tsoa.

@fhewitt I'm cc'ing you hear so we can continue the conversation here that you started in #500

fhewitt:
Ha I missed this one, searched for "Utility" based on TypeScript doc and "Pick". I re-add the documentation link form my ticket: https://www.typescriptlang.org/docs/handbook/utility-types.html

Same point than @manast-apsis : all utility types would be great, but `Omit` and `Pick` are probably the most the commons in the context of REST API.

Searching for a workaround and investigating further, I've tested handmade MappedType:
```
export type PickTest<T, K extends keyof T> = {
    [P in K]: T[P];
};
```
It didn't work either. But does not raise any error at swagger/routes generation. The object that use PickTest is generated in swagger file but with empty properties.

That seems to be a worst issue because it fail only at run time.

fhewitt:
Also strongly linked with #469 which is a duplicate too.

lucas-rudd:
@fhewitt any update on this? Or, has any work already been started?
Support for Pick would be fantastic, it allows our code base to be a bit more maintainable to just Pick properties from an Interface rather than create an entire new type for request/responses of our REST APIs.

WoH:
There's a PR up #505 where you can track the progress. Please refrain from asking for ETAs but feel encouraged to take a look and submit your review if you're interested in seeing this go live.

manast-apsis:
The mentioned PR #505 I am afraid does currently not resolve this issue.

[...]

PR #608, opened 2020-03-09
Pick, Omit and Partial Support #372

givethemheller:
Addresses Issue
https://github.com/lukeautry/tsoa/issues/372

This PR adds utility type support for Pick, Partial and Omit.

Currently, this is working when the utility type is used in the body of the interface. My test cases are failing when the utility type is used in the declaration. 

e.g.
```
export interface Record {
   goodType: string
   redactType: string
}

// Fails tests
export interface RedactedRecord extends Pick<Record, "goodType"> {}

// passes tests 
export interface TestUtilities{
   // creates appropriate reference objects and adds them to the schemas and the generated 
   OpenApi spec.
   pickRedaction: Pick<Record, "goodType">
   partialTest: Partial<Record>
   omitTest: Omit<Record, "redactType">
}
```

If some comments can guide me in on where to make the changes for the declaration level use of utility type or point me at how to use the TS architecture to do this in a better way than what I am, please point.

Thanks.

### DIFF
--- a/src/metadataGeneration/typeResolver.ts
+++ b/src/metadataGeneration/typeResolver.ts
@@ -22,7 +22,7 @@ export class TypeResolver {
     private readonly parentNode?: ts.Node,
     private readonly extractEnum = true,
     private context: Context = {},
-  ) {}
+  ) { }
 
   public static clearCache() {
     Object.keys(localReferenceTypeCache).forEach(key => {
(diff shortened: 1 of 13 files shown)


WoH:
Is it fine with you to close this one and instead take a more general approach to resolving these as, what they are a the end of the day, type aliases + mapped/conditional types, which I implemented in 3.x?

givethemheller:
declining this - most implementation is in the 3.x branch. Will move onto query strings and header params

PR #611, opened 2020-03-11
allow quoting of patterns with "`" #563

### LINKED ISSUE #563: jsDoc cannot contain '@'
## Sorting

* **I'm submitting a ...**
  - [X] bug report
  - [ ] feature request
  - [ ] support request

* I confirm that I

## Expected Behavior

```ts
/**
 * @format email 
 * @pattern ^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\\.[a-zA-Z0-9-.]+$ 
 */
email: string
``` 
It should create a validator containing the full pattern :

```ts
"email": { 
  "dataType": "string", 
  "required": true, 
  "validators": { 
    "pattern": { 
      "value": "^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\\.[a-zA-Z0-9-.]+$"
    } 
  }
},
```

## Current Behavior

The pattern is split at @ and only the first part of the regex is used

```ts
"email": { 
  "dataType": "string", 
  "required": true, 
  "validators": { 
    "pattern": { 
      "value": "^[a-zA-Z0-9_.+-]+" 
    } 
  } 
},
```

## Context (Environment)
Version of the library: 2.5.12
Version of NodeJS: v10.15.3

jfrconley:
### All Submissions:

**Closing issues**

closes #563

**Test plan**
* Added test properties that specifically test regex that contain @ symbols.
* Bugfix solves issue with email property test where only half the regex was included.

### DIFF
--- a/src/utils/validatorUtils.ts
+++ b/src/utils/validatorUtils.ts
@@ -83,11 +83,11 @@ export function getParameterValidators(parameter: ts.ParameterDeclaration, param
           break;
         case 'pattern':
           if (typeof value !== 'string') {
-            throw new GenerateMetadataError(`${name} patameter use string.`);
+            throw new GenerateMetadataError(`${name} parameter use string.`);
           }
           validateObj[name] = {
             errorMsg: getErrorMsg(comment),
-            value,
+            value: removeSurroundingQuotes(value),
           };
           break;
         default:
@@ -173,11 +173,11 @@ export function getPropertyValidators(property: ts.PropertyDeclaration | ts.Type
           break;
         case 'pattern':
           if (typeof value !== 'string') {
-            throw new GenerateMetadataError(`${name} patameter use string.`);
+            throw new GenerateMetadataError(`${name} parameter use string.`);
           }
           validateObj[name] = {
             errorMsg: getErrorMsg(comment),
-            value,
+            value: removeSurroundingQuotes(value),
           };
           break;
         default:
@@ -219,4 +219,11 @@ function getParameterTagSupport() {
     'minDate',
     'maxDate',
   ];
+}
+
+function removeSurroundingQuotes(str: string) {
+  if (str.startsWith('`') && str.endsWith('`')) {
+    return str.substring(1, str.length - 1);
+  }
+  return str;
 }
--- a/src/utils/validatorUtils.ts
+++ b/src/utils/validatorUtils.ts
@@ -225,5 +225,8 @@ function removeSurroundingQuotes(str: string) {
   if (str.startsWith('`') && str.endsWith('`')) {
     return str.substring(1, str.length - 1);
   }
+  if (str.startsWith('```') && str.endsWith('```')) {
+    return str.substring(3, str.length - 3);
+  }
   return str;
 }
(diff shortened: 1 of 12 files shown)


tests/integration/openapi3-express.spec.ts
```diff
@@ -60,11 +62,6 @@ describe('OpenAPI3 Express Server', () => {
       word: 'word',
       fourtyTwo: 42,
       intersectionAlias: { value1: 'value1', value2: 'value2' },
-      intersectionAlias2: { value1: 'value1', value2: 'value2', value3: 'string' },
```
WoH: What's the rationale for removing these?

tests/integration/openapi3-express.spec.ts
```diff
@@ -148,14 +146,14 @@ describe('OpenAPI3 Express Server', () => {
     bodyModel.stringMax10Lenght = 'abcdefghijk';
     bodyModel.stringMin5Lenght = 'abcd';
     bodyModel.stringPatternAZaz = 'ab01234';
+    bodyModel.quotedStringPatternA = 'A';
 
     bodyModel.arrayMax5Item = [0, 1, 2, 3, 4, 6, 7, 8, 9];
     bodyModel.arrayMin2Item = [0];
     bodyModel.arrayUniqueItem = [0, 0, 1, 1];
-    bodyModel.intersection = { value1: '' } as any;
```
WoH: Why remove these?

tests/unit/swagger/definitionsGeneration/metadata.spec.ts
```diff
@@ -419,7 +419,7 @@ describe('Metadata generation', () => {
       expect(genderParam.parameterName).to.equal('gender');
       expect(genderParam.description).to.equal('Gender description');
       expect(genderParam.required).to.be.true;
-      expect(genderParam.type.dataType).to.equal('enum');
```
WoH: This should not change this, should it?

tests/integration/openapi3-express.spec.ts
jfrconley: Was not intended. I have added a commit fix unintended changes in this file.

tests/integration/openapi3-express.spec.ts
jfrconley: Was not intended. I have added a commit fix unintended changes in this file.

tests/unit/swagger/definitionsGeneration/metadata.spec.ts
jfrconley: Issue did not appear to be fixed on 3.x, and the change is so simple that I kept it in.

src/utils/validatorUtils.ts
```diff
@@ -220,3 +220,10 @@ function getParameterTagSupport() {
     'maxDate',
   ];
 }
+
+function removeSurroundingQuotes(str: string) {
+  if (str.startsWith('`') && str.endsWith('`')) {
```
WoH: I really like this approach. My only remaining question is whether we should unquote triple quotes aswell which seems to be a legal way of escaping, too (?)

jfrconley: If it's legal in JSDoc, we should probably support it as well. For pattern specifically it's not a big deal, but if this gets added for other tags it might be needed. I'll add it real quick

WoH:
Thank you very much!

PR #633, opened 2020-03-28
Heller/feature/oapi 3x hosts

givethemheller:
Previous PR here. https://github.com/lukeautry/tsoa/pull/595
Rebranched to avoid a rebase nightmare while moving to 3.x. 

The 3.x Spec was attempting to use the host, scheme to define the endpoints. The new definition in the 3.x spec is just plain old "hosts". E.g. a host is https://myapi.com

CLI validation has been setup to reject 2.x specs that include hosts, and reject 3.x specs that include host or scheme configs. 

It did take a bit of modification of the default configs and adding a spec toggle to the default generators to enforce OAPI 3.x configs. 

test plan = unit test modifications
its getting late...

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -56,7 +56,7 @@ export interface SwaggerConfig {
   /**
    * API hosts array, e.g. [https://locahost:3000, https://myapi.com]
    */
-  hosts?: string[];
+  servers?: OpenAPI3Server[];
 
   /**
    * API version number; defaults to npm package version
--- a/src/config.ts
+++ b/src/config.ts
@@ -49,8 +49,14 @@ export interface SwaggerConfig {
 
   /**
    * API host, expressTemplat e.g. localhost:3000 or myapi.com, use null for relative urls
+   * OAPI2 compliant
    */
   host?: string | null;
+
+  /**
+   * API hosts array, e.g. [https://locahost:3000, https://myapi.com]
+   */
+  hosts?: string[];
 
   /**
    * API version number; defaults to npm package version
--- a/src/swagger/specGenerator3.ts
+++ b/src/swagger/specGenerator3.ts
@@ -116,21 +116,15 @@ export class SpecGenerator3 extends SpecGenerator {
 
   private buildServers() {
     const basePath = normalisePath(this.config.basePath as string, '/', undefined, false);
-    const scheme = this.config.schemes ? this.config.schemes[0] : 'https';
 
-    let url;
-    if (this.config.host === null) {
-      url = basePath;
+    let urls;
+    if (!this.config.hosts || this.config.hosts.length === 0) {
+      urls = [`${basePath}`];
     } else {
-      const host = this.config.host || 'localhost:3000';
-      url = `${scheme}://${host}${basePath}`;
+      urls = this.config.hosts;
     }
-
-    return [
-      {
-        url,
-      } as Swagger.Server,
-    ];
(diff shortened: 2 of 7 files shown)


src/config.ts
```diff
@@ -49,9 +49,15 @@ export interface SwaggerConfig {
...
   /**
    * API host, expressTemplat e.g. localhost:3000 or myapi.com, use null for relative urls
+   * OAPI2 compliant
    */
   host?: string | null;
 
+  /**
+   * API hosts array, e.g. [https://locahost:3000, https://myapi.com]
+   */
+  hosts?: string[];
```
WoH:
```suggestion
  servers?: OpenAPI3Server[];
```
See https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#serverObject

tests/prepare.ts
```diff
@@ -63,7 +76,7 @@ const log = async <T>(label: string, fn: () => Promise<T>) => {
           middleware: 'express',
           routesDir: './tests/fixtures/express',
         },
-        defaultOptions,
+        defaultOptions(2),
```
WoH: Let's not call this over and over again. Assign the default options, assign `openApi3options` based on the `defaultOptions` and use these.

WoH:
[This may be helpful](https://github.com/metadevpro/openapi3-ts/blob/master/src/model/OpenApi.ts#L45)

src/swagger/specGenerator3.ts
```diff
@@ -116,21 +116,15 @@ export class SpecGenerator3 extends SpecGenerator {
...
   private buildServers() {
     const basePath = normalisePath(this.config.basePath as string, '/', undefined, false);
-    const scheme = this.config.schemes ? this.config.schemes[0] : 'https';
 
-    let url;
-    if (this.config.host === null) {
-      url = basePath;
+    let urls;
+    if (!this.config.hosts || this.config.hosts.length === 0) {
+      urls = [''];
```
endor, line 122: So we don't want to have `https://localhost:3000` as a default anymore?

src/swagger/specGenerator3.ts
```diff
@@ -116,21 +116,15 @@ export class SpecGenerator3 extends SpecGenerator {
...
-      url = `${scheme}://${host}${basePath}`;
+      urls = this.config.hosts;
     }
-
-    return [
-      {
-        url,
-      } as Swagger.Server,
-    ];
+    const servers = urls.map(url => ({ url: `${url}${basePath}` }));
```
endor, line 126: Hm.. what happens when `url` ends with a `/` and `basePath` starts with a `/`?

src/config.ts
```diff
@@ -49,9 +49,15 @@ export interface SwaggerConfig {
...
   /**
    * API host, expressTemplat e.g. localhost:3000 or myapi.com, use null for relative urls
+   * OAPI2 compliant
    */
   host?: string | null;
 
+  /**
+   * API hosts array, e.g. [https://locahost:3000, https://myapi.com]
+   */
+  servers?: OpenAPI3Server[];
```
endor, line 59: Should this be `hosts`?

WoH: No, should be servers since OAPI3 calls it servers and it's a difference interface.
Hosts don't exist any more, I don't want to support it either.

endor: Ah, ok, was confused by the comment above and the check for `this.config.hosts`.. Maybe I'm not reading this correctly.

PR #643, opened 2020-04-03
feat: ignore class methods

### LINKED ISSUE #631: feature: class methods and getter
I'm using tsoa with a typegoose model and I have some virtuals fields with getter and some methods used in backend. 
I want to be able to display my virtuals in my swagger definitions and ignore my methods without adding any jsdoc comment to ignore it.

## Sorting

* **I'm submitting a ...**
  - [ ] bug report
  - [x] feature request
  - [ ] support request

* I confirm that I

## Expected Behavior

```yaml
        Datasource:
            properties:
                name:
                    type: string
                pausedAt:
                    type: string
                    format: date-time
                isPaused:
                    type: boolean
            required:
                - name
                - pausedAt
            type: object
```

```ts
export class Datasource {

  @prop({
    trim: true,
    required: [true, 'NameEmpty']
  })
  public name!: string

  @prop({
    default: () => null
  })
  public pausedAt!: Date | null

  public get isPaused () {
    return this.pausedAt !== null
  }
}
```

## Current Behavior

An error occur with the getter or the method without an ignore jsdoc comment.

Eywek:
### All Submissions:

**Closing issues**

Closes #631 

The ignore part seems to works fine.

But for the toJSON scan I'm having trouble to implement it, I've found the way to get the return type of the toJSON methods but I haven't found a way to properly export that type to the swagger definitions. So if you have any advice / ideas  I'd be glad to have them

### DIFF
--- a/src/metadataGeneration/typeResolver.ts
+++ b/src/metadataGeneration/typeResolver.ts
@@ -459,18 +459,19 @@ export class TypeResolver {
     const description = this.getNodeDescription(modelType);
 
     // Handle toJSON methods
-    let toJSON: ts.ClassElement | ts.TypeElement | undefined;
-    if (ts.isClassDeclaration(modelType)) {
-      toJSON = modelType.members.find(member => member.name && member.name.getText() === 'toJSON');
-    } else {
-      toJSON = modelType.members.find(member => member.name && member.name.getText() === 'toJSON');
+    let toJSON: ts.Symbol | undefined;
+    if ('symbol' in modelType) {
+      // @ts-ignore
+      const type = this.current.typeChecker.getDeclaredTypeOfSymbol(modelType.symbol);
+      toJSON = this.current.typeChecker.getPropertyOfType(type, 'toJSON');
     }
 
-    if (toJSON && (ts.isMethodDeclaration(toJSON) || ts.isMethodSignature(toJSON))) {
-      let nodeType = toJSON.type;
+    if (toJSON && toJSON.declarations.length > 0) {
+      const declaration = toJSON.declarations[0] as ts.MethodDeclaration;
+      let nodeType = declaration.type;
       if (!nodeType) {
         const typeChecker = this.current.typeChecker;
-        const signature = typeChecker.getSignatureFromDeclaration(toJSON);
+        const signature = typeChecker.getSignatureFromDeclaration(declaration);
         const implicitType = typeChecker.getReturnTypeOfSignature(signature!);
         nodeType = typeChecker.typeToTypeNode(implicitType) as ts.TypeNode;
       }
--- a/src/metadataGeneration/typeResolver.ts
+++ b/src/metadataGeneration/typeResolver.ts
@@ -144,7 +144,7 @@ export class TypeResolver {
     if (ts.isMappedTypeNode(this.typeNode) && this.referencer) {
       const isNotIgnored = (e: ts.Declaration) => {
         const ignore = isExistJSDocTag(e, tag => tag.tagName.text === 'ignore');
-        return !ignore;
+        return !ignore && (ts.isPropertyDeclaration(e) || ts.isPropertySignature(e) || ts.isParameter(e));
       };
 
       const type = this.current.typeChecker.getTypeFromTypeNode(this.referencer);
--- a/src/metadataGeneration/typeResolver.ts
+++ b/src/metadataGeneration/typeResolver.ts
@@ -254,6 +254,22 @@ export class TypeResolver {
         return stringMetaType;
       }
 
+      const type = this.current.typeChecker.getTypeFromTypeNode(this.typeNode);
+      const toJSON = this.current.typeChecker.getPropertyOfType(type, 'toJSON');
+      if (toJSON) {
+        const declaration = toJSON.declarations[0] as ts.MethodDeclaration;
+        const signature = this.current.typeChecker.getSignatureFromDeclaration(declaration);
+        const returnType = this.current.typeChecker.getReturnTypeOfSignature(signature!);
+        try {
+          const resolved = new TypeResolver(this.current.typeChecker.typeToTypeNode(returnType)!, this.current, this.parentNode, this.context).resolve() as Tsoa.ReferenceType;
+          resolved.refName = typeReference.getText();
+          this.current.AddReferenceType(resolved);
+          return resolved;
+        } catch {
+          // ignore methods
+        }
+      }
+
       if (this.context[typeReference.typeName.text]) {
         return new TypeResolver(this.context[typeReference.typeName.text], this.current, this.parentNode, this.context).resolve();
       }
--- a/src/metadataGeneration/typeResolver.ts
+++ b/src/metadataGeneration/typeResolver.ts
@@ -254,29 +254,6 @@ export class TypeResolver {
         return stringMetaType;
       }
 
-      const type = this.current.typeChecker.getTypeFromTypeNode(this.typeNode);
-      const toJSON = this.current.typeChecker.getPropertyOfType(type, 'toJSON');
-      if (toJSON) {
-        const declaration = toJSON.declarations[0] as ts.MethodDeclaration;
-
-        let nodeType = declaration.type;
-        if (!nodeType) {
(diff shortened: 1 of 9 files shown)


WoH:
You should be able to filer the Declaration[] to find the toJSON declaration, which should have a typeNode. Instead of building the metadata for the class, return a reference to the thpeNode. If the typeNode does not exist, fall back to the typeNode, the type checker infers. You can have a look at the MethodGenerator for a Controller, which essentially does the same thing.

Eywek:
Thank you.
Tests are failing because of a missing format for a `DateAlias` but I don't know how to properly handle that case.

WoH:
I think we handle this in a special case in the TypeResolver usually. I'll take a look

src/metadataGeneration/typeResolver.ts
```diff
@@ -254,6 +254,29 @@ export class TypeResolver {
         return stringMetaType;
       }
 
+      const type = this.current.typeChecker.getTypeFromTypeNode(this.typeNode);
+      const toJSON = this.current.typeChecker.getPropertyOfType(type, 'toJSON');
```
WoH, line 258: We should not use the typeChecker by default. Best way is always to recursively call the TR on the typeNode and only fall back to the inferred type if there's no explicit typeNode.

src/metadataGeneration/typeResolver.ts
```diff
@@ -254,6 +254,29 @@ export class TypeResolver {
...
+      if (toJSON) {
+        const declaration = toJSON.declarations[0] as ts.MethodDeclaration;
+
+        let nodeType = declaration.type;
+        if (!nodeType) {
+          const typeChecker = this.current.typeChecker;
+          const signature = typeChecker.getSignatureFromDeclaration(declaration);
```
WoH, line 265: We should not use the typeChecker by default. Best way is always to recursively call the TR on the typeNode and only fall back to the inferred type if there's no explicit typeNode.

Eywek: I don't understand, what do you mean by `TR`? And you're saying we're using the typeChecker by default but we only do it if `declaration.type` is undefined

WoH: You're right, sorry about that, I missed that.

Eywek: So can I resolve this?

src/metadataGeneration/typeResolver.ts
```diff
@@ -254,6 +254,29 @@ export class TypeResolver {
         return stringMetaType;
       }
 
+      const type = this.current.typeChecker.getTypeFromTypeNode(this.typeNode);
```
WoH, line 257: I didn't see this initially, resolving based on toJSON() only makes sense when we are at the point of resolving a Model right around [here](https://github.com/lukeautry/tsoa/blob/e151d2ec01b741c7bc0952aeb71b4025f9278ef2/src/metadataGeneration/typeResolver.ts#L480)

src/metadataGeneration/typeResolver.ts
```diff
@@ -455,15 +455,45 @@ export class TypeResolver {
   }
 
   private getModelReference(modelType: ts.InterfaceDeclaration | ts.ClassDeclaration, name: string) {
+    const example = this.getNodeExample(modelType);
+    const description = this.getNodeDescription(modelType);
+
+    // Handle toJSON methods
+    let toJSON: ts.ClassElement | ts.TypeElement | undefined;
+    if (ts.isClassDeclaration(modelType)) {
+      toJSON = modelType.members.find(member => member.name && member.name.getText() === 'toJSON');
```
WoH, line 464: This does not consider the prototype chain, right? Can we ask the TypeChecker for a declaration here?

Eywek: I don't understand what you're talking about sorry

WoH: Essentially I'm saying that the `toJson()` method may exist on a class we inherit from. It may make sense to ask the type checker instead of iterating only the class declaration's members.

Eywek: Hmm, right, I'm gonna check that

Eywek: So, I've found that I can use `modelType.symbol` to get the type and then use `getPropertyOfType` to handle inheritance
But `modelType.symbol` isn't defined in types, maybe you know why? I'm not really proud of using `@ts-ignore`

WoH: I'll try to find time to debug so I can hopefully see if there's anything major to improve here, looks pretty solid now at first glance, I have some vague ideas but no way to tell if it'll work without actually spending time on this.

src/metadataGeneration/typeResolver.ts
```diff
@@ -455,15 +455,46 @@ export class TypeResolver {
...
+
+    if (toJSON && toJSON.declarations.length > 0) {
+      const declaration = toJSON.declarations[0] as ts.MethodDeclaration;
+      let nodeType = declaration.type;
+      if (!nodeType) {
+        const typeChecker = this.current.typeChecker;
+        const signature = typeChecker.getSignatureFromDeclaration(declaration);
+        const implicitType = typeChecker.getReturnTypeOfSignature(signature!);
+        nodeType = typeChecker.typeToTypeNode(implicitType) as ts.TypeNode;
+      }
```
WoH, line 477: We could probably handle anonymous classes but it certainly won't be a blocker:
```suggestion
    if (!modelType.name) {
      throw new GenerateMetadataError("Can't get Symbol from anonymous class", modelType);
    }

    const type = this.current.typeChecker.getTypeAtLocation(modelType.name);
    const toJSON = this.current.typeChecker.getPropertyOfType(type, 'toJSON');

    if (toJSON && toJSON.valueDeclaration && (ts.isMethodDeclaration(toJSON.valueDeclaration) || ts.isMethodSignature(toJSON.valueDeclaration))) {
      let nodeType = toJSON.valueDeclaration.type;
      if (!nodeType) {
        const signature = this.current.typeChecker.getSignatureFromDeclaration(toJSON.valueDeclaration);
        const implicitType = this.current.typeChecker.getReturnTypeOfSignature(signature!);
        nodeType = this.current.typeChecker.typeToTypeNode(implicitType) as ts.TypeNode;
      }
```

WoH: This is my best approach that's reasonably safe: We can get the TypeAtLocation with the identifier and use that to handle toJSON methods iv available.

Eywek: Great thank you! I'll test and commit this tomorrow or monday.
No problems!

Eywek:
It should be good @WoH

WoH:
> It should be good @WoH

Would you mind adding an integration test that ensures that (without explicitly calling `toJSON()`) returned classes are serialized properly?

Code lgtm

Eywek:
What about this one https://github.com/lukeautry/tsoa/pull/643/files#diff-d42372cdc5a5c6725a6e038616254163R62, https://github.com/lukeautry/tsoa/pull/643/files#diff-19811e20ee8ebdc5405ac00d6cd74d94R813 ?

WoH:
> What about this one https://github.com/lukeautry/tsoa/pull/643/files#diff-d42372cdc5a5c6725a6e038616254163R62, https://github.com/lukeautry/tsoa/pull/643/files#diff-19811e20ee8ebdc5405ac00d6cd74d94R813 ?

These make sure the spec gets generated correctly, the integration tests call a tsoa server and make sure the returned data matches the specification (at runtime)

Eywek:
Hm right, there is no checkbox because the fork is on an organization, but maybe you can commit without push and provide me the patch file of your commit. Otherwise I will try to add integrations tests on my own at the end of the day.

WoH:
Here's the patch: https://gist.githubusercontent.com/WoH/9f6618aaaaf0360941823d1625ddac70/raw/7820cec3e28ec339ff81d46ac8933146930c5dba/integration.diff

Eywek:
Applied thank you

PR #654, opened 2020-04-08
deps: add glob

Eywek:
Glob is used but not in `package.json` 
https://github.com/lukeautry/tsoa/blob/3.x/src/utils/importClassesFromDirectories.ts#L9


WoH:
Thank you, good catch.

PR #661, opened 2020-04-17
ValidateError should extend Error

### LINKED ISSUE #658: ValidateError should extend Error
## Sorting

* **I'm submitting a ...**
  - [x] bug report
  - [ ] feature request
  - [ ] support request

* I confirm that I

## Expected Behavior
- `ValidateError` should extend `Error`

## Current Behavior
- `ValidateError` does not extend `Error`

As a result, tools like [Sentry](https://sentry.io/) do not fully work with `ValidateError`. Sentry makes a [synthetic exception](https://docs.sentry.io/platforms/node/#hints-for-events) and does not include it's full exception info, it only reports:
```
Non-Error exception captured with keys: fields, message, name, status
```

## Possible Solution
- update `ValidateError` to extend `Error`

## Steps to Reproduce
1. Generate an express server
2. Make an API request that triggers a validation error, for example: https://github.com/lukeautry/tsoa/blob/313012a1214f0afa95756cf23bbc13f46277499e/tests/integration/express-server.spec.ts#L463-L470
3. `ValidateError` should be thrown, and it should extend `Error`

## Context (Environment)
Version of the library: 2.5.13
Version of NodeJS: 10.16.3
* Confirm you were using yarn not npm - I use npm, not yarn

## Detailed Description
The simplest non-breaking change would be to update this line: https://github.com/lukeautry/tsoa/blob/533fe059cc8373e31e8ab6d7698c18ae3ea3b709/src/routeGeneration/templateHelpers.ts#L746
to this:
```ts
export class ValidateError extends Error implements Exception {
```

## Breaking change?
- Yes

aldenquimby:
### All Submissions:

**Closing issues**
Closes #658

- not ideal to have 2 separate validation error classes
- when I initially updated `ValidateError` to extend `Error`, all server tests passed except hapi, which had this message:
```
  1) Hapi Server
       returns error if missing required query parameter:
     Error: the object {
  "error": "Error: expected 400 \"Bad Request\", got 500 \"Internal Server Error\""
  "response": "Error: cannot GET /v1/GetTest/1/true/test?booleanParam=true&stringParam=test1234 (500)"
} was thrown, throw an Error :)
      at process._tickCallback (internal/process/next_tick.js:68:7)
```
- and the corresponding server side error was:
```
Debug: internal, implementation, error 
    Error: Cannot wrap an error
    at new module.exports (/Users/alden/Desktop/alden-tsoa/tsoa/node_modules/@hapi/hoek/lib/error.js:23:19)
    at Object.module.exports [as assert] (/Users/alden/Desktop/alden-tsoa/tsoa/node_modules/@hapi/hoek/lib/assert.js:20:11)
    at internals.Toolkit.response (/Users/alden/Desktop/alden-tsoa/tsoa/node_modules/@hapi/hapi/lib/toolkit.js:174:14)
    at handler (/Users/alden/Desktop/alden-tsoa/tsoa/tests/fixtures/hapi/routes.ts:1091:14)
    at module.exports.internals.Manager.execute (/Users/alden/Desktop/alden-tsoa/tsoa/node_modules/@hapi/hapi/lib/toolkit.js:41:33)
    at Object.internals.handler (/Users/alden/Desktop/alden-tsoa/tsoa/node_modules/@hapi/hapi/lib/handler.js:46:48)
    at exports.execute (/Users/alden/Desktop/alden-tsoa/tsoa/node_modules/@hapi/hapi/lib/handler.js:31:36)
    at Request._lifecycle (/Users/alden/Desktop/alden-tsoa/tsoa/node_modules/@hapi/hapi/lib/request.js:312:68)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```
- which comes from the template here: https://github.com/lukeautry/tsoa/blob/82d02735eacda1842cef2d9587df46c93483a72b/src/routeGeneration/templates/hapi.hbs#L81
- very open to better ideas for fixing this issue if you have them!

**Test plan**
- unit tests assert that ValidateError extends Error, which was the primary goal

### DIFF
--- a/src/routeGeneration/templates/hapi.hbs
+++ b/src/routeGeneration/templates/hapi.hbs
@@ -2,9 +2,9 @@
 /* eslint-disable */
 // WARNING: This file was auto-generated with tsoa. Please do not modify it. Re-run tsoa to re-generate this file: https://github.com/lukeautry/tsoa
 {{#if canImportByAlias}}
-  import { Controller, ValidationService, FieldErrors, ValidateError, TsoaRoute } from 'tsoa';
+  import { Controller, ValidationService, FieldErrors, ValidateException, TsoaRoute } from 'tsoa';
 {{else}}
-  import { Controller, ValidationService, FieldErrors, ValidateError, TsoaRoute } from '../../../src';
+  import { Controller, ValidationService, FieldErrors, ValidateException, TsoaRoute } from '../../../src';
 {{/if}}
 {{#if iocModule}}
 import { iocContainer } from '{{iocModule}}';
@@ -204,7 +204,7 @@ export function RegisterRoutes(server: any) {
             }
         });
         if (Object.keys(errorFields).length > 0) {
-            throw new ValidateError(errorFields, '');
+            throw new ValidateException(errorFields, '');
         }
         return values;
     }
--- a/src/routeGeneration/templates/hapi.hbs
+++ b/src/routeGeneration/templates/hapi.hbs
@@ -2,9 +2,9 @@
 /* eslint-disable */
 // WARNING: This file was auto-generated with tsoa. Please do not modify it. Re-run tsoa to re-generate this file: https://github.com/lukeautry/tsoa
 {{#if canImportByAlias}}
-  import { Controller, ValidationService, FieldErrors, ValidateException, TsoaRoute } from 'tsoa';
+  import { Controller, ValidationService, FieldErrors, ValidateError, TsoaRoute } from 'tsoa';
 {{else}}
-  import { Controller, ValidationService, FieldErrors, ValidateException, TsoaRoute } from '../../../src';
+  import { Controller, ValidationService, FieldErrors, ValidateError, TsoaRoute } from '../../../src';
 {{/if}}
 {{#if iocModule}}
 import { iocContainer } from '{{iocModule}}';
@@ -16,6 +16,7 @@ import { {{name}} } from '{{modulePath}}';
 {{#if authenticationModule}}
 import { hapiAuthentication } from '{{authenticationModule}}';
 {{/if}}
+import { boomify, Payload } from '@hapi/boom';
 
 // WARNING: This file was auto-generated with tsoa. Please do not modify it. Re-run tsoa to re-generate this file: https://github.com/lukeautry/tsoa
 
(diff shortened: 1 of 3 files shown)


WoH:
I think we should just throw the Error and not catch it so hapi can take over. 
Maybe we should [`boomify`](https://hapi.dev/module/boom/api/?v=9.1.0#boomifyerr-options) the error though. 
Not using hapi right now, so any input from someone who's using hapi more regularly would be appreciated.

aldenquimby:
@WoH good idea. I've never used hapi either, but reading the docs it seems we should use boomify. Letting the error bubble up would end up throwing a 500 according to [these docs](https://github.com/hapijs/hapi/blob/master/API.md#errors), which we don't want.

I'll give boomify a try on this PR. Thank you for the review!

src/routeGeneration/templates/hapi.hbs
```diff
@@ -77,9 +78,14 @@ export function RegisterRoutes(server: any) {
...
-                        return h
-                            .response(err)
-                            .code(err.status || 500);
+                        const boomErr = boomify(err);
+                        boomErr.output.statusCode = err.status || 500;
+                        boomErr.output.payload = {
+                            name: err.name,
+                            fields: err.fields,
+                            message: err.message,
+                        } as unknown as Payload;
```
aldenquimby, line 87: The `@hapi/boom` [Payload type](https://github.com/hapijs/boom/blob/7e94679c2111ee49f88a787499a36aedba207c83/lib/index.d.ts#L93-L108) is incorrect, which is why we need to force cast here.

The [docs here](https://github.com/hapijs/hapi/blob/fa76016359245c9f46a2332fd8bab82ae3b5bba9/API.md#error-transformation) say you can make payload anything you want, so the typescript type should have all optional `?` keys and should allow extra keys

aldenquimby:
@WoH I just added a working version of `boomify`. Is anyone on the tsoa team a hapi user? Would be great to get eyes on here from someone that knows hapi

src/routeGeneration/templates/hapi.hbs
WoH: Can you try to set this during the initial call to `boomify()`: `boomify(err, {decorate: {fields, message}})`

aldenquimby: Yes I gave that a try and unfortunately it doesn't work. From the docs:
```
decorate - an option with extra properties to set on the error object.
```

So `decorate` adds fields to the JS error object, but those fields don't end up in the payload response from the server

WoH:
I think the integration tests for hapi give me sufficient confidence that this will work as is.

PR #662, opened 2020-04-19
feat: handle `Record<>` mapped type

### LINKED ISSUE #653: mapped type: `Record<'foo', string>`
I'm trying to use `Record<'id', string>;`.
I've seen https://github.com/lukeautry/tsoa/pull/505 and it seems that it's known that `Record<>` aren't supported, but I wanted to have some update on this.

## Sorting

* **I'm submitting a ...**
  - [ ] bug report
  - [x] feature request
  - [ ] support request

* I confirm that I

## Expected Behavior

```ts
export interface TestModel {
  recordType?: Record<'id', string>;
}
```

```
There was a problem resolving type of 'Record<'id', string>'.
There was a problem resolving type of 'TestModel'.

/private/tmp/tsoa/src/metadataGeneration/typeResolver.ts:151
      const tsProperties = this.current.typeChecker.getPropertiesOfType(type).map(symbol => {
                                                                              ^
TypeError: Cannot read property '0' of undefined
    at /private/tmp/tsoa/src/metadataGeneration/typeResolver.ts:151
```

Thank you

Eywek:
### All Submissions:

**Closing issues**

Closes #653

Some code is duplicated with jsdoc tags, but I haven't found a better way to handle this.

### DIFF
--- a/src/metadataGeneration/typeResolver.ts
+++ b/src/metadataGeneration/typeResolver.ts
@@ -143,69 +143,66 @@ export class TypeResolver {
 
     if (ts.isMappedTypeNode(this.typeNode) && this.referencer) {
       const type = this.current.typeChecker.getTypeFromTypeNode(this.referencer);
+      const mappedTypeNode = this.typeNode;
       const typeChecker = this.current.typeChecker;
-      const properties: Tsoa.Property[] = [];
-      for (const property of type.getProperties()) {
-        const propertyType = typeChecker.getTypeOfSymbolAtLocation(property, this.typeNode);
-        const jsDocTags = property.getJsDocTags();
-
+      const getDeclaration = (prop: ts.Symbol) => prop.declarations && (prop.declarations[0] as ts.Declaration | undefined);
+      const isIgnored = (prop: ts.Symbol) => {
+        const declaration = getDeclaration(prop);
+        return (
+          prop.getJsDocTags().find(tag => tag.name === 'ignore') !== undefined ||
+          (declaration !== undefined && !ts.isPropertyDeclaration(declaration) && !ts.isPropertySignature(declaration) && !ts.isParameter(declaration))
+        );
+      };
+      const properties: Tsoa.Property[] = type
+        .getProperties()
         // Ignore methods, getter, setter and @ignored props
-        const declaration = property.declarations && property.declarations[0] as ts.Declaration | undefined;
-        const haveIgnoreTag = jsDocTags.find(tag => tag.name === 'ignore');
-        if (
-          haveIgnoreTag ||
-          (declaration !== undefined &&
-          !ts.isPropertyDeclaration(declaration) &&
-          !ts.isPropertySignature(declaration) &&
-          !ts.isParameter(declaration))
-        ) {
-          continue // ignore
-        }
+        .filter(property => isIgnored(property) === false)
+        // Transform to property
+        .map(property => {
+          const propertyType = typeChecker.getTypeOfSymbolAtLocation(property, this.typeNode);
+          const jsDocTags = property.getJsDocTags();
+          const declaration = getDeclaration(property) as ts.PropertySignature | ts.PropertyDeclaration | ts.ParameterDeclaration | undefined;
 
-        // Resolve default value, required and typeNode
-        const defaultTag = jsDocTags.find(tag => tag.name === 'default');
-        let defaultValue = defaultTag !== undefined ? defaultTag.text : undefined;
-        let required = false;
-        let typeNode = this.current.typeChecker.typeToTypeNode(propertyType)!;
-        if (declaration) {
-          required = !declaration.questionToken && !declaration.initializer; // no `?` and no default value
-          defaultValue = getInitializerValue(declaration.initializer); // handle `public foo = 'default'`
-          if (declaration.type) {
-            // preserve aliases
-            typeNode = declaration.type;
+          if (declaration && ts.isPropertySignature(declaration)) {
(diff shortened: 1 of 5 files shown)


Eywek:
@WoH did you have time to check this one?

WoH:
Not yet, I'll try to take a look over the weekend.
I'm worried about Type Alias compatibility, maybe I can come up with a way to ensure there are no unintended side effects here while reusing the existing jsDoc parsing helpers if possible, which is why I'd like to ask for some more time to review this.

src/metadataGeneration/typeResolver.ts
```diff
@@ -142,23 +142,70 @@ export class TypeResolver {
...
-      const properties = tsProperties.filter(isNotIgnored).map(property => {
-        if (ts.isPropertySignature(property)) {
-          return this.propertyFromSignature(property, mappedTypeNode.questionToken);
-        } else if (ts.isPropertyDeclaration(property) || ts.isParameter(property)) {
-          return this.propertyFromDeclaration(property, mappedTypeNode.questionToken);
-        } else {
-          throw new GenerateMetadataError(`could not resolve property of kind ${property.kind}. Please report this as an issue`, property);
+      const typeChecker = this.current.typeChecker;
+      const properties: Tsoa.Property[] = [];
+      for (const property of type.getProperties()) {
```
WoH, line 148: We should already map to the valuedeclaration and filter here.

WoH:
First observations: 
- `Record<string, T>` won't set `additionalProperties`
- `Record<'name', Alias> won't ref to the alias from the property but inline and lose info.

You're removing calls to `getPropertyFromSignature` and `getPropertyFromDeclaration` and duplicate a lot of work. If these methods are not sufficient to resolve, we should improve these.

Eywek:
> You're removing calls to propertyFromSignature and propertyFromDeclaration and duplicate a lot of work. If these methods are not sufficient to resolve, we should improve these.

I've tried to remove some duplication and use these methods. If you have some advice to remove other duplications I'm listening
I'll look for broken cases

WoH:
> EDIT: I don't understand what you mean by Record<string, T>

A `Record<string, T>` is equivalent to `{[name:string]: T}`.
Example: `Record<string, number>` is like `{[name: string]: number}`

Eywek:
Hm right I see. I don't really know how to properly handle that case and I don't know if `Record<>` should be used for this type of things anyway.
If you have any idea of how we can implement it I will try but otherwise where do you want to document it? In the readme or in a new `docs/` file?

WoH:
> If you have any idea of how we can implement it I will try but otherwise where do you want to document it? In the readme or in a new docs/ file?

I'm working on new docs for the 3.0 stable, if I don't remember to document this, just remind me again.
No action needed from your side here.

I'm trying to understand why you need to reimplement the jsDoc parsing. Is the virtual node creating issues with the existing jsdoc parsing methods?

Eywek:
> I'm trying to understand why you need to reimplement the jsDoc parsing. Is the virtual node creating issues with the existing jsdoc parsing methods (getNodeFormat, getNodeDescription, getNodeExample, getJSDocComment(node, 'default'))?

For example, when I'm trying to use `isExistJSDocTag(typeNode, tag => tag.tagName.text === 'ignore')` to ignore, it doesn't work on the `hidden` property (on the `ErrorResponseModel` interface which is used with `Omit<>` later). Furthermore, I can't use `getNodeDescription`, `getNodeExample`, `getNodeFormat` and `getJSDocComment` because I don't have a `node`, [here](https://github.com/lukeautry/tsoa/pull/662/files#diff-04ed568b4a647b0bfd718a5afdb822fbR194) `declaration` (which is called `node` in your methods) is undefined.

WoH:
Ok, thanks for clarifying. I see that there is is are `tags` present on the `property`-Symbol, but I don't see how we may use that to set JsDoc annotations. This path only gets triggered with 'record-foo' and 'record-bar' right now, potentially also with some indexed type (?). In the first case there should not be any useful tags present, in the latter, they may potentially even be counter intuitive.

Trying to get some JSDoc from the `propertyType` seems hard, so maybe we shouldn't set these at all for now.

Eywek:
So you want to remove https://github.com/lukeautry/tsoa/pull/662/files#diff-04ed568b4a647b0bfd718a5afdb822fbR183-R192 and https://github.com/lukeautry/tsoa/pull/662/files#diff-04ed568b4a647b0bfd718a5afdb822fbR173-R174 right?

I haven't found a way to test this code anyway so delete it could be a good move

WoH:
Correct

Eywek:
It's removed, about the description part, currently it resolves the description of the `Record<>` type. I don't know if it's wanted or not

WoH:
> It's removed, about the description part, currently it resolves the description of the `Record<>` type. I don't know if it's wanted or not

```js
               description: 'Construct a type with a set of properties K of type T',
```

should come from the type alias description which is the of the property description, right?
I'm talking about the 2x `description: undefined` as the property descriptions.

Eywek:
Hm, right

WoH:
I wouldn't be sure where/when/how I'd want to set that, if there is a way to actually do that which makes sense, I'd be open, but currently, I don't think we actually need that (at least I can't tell)

Eywek:
It should be good

WoH:
LGTM, thank you so much for contributing and your patience!

Eywek:
When do you plan to release?

WoH:
Hope I'll be ready until tomorrow evening if that sounds good.

PR #666, opened 2020-04-21
Add support for @Hidden() decorator on Query params

### LINKED ISSUE #582: Add support for @Hidden() decorator on Query Parameters
## Sorting

* **I'm submitting a ...**
  - [ ] bug report
  - [X] feature request
  - [ ] support request

* I confirm that I

## Expected Behavior

A developer should be able to hide a specific query parameter that they may not want the general public to know about but still need from their UI. Example:

The generated swagger would not show anything for `superSecretFlag`

```ts
export class SampleController {
  @Get('id')
  public async find(@Path('id') id:string, @Hidden() superSecretFlag: boolean = false): Promise<any> {  }
}
```

## Current Behavior

Adding Hidden next to a parameter does not seem to do anything

## Context (Environment)
Version of the library: 2.3.8
Version of NodeJS: 10.16.3

aldenquimby:
### All Submissions:

**Closing issues**
Closes #582
Relates to #310 

- Why this approach? Went with `@Hidden` because it's consistent with tsoa controllers and methods, and was the agreed approach from discussion on #582 
- Alternative 1: add a `hidden` property to `@Query`. This would align with how Swagger's Java annotations work `@ApiParam(hidden = true)`, see [here](http://docs.swagger.io/swagger-core/v1.5.X/apidocs/index.html?io/swagger/annotations/ApiParam.html)
- Alternative 2: don't put this in tsoa. Instead, allow users to configure arbitrary filters that extend tsoa, similar to how Swagger .NET works, see [here](https://github.com/domaindrivendev/Swashbuckle.AspNetCore#extend-generator-with-operation-schema--document-filters)

**Test plan**
- found all existing tests for `@Hidden` and added a query unit test to each

### DIFF
--- a/src/metadataGeneration/parameterGenerator.ts
+++ b/src/metadataGeneration/parameterGenerator.ts
@@ -1,5 +1,5 @@
 import * as ts from 'typescript';
-import { getNodeFirstDecoratorName, getNodeFirstDecoratorValue } from './../utils/decoratorUtils';
+import { getDecorators, getNodeFirstDecoratorName, getNodeFirstDecoratorValue } from './../utils/decoratorUtils';
 import { getJSDocTags } from './../utils/jsDocUtils';
 import { getParameterValidators } from './../utils/validatorUtils';
 import { GenerateMetadataError } from './exceptions';
@@ -11,7 +11,7 @@ import { TypeResolver } from './typeResolver';
 export class ParameterGenerator {
   constructor(private readonly parameter: ts.ParameterDeclaration, private readonly method: string, private readonly path: string, private readonly current: MetadataGenerator) {}
 
-  public Generate(): Tsoa.Parameter {
+  public Generate(): Tsoa.Parameter | null {
     const decoratorName = getNodeFirstDecoratorName(this.parameter, identifier => this.supportParameterDecorator(identifier.text));
 
     switch (decoratorName) {
@@ -107,7 +107,7 @@ export class ParameterGenerator {
     };
   }
 
-  private getQueryParameter(parameter: ts.ParameterDeclaration): Tsoa.Parameter {
+  private getQueryParameter(parameter: ts.ParameterDeclaration): Tsoa.Parameter | null {
     const parameterName = (parameter.name as ts.Identifier).text;
     const type = this.getValidatedType(parameter);
 
@@ -121,6 +121,13 @@ export class ParameterGenerator {
       required: !parameter.questionToken && !parameter.initializer,
       validators: getParameterValidators(this.parameter, parameterName),
     };
+
+    if (this.getQueryParamterIsHidden(parameter)) {
+      if (commonProperties.required) {
+        throw new GenerateMetadataError(`@Query('${parameterName}') Can't support @Hidden because it is required (does not allow undefined and does not have a default value).`);
+      }
+      return null;
+    }
 
     if (type.dataType === 'array') {
       const arrayType = type as Tsoa.ArrayType;
@@ -230,5 +237,19 @@ export class ParameterGenerator {
       typeNode = this.current.typeChecker.typeToTypeNode(type) as ts.TypeNode;
     }
     return new TypeResolver(typeNode, this.current, parameter).resolve();
+  }
+
+  private getQueryParamterIsHidden(parameter: ts.ParameterDeclaration) {
+    const hiddenDecorators = getDecorators(parameter, identifier => identifier.text === 'Hidden');
+    if (!hiddenDecorators || !hiddenDecorators.length) {
+      return false;
+    }
+
+    if (hiddenDecorators.length > 1) {
+      const parameterName = (parameter.name as ts.Identifier).text;
+      throw new GenerateMetadataError(`Only one Hidden decorator allowed on @Query('${parameterName}').`);
+    }
+
+    return true;
   }
 }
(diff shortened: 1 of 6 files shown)


README.MD
```diff
@@ -814,7 +814,7 @@ Excludes this endpoint from the generated OpenAPI Specification document.
   }
 ```
 
-It can also be set at the controller level to exclude all of its endpoints from the OpenAPI Specification document.
+Use on controllers to exclude all of its endpoints from the generated OpenAPI Specification document.
```
aldenquimby, line 817: I tweaked the language here so the sentences for method + controller + query all have parallel construction

WoH:
Ah, I forgot about that issue, that's really unfortunate.
I would've preferred an API change that's more in line with our current way to express this logic.
Since Properties on Classes or interfaces can be annotated with` /** @ignore */` JSDoc, the better approach would be to support 

```ts
  /**
   * @ignore optionalSecret
   * @ignore defaultSecret
   */
  @Get()
  public async find(
    @Query() normalParam: string,
    @Query() optionalSecret?: string,
    @Query() defaultSecret = true,
  ): Promise<any> {
    return;
  }
}
```

since it's more in line with other parameter modifiers, see [here](https://github.com/lukeautry/tsoa/blob/master/tests/fixtures/controllers/parameterController.ts)

I'll have to think about supporting the `@Hidden()` Decorator instead, please let me get back to you on this.

aldenquimby:
No worries, I'm happy to modify this PR!

Does jsdoc allow using `@ignore` together with `@param`? I don't see any examples from googling, and when I try your proposed syntax locally, my editor gives me a warning:
image

WoH:
> @WoH no worries, I'm happy to modify this PR!
> 
> Does jsdoc allow using `@ignore` together with `@param`?

I don't think JSDoc natively supports ignoring individual parameters, just like the JSON-Schema keywords we support (`@isInt`, `@minimum` etc. )

aldenquimby:
@WoH what are your latest thoughts on this PR? I can either rebase and resolve conflicts, or we can close it out if you'd prefer to go another direction.

I'm in favor of `@Hidden` because it's consistent with how tsoa hides controllers and methods.

My second choice would be `@Query({ hidden: true })` because that's how swagger java handles this.

My last choice would be using jsdocs because we'd need to create a new comment standard ([`@ignore` doesn't work on parameters](https://jsdoc.app/tags-ignore.html)). The current tsoa comments are nice because they follow jsdoc and json-schema standards.

src/metadataGeneration/parameterGenerator.ts
```diff
@@ -107,7 +107,7 @@ export class ParameterGenerator {
     };
   }
 
-  private getQueryParameter(parameter: ts.ParameterDeclaration): Tsoa.Parameter {
+  private getQueryParameter(parameter: ts.ParameterDeclaration): Tsoa.Parameter | null {
```
endor, line 110: Out of curiosity.. why did you go with `null` instead of `undefined`?

WoH:
@endor What do you think about the API for hiding parameters.
Should we go with `@Hidden()`, which is more more in line with how we hide Controllers/Methods or instead use JSDoc (`@ignore param`) which is similar to other modifiers for parameters with (i.e. `@isInt param` or `@minimum param 3`)

endor:
I think both are okay. At first I wanted to say that I have a slight tendency towards the JSDoc style, but thinking about it, most (all?) of the tags we have there are used for validation, but `ignore/hidden` isn't. So I think `@Hidden` might be a better solution.

WoH:
> I think both are okay. At first I wanted to say that I have a slight tendency towards the JSDoc style, but thinking about it, most (all?) of the tags we have there are used for validation, but ignore/hidden isn't. So I think @Hidden might be a better solution.

We use `@ignore` on interface/class properties aswell, but I'm fine with `@Hidden()` if you are.
I'll add a review, but rebasing this should be fine @aldenquimby

endor:
@WoH Well, that's a good point. If we want to have backwards compatibility, we could try to support both - the decorator and the jsdoc comment - for all cases: interface/class, property, method, controller. Does that make sense? If so, then I guess we still need an open issue after this PR.

src/metadataGeneration/parameterGenerator.ts
aldenquimby: I chose `null` because to me that means `there is no query parameter`, where `undefined` would mean something closer to `I'm not sure if there is a query parameter`. I also saw [this similar PR](https://github.com/lukeautry/tsoa/pull/310/files#diff-617900d54e0562ba65c1a694adb27dddR18) which used `null` too.

Happy to swap to undefined if that's preferred though!

endor: If it's already used in other, similar places, we can just leave it. Personally I prefer `undefined`, because to me it's more in line with optional parameters.

aldenquimby:
@WoH @endor I rebased this PR and split out the doc changes: https://github.com/tsoa-community/docs/pull/2. Please let me know if there's anything else I should do for this PR.

(As an aside, my company upgraded to tsoa 3 this week and had no hiccups - the upgrade guide was solid. Thank you both for your contributions to this library!)

WoH:
Thanks a lot for adding this and bearing with me.

PR #676, opened 2020-04-30
fix: avoid crash with indexed types

Eywek:
### All Submissions:

### Note

Without this, tsoa is crashing because `hasType()` is called with undefined.
I've still have some trouble with the `Record<>`, it's not properly defined because type here https://github.com/Kiwup/tsoa/blob/indexed-fixes/src/metadataGeneration/typeResolver.ts#L145 isn't found


WoH:
> I've still have some trouble with the Record<>, it's not properly defined because type here https://github.com/Kiwup/tsoa/blob/indexed-fixes/src/metadataGeneration/typeResolver.ts#L145 isn't found

I assume this happening regardless of this PR (?). Can you get a reproduction case there?
Can't come up with an example where you'd end up resolving a mapped type that's not referenced by a type reference.

tests/unit/swagger/schemaDetails3.spec.ts
```diff
@@ -1156,6 +1157,31 @@ describe('Definition generation for OpenAPI 3.0.0', () => {
...
                   },
                   id: { type: 'number', format: 'double', default: undefined, description: undefined, example: undefined },
+                  indexedResponse: {
+                    default: undefined,
+                    description: undefined,
+                    example: undefined,
+                    format: undefined,
+                    properties: {
+                      data: {
+                        $ref: '#/components/schemas/Record',
```
WoH, line 1167: This does not seem right.

Eywek: This is what I was talking about, and it's not related to this PR, it's just something I've noticed trying to debug my projects where I have tsoa (which was crashing without this PR)

WoH: It's good that you found this issue, but I think the change you are proposing here is trying to recover from a situation we should prevent earlier on.
As you noted, not having the referencer here is the red flag: It looks like we are not correctly propagating that. If you try to reconstruct the type afterwards without the referencer, we are missing major context we need, especially the type arguments.

Eywek: Hm right, so what do you think we need to do here?

WoH: If you want to debug, try to step through from the point where the TypeReference node is resolved until the point where the referencer is gone.Somewhere in between is the issue :laughing:  Hope the "Pretest"debug config comes in handy. 
Wish I could tell you what the issue is, but I can't tell until I get time to debug this myself.

Eywek: I've updated test model with a version of type more simple which still cause this issue.

I've tried to inspect, but I haven't found anything, the only things I have noticed are that [declaration](https://github.com/lukeautry/tsoa/pull/676/files#diff-04ed568b4a647b0bfd718a5afdb822fbR262) have `aliasTypeArguments` and `aliasSymbol` properties which seems to be the final type we need (`Record<'id', string>`) and when we use `typeToTypeNode` on this declaration ([here](https://github.com/lukeautry/tsoa/pull/676/files#diff-04ed568b4a647b0bfd718a5afdb822fbR264)) we have an `anonymous` node where `pos` aren't defined (which I think cause the issue).
And then, when we are in the [mapped type scope](https://github.com/lukeautry/tsoa/pull/676/files#diff-04ed568b4a647b0bfd718a5afdb822fbR264), using `getTypeFromTypeNode` on `this.referencer` doesn't work, indeed it create an error node. I've tried to use `getTypeFromTypeNode` on `this.typeNode` and we get back a type with `aliasTypeArguments` and `aliasSymbol`, and using `type.getProperties()` doesn't returns any props.

Eywek:
Did you have time to check that @WoH? I've tried to inspect this on my own (cf. review comments) but I haven't found a way to handle it and I don't really know how to debug more

WoH:
Not yet, trying to find (meaningful) time during the week is hard.

WoH:
Alright, I was able to get us to a point where we can start working on this.
The issue comes from `propertyFromDeclaration` and `propertyFromSignature`.
Both of these don't propagate the referencer. If we add the referencer to the `new TypeResolver()` call we get the proper error message:

```
Error: Could not determine the keys on any
At: /home/woh/Code/tsoa/tsoa_origin/tests/fixtures/testModel.ts:626:626.
This was caused by 'public indexedResponse?: IndexRecordAlias['foo'];'
```

Eywek:
Hm, sorry but I haven't understood what you mean with:
> referencer -> getTypeFromTypeNode -> getProperties("indexedResponse").type.getProperty("id")

WoH:
> Hm, sorry but I haven't understood what you mean with:
> 
> > referencer -> getTypeFromTypeNode -> getProperties("indexedResponse").type.getProperty("id")

What I was trying to get at is:

> The issue comes from propertyFromDeclaration and propertyFromSignature.
> Both of these don't propagate the referencer. If we add the referencer to the new TypeResolver() call we get the proper error message

After adding this, you could set a breakpoint before the crash happens and get the type tsoa should resolve by calling:

```ts
this.current.typeChecker.getTypeFromTypeNode(this.referencer).getProperty("indexedResponse").type.getProperty("id")
```

I assume we could be clever here and not propagate the original referencer in propertyFrom(Signature|Declaration), but instead set the type or the inferred type of the Signature|Declaration as the referencer. This way, when we are resolving the indexed type, we can call getProperty(name) and get something we can resolve subsequently

Eywek:
Thank you, now, it's working, the only remaining issue is that we have the following type in the `propertyFromDeclaration`:
```js
{
  dataType: 'refAlias',
  default: undefined,
  description: 'Construct a type with a set of properties K of type T',
  refName: 'Record',
  format: undefined,
  type: { dataType: 'nestedObjectLiteral', properties: [ [Object] ] },
  validators: {}
}
```
We have the right type in `type` it's why I've done that to test:
```ts
// @ts-ignore
if (type.refName === 'Record') {
  // @ts-ignore
  type = type.type
}
```
Because with the `refAlias` datatype, `type` is not used.

Maybe you have any idea on this?

WoH:
If I understand your issue correctly, I think do. I believe we have a naming issue here which we should fix in order to avoid a special case. The origin seems to be how we find out what a reference is named. For non-synthetic nodes, that's pretty simple bc TS exposes getText to get the name. Otherwise, we chose a name that does not factor in type arguments [here](https://github.com/lukeautry/tsoa/blob/master/src/metadataGeneration/typeResolver.ts#L441).
If we can add these, we should generate a name that does not clash with other types unless they produce the same piece of OAS.

Eywek:
Right, It seems to work now

WoH:
Looks pretty good now, my main worry is the args to name, that thing looks quite fragile right now.

Eywek:
Sure, but I haven't found if actual code that can help me with it. Maybe you want to add more tests?

src/metadataGeneration/typeResolver.ts
```diff
@@ -813,7 +836,7 @@ export class TypeResolver {
       format: this.getNodeFormat(propertySignature),
       name: identifier.text,
       required,
-      type: new TypeResolver(propertySignature.type, this.current, propertySignature.type.parent, this.context).resolve(),
+      type: new TypeResolver(propertySignature.type, this.current, propertySignature.type.parent, this.context, propertySignature.type as ts.TypeReferenceType).resolve(),
```
WoH, line 839: Maybe we can check if it's a ref. If it is, we can set is without casting, if it is not we don't need it afaik.

src/metadataGeneration/typeResolver.ts
```diff
@@ -832,7 +855,7 @@ export class TypeResolver {
       throw new GenerateMetadataError(`No valid type found for property declaration.`);
     }
 
-    const type = new TypeResolver(typeNode, this.current, propertyDeclaration, this.context).resolve();
+    const type = new TypeResolver(typeNode, this.current, propertyDeclaration, this.context, typeNode as ts.TypeReferenceType).resolve();
```
WoH, line 858: Maybe we can check if it's a ref. If it is, we can set is without casting, if it is not we don't need it afaik.

WoH:
> Sure, but I haven't found if actual code that can help me with it. Maybe you want to add more tests?

Let me see if I can come up with something. First thing that comes to mind is an object literal, i.e.:
`Record<"id", {myProp1: string, myProp2: boolean}>`

src/metadataGeneration/typeResolver.ts
Eywek: It seems that `typeNode` (in our case) is a `IndexedAccessType` instead of a `TypeReferenceType`, so if we can a `if` it don't work anymore

Eywek:
> Let me see if I can come up with something. First thing that comes to mind is an object literal, i.e.:
> `Record<"id", {myProp1: string, myProp2: boolean}>`

Indeed, with that it crash in the `getModelTypeDeclaration` method (No matching model found for referenced type) because the `type` is
```js
IdentifierObject {
  pos: -1,
  end: -1,
  flags: 8,
  modifierFlagsCache: 0,
  transformFlags: 0,
  parent: undefined,
  kind: 75,
  escapedText: '... 5 more ...',
  originalKeywordKind: undefined,
  autoGenerateFlags: 0,
  autoGenerateId: 0
}
```

src/metadataGeneration/typeResolver.ts
WoH: This means we should either change our expectations of a `referencer` or find a good assertion function that makes sure whatever we pass here won't crash. As is, this will eventually crash.

Eywek:
I don't really have time to work on this right now, so feel free to fix few missing things if you want to @WoH. I've added you on the repo.

Eywek:
So, I've found some time to check with this.
I've added a test for `Record<"id", {myProp1: string, myProp2: boolean}>` as you asked, which generate `#/components/schemas/Record_id.any_` for now.
I've update the type of `this.referencer` to `typeNode`, it haven't raise any errors and tests aren't failing

I don't know if you want to fix the issue about object literals or add tests for the referencer type.

WoH:
I think any kind of reasonable bail out is more useful than the current behavior, so I think if we can ensure the issue is at least communicated I'm down to merge this.
Will take a look this week.

Eywek:
Hey, did you have time to check and merge this one?

WoH:
I tried to find a way to make this work for a broader set of types but haven't come up with anything that feels useful yet.

Eywek:
So you want to find a way to make this work for a broader set of types before merging?

Eywek:
Rebased to resolve conflicts, @WoH do you have any plan for this PR?

Eywek:
Thank you! When do you plan to release @WoH?

Eywek:
Up, do you plan to release soon @WoH? 
I'm waiting for this to use the official version
Thank you

WoH:
I've been busy with academic work and my job which are higher prio right now. Since a new release includes the split into runtime and cli using lerna I want to have time to ensure the upgrade is smooth for everyone without me accidentally breaking things. I don't tend to give ETAs on hobbies but I hope it'll be done properly throughout next week.

WoH:
Published as 3.3.0-alpha.0 (tag v3beta). This includes not `@tsoa/runtime` and `@tsoa/cli`

PR #682, opened 2020-05-05
Add ability to define same responses for all methods in controller.

### LINKED ISSUE #680: Is it possible to define error response on @Route level?
Hello, i have some protected routes that available only after authorization, so i think it is good to add possible response 401 to all controller's methods by one line. Also it is easy to forget decorate for method when project already in support stage.
## Sorting

- **I'm submitting a ...**

  - [ ] bug report
  - [x] feature request
  - [x] support request

- I confirm that I

## Expected Behavior
I expect that after i define `@Response` on `@Route` level swagger config will add the same response type to every controller's method.

```ts
@Route('boards')
@Tags('Boards')
@Security('jwt')
@Response('401', 'Unathorized') <<
export class BoardsController extends Controller {
```

## Current Behavior
No errors defined in swagger config.

## Context (Environment)

Version of the library: ^2.5.14
Version of NodeJS: v12.10.0

ivan-shaban:
### All Submissions:

**Closing issues**

closes #680 

Why this approach? Seems like sometimes need to define the same responses for all methods in controller, e.g. when they require authorization. For now it is possible only to define all responces manually. It is not handy and may proceed human error (e.g. to forgot to define responce for new method). 

Solution: Ability to define common responses in controller on route level, which common responce should has `name`, otherwise error will be created. All common responses merges with method's personal responses.

**Test plan**

I don't understand exactly where and what test, please point me what to do if i miss something.

### DIFF
--- a/src/metadataGeneration/controllerGenerator.ts
+++ b/src/metadataGeneration/controllerGenerator.ts
@@ -46,7 +46,7 @@ export class ControllerGenerator {
   private buildMethods() {
     return this.node.members
       .filter(m => m.kind === ts.SyntaxKind.MethodDeclaration)
-      .map((m: ts.MethodDeclaration) => new MethodGenerator(m, this.current, this.tags, this.security, this.isHidden, this.commonResponses))
+      .map((m: ts.MethodDeclaration) => new MethodGenerator(m, this.current, this.commonResponses, this.tags, this.security, this.isHidden))
       .filter(generator => generator.IsValid())
       .map(generator => generator.Generate());
   }
@@ -77,7 +77,7 @@ export class ControllerGenerator {
 
       const [name, description, examples] = getDecoratorValues(decorator, this.current.typeChecker);
       if (!name) {
-        throw new GenerateMetadataError(`Controller's responses should has explicit name.`);
+        throw new GenerateMetadataError(`Controller's responses should have an explicit name.`);
       }
 
       return {
--- a/src/metadataGeneration/controllerGenerator.ts
+++ b/src/metadataGeneration/controllerGenerator.ts
@@ -1,21 +1,24 @@
 import * as ts from 'typescript';
-import { getDecorators, getSecurites } from './../utils/decoratorUtils';
+import { getDecorators, getDecoratorValues, getSecurites } from './../utils/decoratorUtils';
 import { GenerateMetadataError } from './exceptions';
 import { MetadataGenerator } from './metadataGenerator';
 import { MethodGenerator } from './methodGenerator';
 import { Tsoa } from './tsoa';
+import { TypeResolver } from './typeResolver';
 
 export class ControllerGenerator {
   private readonly path?: string;
   private readonly tags?: string[];
   private readonly security?: Tsoa.Security[];
   private readonly isHidden?: boolean;
+  private readonly commonResponses: Tsoa.Response[];
 
   constructor(private readonly node: ts.ClassDeclaration, private readonly current: MetadataGenerator) {
     this.path = this.getPath();
     this.tags = this.getTags();
     this.security = this.getSecurity();
     this.isHidden = this.getIsHidden();
+    this.commonResponses = this.getCommonResponses();
   }
 
   public IsValid() {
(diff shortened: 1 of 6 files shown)


src/metadataGeneration/methodGenerator.ts
```diff
@@ -40,7 +41,7 @@ export class MethodGenerator {
       nodeType = typeChecker.typeToTypeNode(implicitType) as ts.TypeNode;
     }
     const type = new TypeResolver(nodeType, this.current).resolve();
-    const responses = this.getMethodResponses();
+    const responses = this.commonResponses ? this.commonResponses.concat(this.getMethodResponses()) : this.getMethodResponses();
```
WoH, line 44: we should always pass a common responses array here, so we don't have to check if it exists

src/metadataGeneration/controllerGenerator.ts
```diff
@@ -63,6 +66,29 @@ export class ControllerGenerator {
...
+    if (!decorators || !decorators.length) {
+      return [];
+    }
+
+    return decorators.map(decorator => {
+      const expression = decorator.parent as ts.CallExpression;
+
+      const [name, description, examples] = getDecoratorValues(decorator, this.current.typeChecker);
+      if (!name) {
+        throw new GenerateMetadataError(`Controller's responses should has explicit name.`);
```
WoH, line 80:
```suggestion
        throw new GenerateMetadataError(`Controller's responses should have an explicit name.`);
```

tests/fixtures/controllers/controllerWithCommonResponses.ts
```diff
@@ -0,0 +1,16 @@
+import { Controller, Get, Response, Post, Route } from '../../../src';
+import { ModelService } from '../services/modelService';
+import { TestModel } from '../testModel';
+
+@Route('Controller')
+@Response('401', 'Unauthorized')
```
WoH, line 6: Let's add a `@Response<Type>('404', 'Not found')` so we can make sure it also works with a type

tests/fixtures/controllers/methodController.ts
```diff
@@ -145,6 +145,7 @@ export class MethodController extends Controller {
   }
 
   @Get('returnAliasedVoidType')
+  // tslint:disable-next-line:no-empty
   public async returnAliasedVoidType(): Promise<VoidAlias1> {}
```
WoH, line 149: I assume this should avoid any issues with tslint:
```suggestion
  public async returnAliasedVoidType(): Promise<VoidAlias1> {
    return;
  }
```

src/metadataGeneration/methodGenerator.ts
ivan-shaban: Agreed, fixed.

src/metadataGeneration/controllerGenerator.ts
ivan-shaban: Agreed, fixed.

tests/fixtures/controllers/methodController.ts
ivan-shaban: It is not my mistake, but i think to disable rule is more clean fix, then adding some redundant code that only aims to disable rule and does nothing, but i can fix like you suggest.

tests/fixtures/controllers/controllerWithCommonResponses.ts
ivan-shaban: MMmm, please check it, i'm not sure that understand you well.

tests/fixtures/controllers/methodController.ts
WoH: I disagree, we should conform to the coding style tslint suggests unless we have a very good reason not to.
I know you didn't introduce the issue, but if you want to address it, this would be the better way.

src/metadataGeneration/controllerGenerator.ts
```diff
@@ -63,6 +66,29 @@ export class ControllerGenerator {
...
   }
 
+  private getCommonResponses(): Tsoa.Response[] {
+    const decorators = getDecorators(this.node, identifier => identifier.text === 'Response');
+    if (!decorators || !decorators.length) {
+      return [];
+    }
+
+    return decorators.map(decorator => {
+      const expression = decorator.parent as ts.CallExpression;
```
WoH, line 76: I'd prefer if we would check here, not cast. Something roughly like:
```suggestion
      const expression = decorator.parent;
      
      if (!ts.isCallExpression(expression)) {
        throw new GenerateMetadataError(`The @Response decorator should only be applied at the method or class level`);
      }
```

tests/fixtures/controllers/methodController.ts
ivan-shaban: Ok, fixed.

src/metadataGeneration/controllerGenerator.ts
ivan-shaban: Hm it is not a big deal to fix it, but i don't understand how that case may happens? In `getCommonResponses` method i have check only controller's (`node`'s) decorators, so it is impossible that it may be method's decorators or any other thing. Is it really need to be fixed?

PR #690, opened 2020-05-07
Use Lerna to split TSOA into runtime and cli packages

### LINKED ISSUE #546: Split `tsoa` package
## Sorting

* **I'm submitting a ...**
  - [ ] bug report
  - [x] feature request
  - [ ] support request

* I confirm that I
    * There was #304 but without any real solution...

## Expected Behavior

Adding TSOA should not drastically increase the size of my deployed application (and it should not add unused dependencies to it).

## Current Behavior

Adding TSOA to a(n empty) project causes 65 npm packages to be added with a size of over 60MB (with 50MB for typescript alone). Especially having to include `typescript` in my deployed productive code is really bad in my opinion.

## Possible Solution

I think it would make much sense to split tsoa into 2 package, one for the CLI (`tsoa-cli`) and one Runtime package (`tsoa-runtime`).
This way `tsoa-cli` could be used as a global or dev dependency and only the needed code is deployed, with the productive runtime dependency. 
Runtime would probably just include: 
* controller base class
* decorators 
* and maybe a few other small things

## Breaking change?

Depending on the chosen package naming, the change might break existing builds and/or deployments (if they would be updated).

Possible options are: 

1. `tsoa`(CLI) --> `tsoa-runtime`
    * would probably not break anything
    * silent update might lead cause possible improvements to be wasted (as long as the CLI package stays in the productive dependencies)
2. `tsoa`(empty) --> `tsoa-cli` --> `tsoa-runtime`
    * basically the same as above, with nicer package name for CLI
3. `tsoa-cli` --> `tsoa` (Runtime)
   * would break the build, but not the runtime (as long as you don't try to actually use CLI features at runtime, but this should usually not be the case)
    * naming schema would be in line what many other (big) packages to (e.g. angular, babel, vue, gulp, grunt, karma, aurelia)
    * instant win for the runtime
    * fixing the build (by adding CLI package to the dev dependencies should be easy
4. `tsoa-cli` --> `tsoa-runtime`; deprecate `tsoa`
    * clear naming
    * would not break anything
    * forces explicit migration

I would prefer option 3 or 4 (event if they are slightly less comfortable), as 1 and 2 feel somewhat dirty to me.

cc: @ryo-nbx

Kriskit:
### All Submissions:

**Closing issues**

closes #546 

This approach potentially adds a tiny bit more complexity in maintenance but you gain more flexibility and for my purposes, less complexity in deployment because I don't have to worry about it having to bundle things like Typescript into my deployment package. 

I think this is a good starting point for splitting up the code in to necessary libraries. It could probably split further. At present I have left the tests on their own in the root of the project because they cover code that is in both the `@tsoa/runtime` and `@tsoa/cli` packages. At the moment `@tsoa/runtime` exports some additional things that weren't exported before (e.g. the `assertNever` utility). Another package called something like `@tsoa/core` could be introduced to house common code that doesn't need exporting at the runtime or cli level.

I've included a `tsoa` package for backwards compatibility as suggested in the issue (#546) so tsoa can still be used as it was before and it exports all the same things that it did previously (with the above exceptions).

I understand this is quite a big structural change but it does help make things simpler if you're bundling things and/or trying to keep your deployments smaller.

**Test plan**
All unit tests run as they used to. The difference is that they reference the modules imported through aliases instead the code directly - which means the `canImportByAlias` flag is no longer necessary. I left it in but set it to `true` by default for backwards compatibility where people have customised templates.

I tested publishing as best as I could by publishing to the GitHub package registry, installed from there on a local test project (both as split modules and as the combined `tsoa` module) and ran the routes and spec commands successfully. *However* I have never published a package to NPM before - I have done my best to set up scripts to accommodate but it might not be what is wanted.


WoH:
@lukeautry do you own the `@tsoa` org on npm?

WoH:
@Kriskit can you move cli/, runtime/ and tsoa/ into a packages/ dir, everything else lgtm after initial review.

Kriskit:
@WoH Done :) I'm happy to test this as a pre-release package or something if they can be published properly. I've got a project with 20+ api's using tsoa and using webpack to bundle it so it'd be interesting to see what the difference is and if I run into any issues.

.gitignore
```diff
@@ -5,3 +5,4 @@ routes.ts
 customRoutes.ts
 .idea/
 yarn-error.log
+tsconfig.tsbuildinfo
```
WoH, line 8: Add new lint, preferrably run `yarn run prettier --write .` once.

.vscode/tasks.json
```diff
@@ -17,6 +17,11 @@
       "label": "prepareFiles",
       "group": "none",
       "args": ["run", "pretest"]
```
WoH, line 19:
```suggestion
      "args": ["run", "pretest"],
      "options": { "cwd": "tests" }
```

lerna.json
```diff
@@ -0,0 +1,14 @@
...
+    "command": {
+      "version": {
+        "allowBranch": "master"
+      },
+      "publish": {
+        "allowBranch": "master"
+      }
+    }
+  }
+
```
WoH, line 14: \n

Kriskit:
I've done all the changes and managed to fix the tests not running (now has 957), for some reason I had to put the spec glob into single quotes 🤷‍♂️

WoH:
We've reached out to npm to get access to the `tsoa` namespace, hope it'll happen soon.
In the meantime, I've looked at the PR some more, for now, it'd be great if we could start moving the tests to their respective packages or at least make sure we don't import from `dist/`, i.e. `@tsoa/cli/dist/cli` which is all over the tests currently.

package.json
```diff
@@ -29,92 +30,40 @@
...
+    "clean": "lerna run clean --stream",
     "prepare": "yarn build",
-    "deploy": "yarn version patch -m \"Release v%s\" && yarn publish",
     "preversion": "yarn test",
-    "postversion": "git push origin master && git push --follow-tags",
-    "pretest": "cross-env NODE_ENV=tsoa_test ts-node ./tests/prepare.ts",
-    "loading-msg-for-tests": "echo '... starting a test run to ensure quality. Note this takes some time as tests do not run in parallel (by design) ...' ",
-    "test": "node ensureYarn.js && tsc -p ./tests/tsconfig.json --noEmit && yarn loading-msg-for-tests && cross-env NODE_ENV=tsoa_test mocha **/*.spec.ts --exit --require ts-node/register --timeout 5000",
+    "test": "yarn --cwd tests test",
+    "lint": "lerna run lint --stream",
```
ivan-shaban, line 40: Because all sub projects has same linting rules maybe it is better to run them from root config, and remove that logic from sub configs, e.g. in root `package.json`:
```    
"lint": "eslint packages/**/*.{ts,tsx}",
```
I agree that build flow should be unique per project, but looks like that linting it is pretty common thing.

ivan-shaban:
BTW as you intend to use lerna, maybe you wanna to use [conventual commits](conventionalcommits.org) as well? It is simplify and automate versioning flow.

WoH:
> BTW as you intend to use lerna, maybe you wanna to use conventual commits as well? It is simplify and automate versioning flow. I can make PR to author's branch with example configuration.

Please file this as an issue so we can discuss it separately

package.json
Kriskit: Yeah I agree, I've moved this now

Kriskit:
@WoH I don't know how you feel about this but since all the tests run using ts-node, I've used tsconfig paths instead of having a dependency on the modules. This removes the `/dist/` imports and doesn't require the code to be transpiled for the tests to run.

WoH:
> I'm not comfortable attempting to split the tests up as I'm really not sure where it would be possible to split them up in a nice way at the moment.

This is just a rough starting point, but:
As far as I know, the only tests that need both are the integrations tests.
The only things that need the runtime are the validation service tests.

WoH:
Quick update: I hope we hear from npm soon, I'd really like to merge this.
The 4 weeks have basically passed now, seems like they have not heard back from the user, at least we were not informed.

WoH:
Update: We got ownership of the org. Could you rebase this PR?

Kriskit:
> Update: We got ownership of the org. Could you rebase this PR?

Pull request is now in line with master 👍

endor:
I don't really know anything about Lerna, so I can't really contribute much here. I trust @WoH's judgement on this.

PR #693, opened 2020-05-11
endpoints with the same path are ignored

### LINKED ISSUE #453: endpoints with the same path are ignored
So this one is easy enough for the developer to figure out on their own and eventually rectify it, but it might be nice to add some good developer experience (DX) by throwing if a user has a duplicate route.

**Bad situation:**
If you have two endpoints in the same route with identical paths, then tsoa will not generate swagger for the second endpoint.

**Expected result:**

Tsoa should throw an error saying, `You can not have duplicate routes. The following route is a duplicate ${duplicateRoutePath}`

**Example:**

```ts
@Route('person')
@Tags('_Person Controller')
export class PersonController extends Controller {

    @Post()
    public async savePerson(
        @Body() newPerson: IPerson
    ): Promise<IPersonSaved> {
         //...
    }

    @Post()
    public async saveSchool(
        @Body() newSchool: ISchool
    ): Promise<ISchoolSaved> {
         //...
    }
}
```

You'll notice that the function names are different but since the paths are identical, it won't make the second one. If we were to throw, it would immediately make the developer aware. They would think "oh I didn't mean to put saveSchool in the person controller" or maybe "I meant to put it on @Post('school')".

This is just to help make tsoa a more pleasurable experience so that more people are interested in using it. :)

ivan-shaban:
### All Submissions:

**Closing issues**

closes #453

### DIFF
--- a/src/metadataGeneration/controllerGenerator.ts
+++ b/src/metadataGeneration/controllerGenerator.ts
@@ -34,15 +34,10 @@ export class ControllerGenerator {
     }
 
     const sourceFile = this.node.parent.getSourceFile();
-    const methods = this.buildMethods();
-    const methodSignatureDuplicates = this.getMethodSignatureDuplicates(methods);
-    if (methodSignatureDuplicates.length) {
-      throw new GenerateMetadataError(`Controller has methods with same routes: \n"${methodSignatureDuplicates.join('\n\n')}".`);
-    }
 
     return {
       location: sourceFile.fileName,
-      methods,
+      methods: this.buildMethods(),
       name: this.node.name.text,
       path: this.path || '',
     };
@@ -54,17 +49,6 @@ export class ControllerGenerator {
       .map((m: ts.MethodDeclaration) => new MethodGenerator(m, this.current, this.commonResponses, this.tags, this.security, this.isHidden))
       .filter(generator => generator.IsValid())
       .map(generator => generator.Generate());
-  }
-
-  private getMethodSignatureDuplicates(methods: Tsoa.Method[]) {
-    const result: string[] = [];
-    methods.forEach(targetMethod => {
-      const duplicates = methods.filter(({ method, path }) => targetMethod.method === method && targetMethod.path === path);
-      if (duplicates.length > 1) {
-        result.push(`@${targetMethod.method}(${targetMethod.path})\n${targetMethod.name}`);
-      }
-    });
-    return result;
   }
 
   private getPath() {
--- a/src/metadataGeneration/controllerGenerator.ts
+++ b/src/metadataGeneration/controllerGenerator.ts
@@ -37,7 +37,7 @@ export class ControllerGenerator {
     const methods = this.buildMethods();
     const methodSignatureDuplicates = this.getMethodSignatureDuplicates(methods);
     if (methodSignatureDuplicates) {
-      throw new GenerateMetadataError(`Controller has methods with same routes: "${methodSignatureDuplicates.join('\n')}".`);
+      throw new GenerateMetadataError(`Controller has methods with same routes: \n"${methodSignatureDuplicates.join('\n\n')}".`);
     }
 
     return {
@@ -61,7 +61,7 @@ export class ControllerGenerator {
     methods.forEach(targetMethod => {
       const duplicates = methods.filter(({ method, path }) => targetMethod.method === method && targetMethod.path === path);
       if (duplicates.length > 1) {
-        result.push(`\n@${targetMethod.method}(${targetMethod.path})\n${targetMethod.name}`);
+        result.push(`@${targetMethod.method}(${targetMethod.path})\n${targetMethod.name}`);
       }
     });
     if (result.length) {
--- a/src/metadataGeneration/controllerGenerator.ts
+++ b/src/metadataGeneration/controllerGenerator.ts
@@ -31,10 +31,15 @@ export class ControllerGenerator {
     }
 
     const sourceFile = this.node.parent.getSourceFile();
+    const methods = this.buildMethods();
+    const methodSignatureDuplicates = this.getMethodSignatureDuplicates(methods);
+    if (methodSignatureDuplicates) {
+      throw new GenerateMetadataError(`Controller has methods with same routes: "${methodSignatureDuplicates.join('\n')}".`);
+    }
 
     return {
       location: sourceFile.fileName,
-      methods: this.buildMethods(),
+      methods,
       name: this.node.name.text,
       path: this.path || '',
     };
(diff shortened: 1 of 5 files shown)


src/metadataGeneration/tsoa.ts
```diff
@@ -14,7 +14,7 @@ export namespace Tsoa {
   export interface Method {
     deprecated?: boolean;
     description?: string;
-    method: 'get' | 'post' | 'put' | 'delete' | 'options' | 'head' | 'patch' | 'head';
+    method: 'get' | 'post' | 'put' | 'delete' | 'options' | 'head' | 'patch';
```
WoH, line 17: Was this removed intentionally?

ivan-shaban: Yes, there was duplicate in enumeration.

endor:
@HoldYourWaffle asked in the original issue whether it's valid to have two separate controllers bound to the same path. Since this PR closes the issue - what happens in that case?

WoH:
> @HoldYourWaffle asked in the original issue whether it's valid to have two separate controllers bound to the same path. Since this PR closes the issue - what happens in that case?

We throw with an Error, listing both conflicting methods. Seems like the best way to handle it, right?

ivan-shaban:
Currently i dont check other controllers methods, i only add validation for one controller's methods. Anyway what you asking about is more tricky that appears, e.g. what sould we do in case if one controller has explicit method route, e.g. `@Get('users/all')` and another controller has route with variable e.g.: `@Get('users/{id}')`? Can we catch that? Should we catch that case? (I understand that i've provided probably poor example, and real scenario should be implemented in another way, but it is still possible case.

src/metadataGeneration/controllerGenerator.ts
```diff
@@ -51,6 +56,21 @@ export class ControllerGenerator {
       .map(generator => generator.Generate());
   }
 
+  private getMethodSignatureDuplicates(methods: Tsoa.Method[]) {
```
endor, line 59: Why not add a return type and just return the empty array if there are no results? The method name would suggest getting an array anyway. And then above you could just check for the length of the array.

tests/unit/swagger/pathGeneration/routeDuplicates.spec.ts
```diff
@@ -0,0 +1,11 @@
+import { expect } from 'chai';
+import { MetadataGenerator } from '../../../../src/metadataGeneration/metadataGenerator';
+import { SpecGenerator2 } from '../../../../src/swagger/specGenerator2';
+import { getDefaultExtendedOptions } from '../../../fixtures/defaultOptions';
+
+it('should reject methods with same routes', () => {
+  expect(() => {
+    const invalidMetadata = new MetadataGenerator('./tests/fixtures/controllers/duplicateMethodsController.ts').Generate();
+    new SpecGenerator2(invalidMetadata, getDefaultExtendedOptions()).GetSpec();
```
endor, line 9: Do you need this line? Shouldn't it already throw above?

src/metadataGeneration/controllerGenerator.ts
ivan-shaban: 1. Typescript infer return type by itself, and it is more flexible way then explicit definition.
2. It doesnt matter to return an empty array or null, anyway i should add condition, tiny benefit i see if we return empty array, so probably, in future, we may concat it with another array without checks that value may be null, but for now returing an empty array left more questions then real sense.

endor: I would disagree.

1. While TS can infer the type, this is about making the code more readable for future readers. I don't want to parse the whole method before I know what's returned. 
2. Sure it matters. If you make it an array, you can directly return `result` instead of having to add the check here, which makes the code more concise. The method name suggests that you return an array, which you sometimes do. So `null` really comes as a surprise when calling this method. It might be different if the method were called `checkMethodSignaturesForDuplicates`.

This is just my personal preference though. What do you think @WoH?

tests/unit/swagger/pathGeneration/routeDuplicates.spec.ts
ivan-shaban: You are right, will be fixed. Just bad copypaste .

endor:
> Currently i dont check other controllers methods, i only add validation for one controller's methods. Anyway what you asking about is more tricky that appears, e.g. what sould we do in case if one controller has explicit method route, e.g. `@Get('users/all')` and another controller has route with variable e.g.: `@Get('users/{id}')`? Can we catch that? Should we catch that case? (I understand that i've provided probably poor example, and real scenario should be implemented in another way, but it is still possible case.

You bring up an interesting example. Does your PR catch something like this in the same controller? I think it might be nice if tsoa were to warn me about this, but I guess it might be also more complex to analyze. 

I think even if we leave this complexity aside, we could still try to catch duplicate methods in different controllers, if they are easily recognizable.

src/metadataGeneration/controllerGenerator.ts
ivan-shaban: 1. If you use modern IDE, it  can help you to check method signature. On other hand if you prefer to define return type explicit, it assumes that you do it in whole project, as result if you decide to change return type you should manually change all the types, it is not a big deal, but if typescript infer return type, most of types will be updated automatic that is more handful.
2. Ok, it has sense. I'll fix.

ivan-shaban:
> > Currently i dont check other controllers methods, i only add validation for one controller's methods. Anyway what you asking about is more tricky that appears, e.g. what sould we do in case if one controller has explicit method route, e.g. `@Get('users/all')` and another controller has route with variable e.g.: `@Get('users/{id}')`? Can we catch that? Should we catch that case? (I understand that i've provided probably poor example, and real scenario should be implemented in another way, but it is still possible case.
> 
> You bring up an interesting example. Does your PR catch something like this in the same controller? I think it might be nice if tsoa were to warn me about this, but I guess it might be also more complex to analyze.
> 
> I think even if we leave this complexity aside, we could still try to catch duplicate methods in different controllers, if they are easily recognizable.

I prefer to implement it in another PR.

ivan-shaban:
I've rework logic, now it checks signatures in all controllers.
BTW If you tell me properly way how to match `user/cool` and `users/{id}`, i can improve PR even more!

WoH:
> I've rework logic, now it checks signatures in all controllers.
BTW If you tell me properly way how to match user/cool and users/{id}, i can improve PR even more!

That overlap may be intended, so I think your check would be asstrict as we would want it to be.

src/metadataGeneration/tsoa.ts
WoH: Good point :laughing:

ivan-shaban:
> > I've rework logic, now it checks signatures in all controllers.
> BTW If you tell me properly way how to match user/cool and users/{id}, i can improve PR even more!
> 
> That overlap may be intended, so I think your check would be asstrict as we would want it to be.

I just can add warning message, not error. Imho, better to explicit notify about possible error.

WoH:
> I just can add warning message, not error. Imho, better to explicit notify about possible error.

I'd split this out into a separate PR tbh.
Current PR lgtm, but I'll let you keep working within the PR if you want to.

ivan-shaban:
Can to show properly \ expected way to match urls?

ivan-shaban:
Let merge this PR, i continue work in another one.

WoH:
> Can you show properly \ expected way to match urls?

I don't know what the most efficient way would be, but I'd try to transform the path templates into regexed and try to find matches.

PR #703, opened 2020-05-19
Fix: @is[NumType] comment tag force follow with variable name issue

### LINKED ISSUE #622: @is[number format] tags only work if a comment is present
## Sorting

* **I'm submitting a ...**
  - [x] bug report
  - [ ] feature request
  - [ ] support request

* I confirm that I

## Expected Behavior
@is[number format] tags (isInt, isLong, isDouble, and isFloat) should not require a comment after the tag.
```ts
/**
 * @isInt 
 */
export type NumberAlias = number;
```
```json
"NumberAlias": {
  "type": "integer",
  "format": "int32"
}
```

## Current Behavior
Currently, unless a comment is included, no number format is generated.
### Without comment
```ts
/**
 * @isInt 
 */
export type NumberAlias = number;
```
```json
"NumberAlias": {
  "type": "number",
  "format": "double"
}
```

### With comment
```ts
/**
 * @isInt test
 */
export type NumberAlias = number;
```
```json
"NumberAlias": {
  "type": "integer",
  "format": "int32"
}
```

## Possible Solution
Problem seems to be in getJSDocTagNames, where a comment is required
```ts
export function getJSDocTagNames(node: ts.Node) {
  let tags: ts.JSDocTag[];
  if (node.kind === ts.SyntaxKind.Parameter) {
    const parameterName = ((node as any).name as ts.Identifier).text;
    tags = getJSDocTags(node.parent as any, tag => {
      return tag.comment !== undefined && tag.comment.startsWith(parameterName);
    });
  } else {
    tags = getJSDocTags(node as any, tag => {
      return tag.comment !== undefined;               <====== Matcher requires comment
    });
  }
  return tags.map(tag => {
    return tag.tagName.text;
  });
}
```

## Steps to Reproduce
1. Create a numeric type alias
2. Add a "@isInt" tag in the jsdoc comment
3. Compile and see that the format and type are wrong

## Context (Environment)
Version of the library: v3.0.4
Version of NodeJS: 12

## Detailed Description
It seem likes the function used to get JSDoc tags from the node will only return tags with a comment included, which should not be required for number format assertions. This is the case for type aliases and classes/interface model properties.

jackey8616:
### All Submissions:

**Closing issues**
Closes #622 

This RP including following changes:
- Fix the issue itself, now the comment in model no more need specify variable name to take effect with correct type & format in SOA.  
- Orphan tag detection:
  Since the method's comment should always comes with parameter name.
  If tsoa detect any orphan tag without name during generation would throw out exception.

### DIFF
--- a/src/utils/jsDocUtils.ts
+++ b/src/utils/jsDocUtils.ts
@@ -1,4 +1,5 @@
 import * as ts from 'typescript';
+import { GenerateMetadataError } from '../metadataGeneration/exceptions';
 
 export function getJSDocDescription(node: ts.Node) {
   const jsDocs = (node as any).jsDoc as ts.JSDoc[];
@@ -17,16 +18,29 @@ export function getJSDocComment(node: ts.Node, tagName: string) {
   return tags[0].comment;
 }
 
-export function getJSDocTagNames(node: ts.Node) {
+export function getJSDocTagNames(node: ts.Node, requireTagName = false) {
   let tags: ts.JSDocTag[];
   if (node.kind === ts.SyntaxKind.Parameter) {
     const parameterName = ((node as any).name as ts.Identifier).text;
     tags = getJSDocTags(node.parent as any, tag => {
-      return tag.comment !== undefined && tag.comment.startsWith(parameterName);
+      if (ts.isJSDocParameterTag(tag)) {
+        // exception process @param tag
+        const inCommentName = (tag as ts.JSDocParameterTag).name as ts.Identifier;
+        if (inCommentName === undefined) {
+          throw new GenerateMetadataError(`Statement in comment: ${tag.getFullText()} has a orphan tag, a parameter name should follows with tag.`);
+        }
+        return inCommentName.text === parameterName || inCommentName.escapedText === parameterName;
+      }
+
+      // other tags like tsoaModel, is[NumberType], etc.
+      if (tag.comment === undefined) {
+        throw new GenerateMetadataError(`Statement in comment: ${tag.getFullText} has a orphan tag, a parameter name should follows with tag.`);
+      }
+      return tag.comment.startsWith(parameterName);
     });
   } else {
     tags = getJSDocTags(node as any, tag => {
-      return tag.comment !== undefined;
+      return requireTagName ? tag.comment !== undefined : true;
     });
   }
   return tags.map(tag => {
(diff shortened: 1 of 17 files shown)


src/utils/jsDocUtils.ts
```diff
@@ -17,16 +18,21 @@ export function getJSDocComment(node: ts.Node, tagName: string) {
   return tags[0].comment;
 }
 
-export function getJSDocTagNames(node: ts.Node) {
+export function getJSDocTagNames(node: ts.Node, requireTagName = false) {
```
WoH, line 21:
```suggestion
export function getJSDocTagNames(node: ts.Node, requireTagName = true) {
```

src/utils/jsDocUtils.ts
```diff
@@ -17,16 +18,21 @@ export function getJSDocComment(node: ts.Node, tagName: string) {
...
+        return false;
+      } else if (tag.comment === undefined) {
+        throw new GenerateMetadataError(`Orphan tag: @${tag.tagName.text || tag.tagName.escapedText} should have a parameter name follows with.`);
+      }
+      return tag.comment.startsWith(parameterName);
     });
   } else {
     tags = getJSDocTags(node as any, tag => {
-      return tag.comment !== undefined;
+      return requireTagName ? tag.comment !== undefined : true;
```
WoH, line 35:
```suggestion
      return requireTagName ? true:  tag.comment !== undefined;
```

jackey8616: I think `requireTagName` represents the match function in `getJSDocTags` should check `tag.comment` is `undefined` or not.  
  
Am I right?

WoH: The match function was `tag => tag.comment !== undefined;`. Now it's `tag => requireTagName ? tag.comment !== undefined : true;`

I guess I just misinterpreted the meaning of the parameter since it, as an API consumer would likely mean "do I need this thing to have a comment" (since that's the behavioral difference). So what I thought is (reading the parameter name) the default for "do I need this thing to have a comment" should be true.

It's not a big difference either way, I'll merge this before the next release anyway.

jackey8616: So I don't need further changes right?

If yes, please mark this resolve, thanks you. 😃

PR #725, opened 2020-06-06
fix: SpecGen3: Render global tag descriptions if available

WoH:
### All Submissions:


endor:
I wonder whether we should do some verification on the data the user provides to make sure it fits with the OpenAPI 3 specification.

WoH:
> I wonder whether we should do some verification on the data the user provides to make sure it fits with the OpenAPI 3 specification.

Valid OAS3 Tags[] in case specVersion: 3  and valid OAS2 Tags[] in case specVersion: 2, but I'd agree that we may want to improve the support. Don't think it has to be part of this PR though.

PR #726, opened 2020-06-06
feat: specMerge: Validate config and add deepmerge option

WoH:
### All Submissions:

This PR introduces a new option to handle specMerging.
The reason for this addition is to handle cases where the merging needs to happen at a level in the spec where arrays are used, i.e. parameters.
In these cases, a tsoa user may want to specify an additional parameter without overriding the parameter array generated by tsoa.

A great example is [file uploads](https://tsoa-community.github.io/docs/file-upload.html#uploading-files) where we can currently only avoid the overriding by pushing the parameter of the override spec up a level (to the `/files/uploadFile` level) where it then unfortunately also applies to all other methods (get, put, delete), just to prevent overriding the post parameters.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -125,10 +125,11 @@ export interface SpecConfig {
    * Possible values:
    *  - 'immediate' is overriding top level elements only thus you can not append a new path or alter an existing value without erasing same level elements.
    *  - 'recursive' proceed to a deep merge and will concat every branches or override or create new values if needed. @see https://www.npmjs.com/package/merge
+   *  - 'deepmerge' uses `deepmerge` to merge, which will concat object branches and concat arrays as well @see https://www.npmjs.com/package/deepmerge
    * The default is set to immediate so it is not breaking previous versions.
    * @default 'immediate'
    */
-  specMerging?: 'immediate' | 'recursive';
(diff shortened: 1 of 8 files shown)


WoH:
We may make this the default behavior in the next major version since it prevents a lot of confustion/inconvenience.

src/config.ts
```diff
@@ -125,10 +125,11 @@ export interface SpecConfig {
    * Possible values:
    *  - 'immediate' is overriding top level elements only thus you can not append a new path or alter an existing value without erasing same level elements.
    *  - 'recursive' proceed to a deep merge and will concat every branches or override or create new values if needed. @see https://www.npmjs.com/package/merge
+   *  - 'deepmerge' uses deepmerge to merge which will concat object branches and concat arrays aswell @see https://www.npmjs.com/package/deepmerge
```
endor, line 128: uses `deepmerge` to merge, which will concat object branches and concat arrays as well

endor:
I think it would be really helpful to add a unit test here to show how it works, since we're also referring to the unit tests as part of the documentation.

WoH:
> I think it would be really helpful to add a unit test here to show how it works, since we're also referring to the unit tests as part of the documentation.

Done

PR #767, opened 2020-08-03
Fixed @Response first parameter type

### LINKED ISSUE #119: @Response creates wrong swagger code
Hello,

after investigating an issue, if found out what's going wrong. There is a swagger validation error when using swagger-parser.

In one of our example we have:

`    @Response<ErrorResponseModel>('Unexpected error')`

This produces the follwing malformed code in swagger.json

```
"responses": {
	"200": {
		"description": "Ok",
		"schema": {
			"$ref": "#/definitions/UserResponseModel"
		}
	},
	"Unexpected error": {
		"description": "",
		"schema": {
			"$ref": "#/definitions/ErrorResponseModel"
		}
	}
},
```	

It should also create a 200 - as no status number is given - and 200 is the default:

```
"responses": {
	"200": {
		"description": "Ok",
		"schema": {
			"$ref": "#/definitions/UserResponseModel"
		}
	},
	"200": {
		"description": "Unexpected error",
		"schema": {
			"$ref": "#/definitions/ErrorResponseModel"
		}
	}
},
```

Everething works fine when using this:

`@Response<ErrorResponseModel>(400, 'Unexpected error')`

spaguette:
1. Accepts only valid HTTP status code in number or string format
2. Does not accept 200 or '200' as it's already given in the Route response

### All Submissions:

**Closing issues**

Closes #119 

1. Following the criteria described in the issue, I dropped the support of the `default` string in the first parameter of the decorator. This might be a regression though, and we might need to add it back explicitly.
2. Should we apply similar changes to the `@SuccessResponse`? If yes, which ones?
3. Not sure how to add the unit tests for this change, please help. On the other hand, should the TS type checking be enough?


packages/runtime/src/decorators/response.ts
```diff
@@ -1,10 +1,18 @@
...
   return () => {
     return;
   };
 }
 
-export function Response<T>(name: string | number, description?: string, example?: T): Function {
+export function Response<T>(
+  // The status code should not be allowed to be 200 or "200"` since that return type is already defined by the return of the route function.
+  // @see https://github.com/lukeautry/tsoa/issues/119
+  name: Exclude<HttpStatusCodeLiteral, 200> | Exclude<HttpStatusCodeLiteralStr, '200'>,
```
WoH, line 12: This does not work. OpenAPI allows for strings like `'default' | '4xx'` etc. This would break compatibility.

spaguette: I see. I will change it to `Exclude<HttpStatusCodeLiteral, 200> | Exclude<string, '200'>` in the next commit. WDYT?

WoH:
Sorry for the late response, I've tried to reason why/how we can accept the PR as I do not want to reject first time contributions usually, but I don't think we should break existing code over this. As explained, you can set the default res code to something else then 200, in which case `@Response<200, ...>` should not be prevented. I hope you understand and still are interested in contributing other PRs.

PR #803, opened 2020-10-01
feat: Add ability to fetch custom IoC container

### LINKED ISSUE #779: Allow routes to have a unique IoC container per request
#### **I'm submitting a ...**

  - [ ] bug report
  - [x] feature request
  - [ ] support request

#### I confirm that I
  - [x] used the [search](https://github.com/lukeautry/tsoa/search?type=Issues) to make sure that a similar issue hasn't already been submitted.

## Description

I'm currently working on a project where I am migrating from [inversify-express-utils](https://github.com/inversify/inversify-express-utils) to tsoa, and I'm looking to replicate some functionality.

In inversify-express-utils, each request gets it's own IoC container. This allows data (in my case an ID) to be bound to the container and then injected into services.

With tsoa, there is only one IoC container which is loaded from the user-defined IoC module and used in the routes:

```ts
const controller: any = iocContainer.get<SomeController>(SomeController);
```

With inversify-express-utils, the Express app is built with middleware that adds a child container to each request using Reflect and reflect-metadata.

```ts
this._app.all("*", (
    req: express.Request,
    res: express.Response,
    next: express.NextFunction
) => {
    (async () => {
        const httpContext = await _self._createHttpContext(req, res, next);
        Reflect.defineMetadata(
            METADATA_KEY.httpContext,
            httpContext,
            req
        );
        next();
    })();
});
```
<sup>https://github.com/inversify/inversify-express-utils/blob/bc40bb7671e0d4d8183266b3b3d563bb62056618/src/server.ts#L102</sup>

This is then extracted when handling the request and used to load the controller.

```ts
const httpContext = this._getHttpContext(req);
httpContext.container.bind<interfaces.HttpContext>(TYPE.HttpContext)
    .toConstantValue(httpContext);

// invoke controller's action
const value = await httpContext.container.getNamed<any>(TYPE.Controller, controllerName)[key](...args);
```
<sup>https://github.com/inversify/inversify-express-utils/blob/bc40bb7671e0d4d8183266b3b3d563bb62056618/src/server.ts#L248</sup>

Currently, I've implemented the middleware that adds the container to the request and I'm using a modified template to load that container out of the request so it can be used as the IoC container. 

Is this something tsoa would support?

If so I'm happy to get a pull request started. Below is the minimal change in my project in order to support this, but more work would need to be done for tsoa.

```diff
diff --git a/express.hbs b/routes-template.hbs
index 8956053..311086d 100755
--- a/express.hbs
+++ b/routes-template.hbs
@@ -14,6 +14,8 @@ import { {{name}} } from '{{modulePath}}';
 import { expressAuthentication } from '{{authenticationModule}}';
 {{/if}}
 {{#if iocModule}}
+import { METADATA_KEY } from './core/constant/keys';
+import { Container } from 'inversify';
 import { iocContainer } from '{{iocModule}}';
 {{/if}}
 import * as express from 'express';
@@ -76,7 +78,8 @@ export function RegisterRoutes(app: express.Express) {
             }

             {{#if ../../iocModule}}
-            const controller: any = iocContainer.get<{{../name}}>({{../name}});
+            const container: Container = Reflect.getMetadata(METADATA_KEY.httpContext, request).container
+            const controller: any = container.get<{{../name}}>({{../name}});
             if (typeof controller['setStatus'] === 'function') {
                 controller.setStatus(undefined);
             }
```

alxwrd:
Hey! Happy hacktoberfest 🎃! This closes #779.

This adds a check to see if `iocContainer` is a function and if it is, calls it with the request to resolve the IoC container. This means that the `iocContainer` can be created/updated/fetched per request.

The reason I've added in the check to see if it's a function is to avoid being a breaking change.

I will need to raise a PR over at [tsoa-community/docs](https://github.com/tsoa-community/docs) to document this change.

---

### DIFF
--- a/packages/cli/src/routeGeneration/templates/express.hbs
+++ b/packages/cli/src/routeGeneration/templates/express.hbs
@@ -11,6 +11,7 @@ import { expressAuthentication } from '{{authenticationModule}}';
 {{/if}}
 {{#if iocModule}}
 import { iocContainer } from '{{iocModule}}';
+import { TsoaIocContainer, TsoaIocContainerMethod } from '@tsoa/runtime';
 {{/if}}
 import * as express from 'express';
 
@@ -72,14 +73,9 @@ export function RegisterRoutes(app: express.Express) {
             }
 
             {{#if ../../iocModule}}
-            function isFunction(obj: any): obj is Function { return typeof obj === 'function'; }
-            interface Container<T> { get<T>(type: any): any; }
+            const container: TsoaIocContainer = typeof iocContainer === 'function' ? (iocContainer as TsoaIocContainerMethod)(request) : iocContainer;
 
-            let container: any = iocContainer;
-            if (isFunction(iocContainer)) {
-                container = iocContainer(request);
-            }
-            const controller: any = (container as Container<{{../name}}> ).get<{{../name}}>({{../name}});
+            const controller: any = container.get<{{../name}}>({{../name}});
             if (typeof controller['setStatus'] === 'function') {
                 controller.setStatus(undefined);
             }
--- a/packages/cli/src/routeGeneration/templates/express.hbs
+++ b/packages/cli/src/routeGeneration/templates/express.hbs
@@ -72,7 +72,14 @@ export function RegisterRoutes(app: express.Express) {
             }
 
             {{#if ../../iocModule}}
-            const controller: any = iocContainer.get<{{../name}}>({{../name}});
+            function isFunction(obj: any): obj is Function { return typeof obj === 'function'; }
+            interface Container<T> { get<T>(type: any): any; }
+
+            let container: any = iocContainer;
+            if (isFunction(iocContainer)) {
+                container = iocContainer(request);
+            }
+            const controller: any = (container as Container<{{../name}}> ).get<{{../name}}>({{../name}});
             if (typeof controller['setStatus'] === 'function') {
                 controller.setStatus(undefined);
             }
--- a/packages/cli/src/routeGeneration/templates/express.hbs
+++ b/packages/cli/src/routeGeneration/templates/express.hbs
@@ -11,7 +11,7 @@ import { expressAuthentication } from '{{authenticationModule}}';
 {{/if}}
 {{#if iocModule}}
 import { iocContainer } from '{{iocModule}}';
-import { TsoaIocContainer, TsoaIocContainerMethod } from '@tsoa/runtime';
+import { IocContainer, IocContainerFactory } from '@tsoa/runtime';
 {{/if}}
 import * as express from 'express';
 
@@ -73,7 +73,7 @@ export function RegisterRoutes(app: express.Express) {
             }
 
             {{#if ../../iocModule}}
-            const container: TsoaIocContainer = typeof iocContainer === 'function' ? (iocContainer as TsoaIocContainerMethod)(request) : iocContainer;
+            const container: IocContainer = typeof iocContainer === 'function' ? (iocContainer as IocContainerFactory)(request) : iocContainer;
 
             const controller: any = container.get<{{../name}}>({{../name}});
             if (typeof controller['setStatus'] === 'function') {
--- a/packages/runtime/src/interfaces/iocModule.ts
+++ b/packages/runtime/src/interfaces/iocModule.ts
@@ -1,5 +1,5 @@
-export interface TsoaIocContainer {
-  get<T>(controller: any): T;
+export interface IocContainer {
+  get<T>(controller: { prototype: T }): T;
 }
 
-export type TsoaIocContainerMethod = (request: any) => TsoaIocContainer;
+export type IocContainerFactory = (request: unknown) => IocContainer;
--- a/packages/runtime/src/interfaces/iocModule.ts
+++ b/packages/runtime/src/interfaces/iocModule.ts
@@ -1,0 +1,5 @@
+export interface TsoaIocContainer {
+  get<T>(controller: any): T;
+}
+
+export type TsoaIocContainerMethod = (request: any) => TsoaIocContainer;
--- a/tests/fixtures/inversify-dynamic-container/ioc.ts
+++ b/tests/fixtures/inversify-dynamic-container/ioc.ts
@@ -2,8 +2,9 @@ import e = require('express');
 import { Container } from 'inversify';
 import { ManagedController } from './managedController';
 import { ManagedService } from './managedService';
+import { IocContainerFactory, IocContainer } from '@tsoa/runtime';
 
-const iocContainer = function (request: e.Request): Container {
+const iocContainer: IocContainerFactory = function (request: e.Request): IocContainer {
   const container = new Container();
   container.bind<string>(Symbol.for('requestPath')).toConstantValue(request.path);
   container.bind<ManagedService>(ManagedService).to(ManagedService);
--- a/tests/fixtures/inversify-dynamic-container/ioc.ts
+++ b/tests/fixtures/inversify-dynamic-container/ioc.ts
@@ -1,0 +1,15 @@
+import { Container } from 'inversify';
+import { ManagedController } from './managedController';
+import { ManagedService } from './managedService';
+
+export let containerMethodCalled = false;
+
+const iocContainer = function (request: Express.Request): Container {
+  const container = new Container();
+  containerMethodCalled = true;
+  container.bind<ManagedService>(ManagedService).to(ManagedService).inSingletonScope();
+  container.bind<ManagedController>(ManagedController).to(ManagedController).inSingletonScope();
+  return container;
+};
+
+export { iocContainer };
--- a/tests/fixtures/inversify-dynamic-container/ioc.ts
+++ b/tests/fixtures/inversify-dynamic-container/ioc.ts
@@ -1,13 +1,12 @@
+import e = require('express');
 import { Container } from 'inversify';
 import { ManagedController } from './managedController';
 import { ManagedService } from './managedService';
 
-export let containerMethodCalled = false;
-
-const iocContainer = function (request: Express.Request): Container {
+const iocContainer = function (request: e.Request): Container {
   const container = new Container();
-  containerMethodCalled = true;
-  container.bind<ManagedService>(ManagedService).to(ManagedService).inSingletonScope();
+  container.bind<string>(Symbol.for('requestPath')).toConstantValue(request.path);
+  container.bind<ManagedService>(ManagedService).to(ManagedService);
   container.bind<ManagedController>(ManagedController).to(ManagedController).inSingletonScope();
   return container;
 };
--- a/tests/fixtures/inversify-dynamic-container/server.ts
+++ b/tests/fixtures/inversify-dynamic-container/server.ts
@@ -16,4 +16,4 @@ app.use((err: any, req: express.Request, res: express.Response, next: express.Ne
   res.status(err.status || 500).send(err.message || 'An error occurred during the request.');
 });
 
-app.listen(3004);
+app.listen();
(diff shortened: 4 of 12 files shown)


packages/cli/src/routeGeneration/templates/express.hbs
```diff
@@ -72,7 +72,14 @@ export function RegisterRoutes(app: express.Express) {
             }
 
             {{#if ../../iocModule}}
-            const controller: any = iocContainer.get<{{../name}}>({{../name}});
+            function isFunction(obj: any): obj is Function { return typeof obj === 'function'; }
```
alxwrd, line 75: I've had to add this function to satisfy TypeScript's [type-guards](https://stackoverflow.com/a/50371820/7220776).

In the case of
```ts
if (typeof iocContainer === 'function' {
  container = iocContainer(request);
}
```
TypeScript can't determine `iocContainer === 'function'`, so will mark `iocContainer(request)` as an error in the case it's not a function.

packages/cli/src/routeGeneration/templates/express.hbs
```diff
@@ -72,7 +72,14 @@ export function RegisterRoutes(app: express.Express) {
...
+            interface Container<T> { get<T>(type: any): any; }
```
alxwrd, line 76: Again, the `Container` interface here is to satisfy TypeScript. Without it, `container<Type>.get` has the error 'Untyped function calls may not accept type arguments.'

tests/fixtures/inversify-dynamic-container/server.ts
```diff
@@ -0,0 +1,19 @@
...
+app.use(bodyParser.json());
+app.use(methodOverride());
+RegisterRoutes(app);
+
+// It's important that this come after the main routes are registered
+app.use((err: any, req: express.Request, res: express.Response, next: express.NextFunction) => {
+  res.status(err.status || 500).send(err.message || 'An error occurred during the request.');
+});
+
+app.listen(3004);
```
WoH, line 19:
```suggestion
app.listen();
```

alxwrd: Ah! Would `app.listen(3010);`, or something, be better?

WoH: I think hardcoding ports is always a bad idea here, but the other servers do it aswell, so I can't fault you for setting one (one that's not used by another server).
If you omit the port, an available one will be picked.

tests/fixtures/inversify-dynamic-container/ioc.ts
```diff
@@ -0,0 +1,15 @@
+import { Container } from 'inversify';
+import { ManagedController } from './managedController';
+import { ManagedService } from './managedService';
+
+export let containerMethodCalled = false;
+
+const iocContainer = function (request: Express.Request): Container {
```
WoH, line 7: I think we should use the request and check that the request is actually used.

tests/fixtures/inversify-dynamic-container/server.ts
alxwrd: Honestly, I copied the fixture from the 'inversify' fixture. 😅. I'll accept the suggestion if the port can be omitted.

tests/fixtures/inversify-dynamic-container/ioc.ts
alxwrd: Sure! What were you thinking? Using the request to resolve a container or just a dummy test to ensure request is passed correctly?

WoH: Don't bind the ManagedService as a singleton, let the request be injected the req into the constructor, return the request's path from a getPath() method (instead of getModel). Something like that. Doesn't have to be the greatest use ever, but should show the idea and the funcitonality

packages/cli/src/routeGeneration/templates/express.hbs
```diff
@@ -72,7 +72,14 @@ export function RegisterRoutes(app: express.Express) {
...
+
+            let container: any = iocContainer;
+            if (isFunction(iocContainer)) {
+                container = iocContainer(request);
+            }
+            const controller: any = (container as Container<{{../name}}> ).get<{{../name}}>({{../name}});
```
WoH, line 82:
```suggestion
            let container: any = typeof iocContainer === 'function' ? iocContainer(request) : iocContainer;
            const controller: any = container.get<{{../name}}>({{../name}});
```

WoH, line 82: Looking back at the issue, the suggested API was container.get(name, request), not container(request).get(name).
What's the rationale for the change here>

alxwrd: Because the change is about fetching a container, it made more sense to be `container(request).get(name)`, as the `container()` method fetches the container, which we then call `.get()` on.

`container.get(name, request)` also didn't seem workable and would be a breaking change.

alxwrd: I think I can drop the `if` and `isFunction`, but without the `Container` interface there's still the 'Untyped function calls may not accept type arguments.' error.
```ts
interface Container<T> { get<T>(type: any): any; }

const container: any = typeof iocContainer === 'function' ? iocContainer(request) : iocContainer;
const controller: any = (container as Container<{{../name}}>).get<{{../name}}>({{../name}});
```

alxwrd:
```
fixtures/inversify/routes.ts:572:73 - error TS2349: This expression is not callable.
  Type 'never' has no call signatures.

572      const container: any = typeof iocContainer === 'function' ? iocContainer(request) : iocContainer;
```

I can get this to work by adding in the function signature in the case that it's not a function.

```ts
const container: any = typeof iocContainer === 'function' ?
    (iocContainer as (req:any) => any)(request) : iocContainer;
```

tests/fixtures/inversify-dynamic-container/ioc.ts
alxwrd: I've updated the test to inject the request path.

WoH:
Alright, I think I can accept this as a new way to create containers per request. A remaining DX issue is the typing thought. We should at least expose type information what the container can be (function request => Container or Container) and export an interface a developer can implement and be sure.

alxwrd:
I can add the following for typing information.

```ts
export interface TsoaIocContainer { get<T>(controller: unknown): unknown; }
export type TsoaIocContainerMethod = (request: any) => TsoaIocContainer
```

This would allow the users ioc module to type the method or container if they desired.
```ts
// src/ioc.ts
export const iocContainer: TsoaIocContainerMethod = (request: Express.Request): TsoaIocContainer => {
    return new Container();
};
```

Tsoa itself could then use these types in the routes file.
```ts
const container: TsoaIocContainer = typeof iocContainer === 'function' ?
    (iocContainer as TsoaIocContainerMethod)(request) : iocContainer;

const controller: any = container.get<{{../name}}>({{../name}});
```

Where would be best to expose these types?

alxwrd:
@WoH My last 2 commits should address the typing issue.

I have a change ready to add the typings the new test included in this PR, but would like the typings applied to the other tests as well?

WoH:
> @WoH My last 2 commits should address the typing issue.
> 
> I have a change ready to add the typings the new test included in this PR, but would like the typings applied to the other tests as well?

Ah I didn't properly explain. The interface is not exported, so, as a tsoa user, I can't import it (`import {Container} from 'tsoa'/'@tsoa/runtime'`). If you move the interface to the interfaces dir as export it, that's already most of what I was asking for.

WoH:
> @WoH My last 2 commits should address the typing issue.
> 
> I have a change ready to add the typings the new test included in this PR, but would like the typings applied to the other tests as well?

My bad, didn't notice the last commits.

packages/runtime/src/interfaces/iocModule.ts
```diff
@@ -0,0 +1,5 @@
+export interface TsoaIocContainer {
+  get<T>(controller: any): T;
+}
+
+export type TsoaIocContainerMethod = (request: any) => TsoaIocContainer;
```
WoH, line 5:
```suggestion
export type TsoaIocContainerFactory = (request: any) => TsoaIocContainer;
```

packages/runtime/src/interfaces/iocModule.ts
```diff
@@ -0,0 +1,5 @@
+export interface TsoaIocContainer {
```
WoH, line 1:
```suggestion
export interface IocContainer {
```

tests/fixtures/inversify-dynamic-container/ioc.ts
```diff
@@ -0,0 +1,14 @@
+import e = require('express');
+import { Container } from 'inversify';
+import { ManagedController } from './managedController';
+import { ManagedService } from './managedService';
+
+const iocContainer = function (request: e.Request): Container {
```
WoH, line 6: Use the new import

WoH:
Alright, this looks pretty good now, only some nitpicking left.

tests/fixtures/inversify-dynamic-container/ioc.ts
alxwrd: Will do! Do you think this is needed on all the tests?

WoH: If you use it for the ioc modules in /inversify and /inversify-cpg while you're at it, I wouldn't complain :smile:

packages/runtime/src/interfaces/iocModule.ts
```diff
@@ -0,0 +1,5 @@
...
+  get<T>(controller: any): T;
+}
```
WoH, line 3:
```suggestion
export interface IocContainer {
  get<T>(controller: { prototype: T }): T;
}
```

packages/runtime/src/interfaces/iocModule.ts
WoH, line 5:
```suggestion
export type IocContainerFactory = (request: unknown) => IocContainer;
```

alxwrd: I tried typing the request as unknow, but I get
```
Type '(request: e.Request) => IocContainer' is not assignable to type 'IocContainerFactory'.
  Types of parameters 'request' and 'request' are incompatible.
    Type 'unknown' is not assignable to type 'Request<ParamsDictionary>'.
```

tests/fixtures/inversify-dynamic-container/ioc.ts
alxwrd: I've just realised I can't type these tests because the containers use `.bind` and `.load` which aren't on the interface.

packages/runtime/src/interfaces/iocModule.ts
alxwrd: :+1:

packages/runtime/src/interfaces/iocModule.ts
WoH: > I tried typing the request as unknow, but I get
> 
> ```
> Type '(request: e.Request) => IocContainer' is not assignable to type 'IocContainerFactory'.
>   Types of parameters 'request' and 'request' are incompatible.
>     Type 'unknown' is not assignable to type 'Request<ParamsDictionary>'.
> ```

In the editor or with TS itself?
I can type the iocContainer as any one of these just fine:

```
const iocContainer = function (request: e.Request): IocContainer {
const iocContainer: IocContainerFactory = function (request: e.Request): IocContainer {
const iocContainer = function (request: e.Request) {
const iocContainer: IocContainerFactory = function (request: e.Request): IocContainer {
```

alxwrd: In vscode. I've committed `(request: unknown)` as the tests passed 🤷‍♂️.

WoH (approved):
LGTM

alxwrd:
hey @WoH. Thanks for your review and your help ironing this out. Do you need anything else from me, or can this be merged?

PR #814, opened 2020-10-15
express: RegisterRoutes: accept express.Router, not just express.Application

cakoose:
### All Submissions:

- [X] Have you written unit tests? **Modified an existing test.**
- [ ] Have you written unit tests that cover the negative cases (i.e.: if bad data is submitted, does the library respond properly)?  **Not applicable here?**

**Test plan**

Modified the express test to use an `express.Router` instance.  Test fails without the fix.

### DIFF
--- a/packages/cli/src/routeGeneration/templates/express.hbs
+++ b/packages/cli/src/routeGeneration/templates/express.hbs
@@ -45,7 +45,7 @@ const validationService = new ValidationService(models);
 
 // WARNING: This file was auto-generated with tsoa. Please do not modify it. Re-run tsoa to re-generate this file: https://github.com/lukeautry/tsoa
 
-export function RegisterRoutes(app: express.Express) {
+export function RegisterRoutes(app: express.Router) {
     // ###########################################################################################################
     //  NOTE: If you do not see routes for all of your controllers in this file, then you might not have informed tsoa of where to look
     //      Please look into the "controllerPathGlobs" config option described in the readme: https://github.com/lukeautry/tsoa
--- a/tests/fixtures/express-router/server.ts
+++ b/tests/fixtures/express-router/server.ts
@@ -1,0 +1,32 @@
+import * as bodyParser from 'body-parser';
+import * as express from 'express';
+import * as methodOverride from 'method-override';
+import '../controllers/rootController';
+
+import { RegisterRoutes } from './routes';
+
+export const app: express.Express = express();
+export const router = express.Router();
+app.use('/v1', router);
+router.use(bodyParser.urlencoded({ extended: true }));
+router.use(bodyParser.json());
+router.use(methodOverride());
+router.use((req: any, res: any, next: any) => {
+  req.stringValue = 'fancyStringForContext';
+  next();
+});
+RegisterRoutes(router);
+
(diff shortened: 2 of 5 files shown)


tests/fixtures/express/server.ts
```diff
@@ -21,14 +21,16 @@ import '../controllers/tagController';
 import { RegisterRoutes } from './routes';
 
 export const app: express.Express = express();
-app.use(bodyParser.urlencoded({ extended: true }));
-app.use(bodyParser.json());
-app.use(methodOverride());
-app.use((req: any, res: any, next: any) => {
+export const router = express.Router();
```
WoH, line 24: Move this to a new fixture, something like express-router/server.ts, leave the old one as is

packages/cli/src/routeGeneration/templates/express.hbs
```diff
@@ -45,7 +45,7 @@ const validationService = new ValidationService(models);
 
 // WARNING: This file was auto-generated with tsoa. Please do not modify it. Re-run tsoa to re-generate this file: https://github.com/lukeautry/tsoa
 
-export function RegisterRoutes(app: express.Express) {
+export function RegisterRoutes(app: express.Router) {
```
WoH, line 48:
```suggestion
export function RegisterRoutes(app: express.Router | express.Express) {
```

cakoose: `express.Express` implements `express.Application`, which implements `express.Router`.

tests/fixtures/express/server.ts
cakoose: My initial concern was that it adds a bunch of work to the test suite for a very small check.  For now I've created an "express-router" test case copied from the "express" test case.  But maybe it can be reduced?

Or maybe I can should just change the "express-openapi3" test to use `express.Router`?

WoH: The amount of specs is excessive, I agree. But we should have a separate fixture.
The specs should be something like [inversify-glob.spec.ts.](https://github.com/lukeautry/tsoa/blob/master/tests/integration/inversify-glob.spec.ts) Just 1-2 tests ensuring the route works.

tests/fixtures/express-router/server.ts
```diff
@@ -0,0 +1,47 @@
...
+import '../controllers/postController';
+import '../controllers/putController';
+
+import '../controllers/methodController';
+import '../controllers/parameterController';
+import '../controllers/securityController';
+import '../controllers/testController';
+import '../controllers/validateController';
+import '../controllers/exampleController';
+import '../controllers/tagController';
```
WoH, line 19: Import one controller, run 1-2 integration tests, done.

WoH:
Tests are failing b/c the new fixture is not registered [here](https://github.com/lukeautry/tsoa/blob/master/tests/prepare.ts#L130).

tests/fixtures/express/server.ts
WoH: If you revert the changes here, I  think we should be good to go.

cakoose: Ah whoops, sorry, not sure how I missed this.

PR #824, opened 2020-10-22
fix: conditional mapped type #823

### LINKED ISSUE #823: bug: conditional mapped type
Hello (I'm back), when we use conditional mapped types with tsoa, original values are written in the spec instead of updated one.

I'm working on a pull-request to fix it.

## Sorting

- **I'm submitting a ...**

  - [x] bug report
  - [ ] feature request
  - [ ] support request

- I confirm that I

## Expected Behavior

```yaml
Serialize__bar-foo.foo-string__:
        properties:
            bar:
                type: string
                enum:
                    - bar
                x-nullable: false
            foo:
                type: string
        required:
            - bar
            - foo
        type: object
```

```ts
type Serialize<T extends any> = {
  [key in keyof T]: T[key] extends 'foo' ? 'bar' :
                    T[key]
}

Serialize<{ bar: 'foo', foo: string }>;
```

## Current Behavior

```yaml
Serialize__bar-foo.foo-string__:
        properties:
            bar:
                type: string
                enum:
                    - foo
                x-nullable: false
            foo:
                type: string
        required:
            - bar
            - foo
        type: object
```

## Context (Environment)

Version of the library: v3.4.0
Version of NodeJS: v14.5.0

Eywek:
### All Submissions:

**Closing issues**

closes #823

### DIFF
--- a/packages/cli/src/metadataGeneration/typeResolver.ts
+++ b/packages/cli/src/metadataGeneration/typeResolver.ts
@@ -161,10 +161,14 @@ export class TypeResolver {
           const propertyType = typeChecker.getTypeOfSymbolAtLocation(property, this.typeNode);
           const declaration = getDeclaration(property) as ts.PropertySignature | ts.PropertyDeclaration | ts.ParameterDeclaration | undefined;
 
-          if (declaration && ts.isPropertySignature(declaration)) {
-            return this.propertyFromSignature(declaration, mappedTypeNode.questionToken);
-          } else if (declaration && (ts.isPropertyDeclaration(declaration) || ts.isParameter(declaration))) {
-            return this.propertyFromDeclaration(declaration, mappedTypeNode.questionToken);
+          // Note: for conditional type node, we want to resolve propertyType, because the declaration doesn't have the right type
+          // for conditional mapped type (see `TestModel.genericMappedType` in tests)
+          if (mappedTypeNode.type && !ts.isConditionalTypeNode(mappedTypeNode.type)) {
+            if (declaration && ts.isPropertySignature(declaration)) {
+              return this.propertyFromSignature(declaration, mappedTypeNode.questionToken);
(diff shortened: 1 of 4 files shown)


packages/cli/src/metadataGeneration/typeResolver.ts
```diff
@@ -161,14 +161,18 @@ export class TypeResolver {
           const propertyType = typeChecker.getTypeOfSymbolAtLocation(property, this.typeNode);
           const declaration = getDeclaration(property) as ts.PropertySignature | ts.PropertyDeclaration | ts.ParameterDeclaration | undefined;
 
-          if (declaration && ts.isPropertySignature(declaration)) {
-            return this.propertyFromSignature(declaration, mappedTypeNode.questionToken);
-          } else if (declaration && (ts.isPropertyDeclaration(declaration) || ts.isParameter(declaration))) {
-            return this.propertyFromDeclaration(declaration, mappedTypeNode.questionToken);
+          // Note: for conditional type node, we want to resolve propertyType, because the declaration doesn't have the right type
```
WoH, line 164: This was never a good idea, I'm just not sure what the best way to get the type of a property in a mapped type is.
Instead of piling up special cases, maybe it's a good time to see what the "proper" resolution would be.

Eywek: Do you have any leads? I haven't seen a magic way to make every cases works easily 😕  But I agree having special cases is not ideal

WoH: I haven't found a good way to get the Type, maybe we could try an approach where we add the Type Arguments to the Context + the iterator (`key` in #824 ) and resolve the TypeNode given the Context.

Eywek: I've tried but I haven't found anything that works, I will try to take time for this later, in the meantime, if you have time, I've allowed edits on this PR so maybe you can try too

Issue #723, opened 2020-06-05
Mechanism to avoid controller state and duplicating success annotations


devnev:
## Sorting

- **I'm submitting a ...**

  - [ ] bug report
  - [x] feature request
  - [ ] support request

- I confirm that I
  - [x] used the [search](https://github.com/lukeautry/tsoa/search?type=Issues) to make sure that a similar issue hasn't already been submit

## Expected Behavior

I can write handlers without storing request state on the controller, i.e. without the `.getStatus`/`.setStatus` methods, and without duplicating information in annotations.

## Current Behavior

If I use `@SuccessResponse` to describe the result of a handler, I still need to create a status state on the controller and `setStatus`/`getStatus`/`getHeaders` methods to have responses actually return the correct status code. To avoid having to add stateful (and possibly concurrency-unsafe) methods to a controller, I've instead opted to use the `@Res` decorator for all responses, including the final return.

e.g.

```ts
@Route('foo')
class MyController {
  @Get('bar')
  bar(
    @Res() created: TsoaResponse<201, {}>,
  ): Promise<void> {
    return created(201, {})
  }
}
```

However, doing this still produces a default success response in the spec:

```yaml
openapi: 3.0.0
paths:
    /foo/bar:
        get:
            operationId: Bar
            responses:
                '201':
                    description: ""
                    content:
                        application/json:
                            schema:
                                properties: {}
                                type: object
                '204':
                    description: 'No content'
```

To eliminate the extra response, I need to duplicate the success case in the annotations:

```ts
@Route('foo')
class MyController {
  @Get('bar')
  @SuccessResponse(201, 'created')
  bar(@Res() created: TsoaResponse<201, {}>): Promise<void> {
    return created(201, {});
  }
}
```

## Possible Solution

If the default response code in the generated template was derived from the success response annotation, I wouldn't need the `@Res` callback, and regular return would work fine. I don't know what happens if there's multiple `SuccessResponse` annotations though - maybe this could cause tsoa to error out?

Another alternative would be to avoid the default response being generated, although that raises the question of what the generated code should do when no response callback was called. Maybe generate an internal error?

## Context (Environment)

Version of the library: 3.1.0
Version of NodeJS: 10.19

- Confirm you were using yarn not npm: [x]

WoH:
I think the idea is good, but I'd rather approach it in a different way. Not using the default type annotation for the default response woud be a major change and in fact one that would hurt a lot of existing users.
[...]

PR #857, opened 2020-12-01
Feat: Response header(Fix #770)

### LINKED ISSUE #770: Support for response headers in generated OpenAPI spec
## Sorting

- **I'm submitting a ...**

  - [ ] bug report
  - [x] feature request
  - [ ] support request

- I confirm that I

## Expected Behavior

Using this kind of controller

```ts
interface ResponseHeaders {
  /**
   * Pagination uses Link header as defined in RFC5988
   */
  Link: string;
}

@Route("/")
export class FooController extends Controller {
  @Get()
  @SuccessHeaders<ResponseHeaders>()
  public async get(
    @Res() failed: TsoaResponse<400, string, ResponseHeaders>
  ): Promise<number> {
    if (false) {
      failed(400, "failure", { Link: "" });
    } else {
      this.setHeader("Link", '<https://localhost:3000/?page=2>; rel="next"');
      return 1;
    }
  }
}
```

It would be nice if the generated OpenAPI specification would have specifications for the response headers as well (as per https://swagger.io/specification/#response-object):

```yaml
"paths": {
  "/": {
    "get": {
      "operationId": "Get",
      "responses": {
        "200": {
          "description": "Ok",
          "content": {
            "application/json": {
              "schema": {
                "type": "number",
                "format": "double"
              }
            }
          },
          "headers": {
            "Link": {
              "description": "Pagination uses Link header as defined in RFC5988",
              "schema": {
                "type": "string"
              }
            }
          }
        },
        "400": {
          "description": "",
          "content": {
            "application/json": {
              "schema": {
                "type": "string"
              }
            }
          },
          "headers": {
            "Link": {
              "description": "Pagination uses Link header as defined in RFC5988",
              "schema": {
                "type": "string"
              }
            }
          }
        }
      },
      "security": [],
      "parameters": []
    }
  }
},
```

I presumed `@SuccessHeaders` is an upcoming feature from #723.

Also, in real life it doesn't make sense to include the link header in error response, but I included it here just to illustrate that `TsoaResponse` should have the same functionality.

## Current Behavior

Currently the generated specification has only the content schemas:

```yaml
"paths": {
  "/": {
    "get": {
      "operationId": "Get",
      "responses": {
        "200": {
          "description": "Ok",
          "content": {
            "application/json": {
              "schema": {
                "type": "number",
                "format": "double"
              }
            }
          }
        },
        "400": {
          "description": "",
          "content": {
            "application/json": {
              "schema": {
                "type": "string"
              }
            }
          }
        }
      },
      "security": [],
      "parameters": []
    }
  }
},
```

## Steps to Reproduce

Working example repository https://github.com/rhietala/tsoa-response-header

## Context (Environment)

Version of the library: 3.2.1
Version of NodeJS: 12.18.3

## Breaking change?

Probably not for `@SuccessHeaders` as is not in 3.x? For `TsoaResponse` probably not either, as this would only add the optional descriptions and spec generation as a new feature, but I don't know.

jackey8616:
### All Submissions:

**Closing issues**

Closes #770 .  

This RP introduce `SuccessResponse` and `Response` decorator with a generic type to define header struture.  
  
`SuccessResponse` wasn't follow with any generic type, so in this PR, is not a big deal.  
but `Response` is already provide a generic type T which specify the example structure,  
usage of `Response` before this PR is `@Response<ExampleClass>`:
In some situation like `need header class but not example class`, the decorator define will be `@Response<null, HeaderClass>`.  

**Test plan**

Test spec is coming out at path.[status code].headers field.

### DIFF
--- a/packages/cli/src/swagger/specGenerator2.ts
+++ b/packages/cli/src/swagger/specGenerator2.ts
@@ -202,7 +202,7 @@ export class SpecGenerator2 extends SpecGenerator {
         if (res.headers.dataType === 'refObject' || res.headers.dataType === 'nestedObjectLiteral') {
           res.headers.properties.forEach((each: Tsoa.Property) => {
             headers[each.name] = {
-              type: this.getSwaggerType(each.type).type,
+              ...this.getSwaggerType(each.type),
               description: each.description,
             };
           });
--- a/packages/cli/src/swagger/specGenerator2.ts
+++ b/packages/cli/src/swagger/specGenerator2.ts
@@ -184,17 +184,17 @@ export class SpecGenerator2 extends SpecGenerator {
   }
 
   protected buildOperation(controllerName: string, method: Tsoa.Method): Swagger.Operation {
-    const swaggerResponses: any = {};
+    const swaggerResponses: { [name: string]: Swagger.Response } = {};
 
     method.responses.forEach((res: Tsoa.Response) => {
       swaggerResponses[res.name] = {
         description: res.description,
       };
       if (res.schema && !isVoidType(res.schema)) {
-        swaggerResponses[res.name].schema = this.getSwaggerType(res.schema);
+        swaggerResponses[res.name].schema = this.getSwaggerType(res.schema) as Swagger.Schema;
       }
       if (res.examples && res.examples[0]) {
-        swaggerResponses[res.name].examples = { 'application/json': res.examples[0] };
+        swaggerResponses[res.name].examples = { 'application/json': res.examples[0] } as Swagger.Example;
       }
     });
 
--- a/packages/cli/src/swagger/specGenerator2.ts
+++ b/packages/cli/src/swagger/specGenerator2.ts
@@ -206,6 +206,8 @@ export class SpecGenerator2 extends SpecGenerator {
               description: each.description,
             };
           });
+        } else {
+          throw new Error('Header should be a class or nested object.');
         }
         swaggerResponses[res.name].headers = headers;
       }
--- a/packages/cli/src/swagger/specGenerator2.ts
+++ b/packages/cli/src/swagger/specGenerator2.ts
@@ -196,6 +196,19 @@ export class SpecGenerator2 extends SpecGenerator {
       if (res.examples && res.examples[0]) {
         swaggerResponses[res.name].examples = { 'application/json': res.examples[0] } as Swagger.Example;
       }
+
+      if (res.headers) {
+        const headers = {};
+        if (res.headers.dataType === 'refObject' || res.headers.dataType === 'nestedObjectLiteral') {
+          res.headers.properties.forEach((each: Tsoa.Property) => {
+            headers[each.name] = {
+              type: this.getSwaggerType(each.type).type,
+              description: each.description,
+            };
+          });
+        }
+        swaggerResponses[res.name].headers = headers;
+      }
     });
 
     return {
--- a/packages/cli/src/swagger/specGenerator2.ts
+++ b/packages/cli/src/swagger/specGenerator2.ts
@@ -207,7 +207,7 @@ export class SpecGenerator2 extends SpecGenerator {
             };
           });
         } else {
-          throw new Error('Header should be a class or nested object.');
+          assertNever(res.headers);
         }
         swaggerResponses[res.name].headers = headers;
       }
--- a/packages/cli/src/swagger/specGenerator3.ts
+++ b/packages/cli/src/swagger/specGenerator3.ts
@@ -314,6 +314,7 @@ export class SpecGenerator3 extends SpecGenerator {
             headers[each.name] = {
               schema: this.getSwaggerType(each.type),
               description: each.description,
+              required: each.required,
             };
           });
         }
--- a/packages/cli/src/swagger/specGenerator3.ts
+++ b/packages/cli/src/swagger/specGenerator3.ts
@@ -318,7 +318,7 @@ export class SpecGenerator3 extends SpecGenerator {
             };
           });
         } else {
-          throw new Error('Header should be a class or nested object.');
+          assertNever(res.headers);
         }
         swaggerResponses[res.name].headers = headers;
       }
--- a/packages/cli/src/swagger/specGenerator3.ts
+++ b/packages/cli/src/swagger/specGenerator3.ts
@@ -317,6 +317,8 @@ export class SpecGenerator3 extends SpecGenerator {
               required: each.required,
             };
           });
+        } else {
+          throw new Error('Header should be a class or nested object.');
         }
         swaggerResponses[res.name].headers = headers;
       }
--- a/packages/cli/src/swagger/specGenerator3.ts
+++ b/packages/cli/src/swagger/specGenerator3.ts
@@ -303,16 +303,16 @@ export class SpecGenerator3 extends SpecGenerator {
       }
 
       if (res.headers) {
-        const headers = {};
+        const headers: { [name: string]: Swagger.Header3 } = {};
         if (res.headers.dataType === 'refObject') {
           headers[res.headers.refName] = {
             schema: this.getSwaggerTypeForReferenceType(res.headers) as Swagger.Schema3,
             description: res.headers.description,
-          } as Swagger.Header3;
+          };
         } else if (res.headers.dataType === 'nestedObjectLiteral') {
           res.headers.properties.forEach((each: Tsoa.Property) => {
             headers[each.name] = {
-              schema: this.getSwaggerType(each.type),
+              schema: this.getSwaggerType(each.type) as Swagger.Schema3,
               description: each.description,
               required: each.required,
             };
--- a/packages/cli/src/swagger/specGenerator3.ts
+++ b/packages/cli/src/swagger/specGenerator3.ts
@@ -301,6 +301,24 @@ export class SpecGenerator3 extends SpecGenerator {
           (swaggerResponses[res.name].content || {})['application/json']['examples'] = examples;
         }
       }
+
+      if (res.headers) {
+        const headers = {};
+        if (res.headers.dataType === 'refObject') {
+          headers[res.headers.refName] = {
+            schema: this.getSwaggerTypeForReferenceType(res.headers) as Swagger.Schema3,
+            description: res.headers.description,
+          } as Swagger.Header3;
+        } else if (res.headers.dataType === 'nestedObjectLiteral') {
+          res.headers.properties.forEach((each: Tsoa.Property) => {
+            headers[each.name] = {
+              schema: this.getSwaggerType(each.type),
+              description: each.description,
+            };
+          });
+        }
+        swaggerResponses[res.name].headers = headers;
+      }
     });
 
     return {
--- a/packages/runtime/src/decorators/response.ts
+++ b/packages/runtime/src/decorators/response.ts
@@ -1,22 +1,10 @@
-export function SuccessResponse(name: string | number, description?: string): Function;
-
-export function SuccessResponse<T>(name: string | number, description?: string): Function;
-
-export function SuccessResponse<T>(name: string | number, description?: string): Function {
+export function SuccessResponse<HeaderType = {}>(name: string | number, description?: string): Function {
   return () => {
     return;
   };
 }
 
-export function Response(name: string | number, description?: string): Function;
-
-export function Response<U>(name: string | number, description?: string): Function;
-
-export function Response<T>(name: string | number, description?: string, example?: T): Function;
-
-export function Response<T, U>(name: string | number, description?: string, example?: T): Function;
-
-export function Response<T, U>(name: string | number, description?: string, example?: T): Function {
+export function Response<ExampleType, HeaderType = {}>(name: string | number, description?: string, example?: ExampleType): Function {
   return () => {
     return;
   };
--- a/packages/runtime/src/decorators/response.ts
+++ b/packages/runtime/src/decorators/response.ts
@@ -1,10 +1,22 @@
-export function SuccessResponse(name: string | number, description?: string): Function {
+export function SuccessResponse(name: string | number, description?: string): Function;
+
+export function SuccessResponse<T>(name: string | number, description?: string): Function;
+
+export function SuccessResponse<T>(name: string | number, description?: string): Function {
   return () => {
     return;
   };
 }
 
-export function Response<T>(name: string | number, description?: string, example?: T): Function {
+export function Response(name: string | number, description?: string): Function;
+
+export function Response<U>(name: string | number, description?: string): Function;
+
+export function Response<T>(name: string | number, description?: string, example?: T): Function;
+
+export function Response<T, U>(name: string | number, description?: string, example?: T): Function;
+
+export function Response<T, U>(name: string | number, description?: string, example?: T): Function {
   return () => {
     return;
   };
--- a/packages/runtime/src/decorators/response.ts
+++ b/packages/runtime/src/decorators/response.ts
@@ -1,10 +1,12 @@
-export function SuccessResponse<HeaderType = {}>(name: string | number, description?: string): Function {
+import { IsValidHeader } from '../utils/isHeaderType';
+
+export function SuccessResponse<HeaderType extends IsValidHeader<HeaderType> = {}>(name: string | number, description?: string): Function {
   return () => {
     return;
   };
 }
 
-export function Response<ExampleType, HeaderType = {}>(name: string | number, description?: string, example?: ExampleType): Function {
+export function Response<ExampleType, HeaderType extends IsValidHeader<HeaderType> = {}>(name: string | number, description?: string, example?: ExampleType): Function {
   return () => {
     return;
   };
--- a/tests/fixtures/controllers/responseHeaderController.ts
+++ b/tests/fixtures/controllers/responseHeaderController.ts
@@ -1,7 +1,7 @@
 import { Controller, Get, Route, SuccessResponse, Response, TsoaResponse, Res } from '@tsoa/runtime';
 
 /**
- * asdasdsda
+ * response header's description
  */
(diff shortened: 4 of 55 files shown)


WoH:
Can you extend the parsing to the third argument of `@Res() Res: TsoaResponse<Status, Body, Headers>`. I added the argument for type checking and anticipation of a PR adding this.

jackey8616:
@WoH I've done the support for TsoaResponse,  
Please review. thank you.

tests/unit/swagger/schemaDetails3.spec.ts
```diff
@@ -430,6 +430,31 @@ describe('Definition generation for OpenAPI 3.0.0', () => {
...
+        });
+
+        it('should generate headers from class response decorator.', () => {
+          const metadata = new MetadataGenerator('./fixtures/controllers/commonResponseHeaderController.ts').Generate();
+          const responseSpec = new SpecGenerator3(metadata, getDefaultExtendedOptions()).GetSpec();
+          const paths = ['Response1', 'Response2'];
+          paths.forEach((path: string) => {
+            const responses = responseSpec.paths[`/CommonResponseHeader/${path}`].get?.responses;
+
+            expect(responses?.[200]?.headers?.CommonResponseHeader).to.not.eq(undefined);
```
WoH, line 454: Let's not just test for existence, but for content.

tests/unit/swagger/schemaDetails3.spec.ts
```diff
@@ -430,6 +430,31 @@ describe('Definition generation for OpenAPI 3.0.0', () => {
...
+        it('should generate headers from method reponse decorator.', () => {
+          const metadata = new MetadataGenerator('./fixtures/controllers/responseHeaderController.ts').Generate();
+          const responseSpec = new SpecGenerator3(metadata, getDefaultExtendedOptions()).GetSpec();
+          const paths = ['SuccessResponseWithHeaderClass', 'SuccessResponseWithObject', 'ResponseWithHeaderClass', 'ResponseWithObject', 'TsoaResponseWithHeaderClass', 'TsoaResponseWithObject'];
+          paths.forEach((path: string) => {
+            const responses = responseSpec.paths[`/ResponseHeader/${path}`].get?.responses;
+
+            expect(responses?.[200]?.headers).to.not.eq(undefined);
+            if (path.includes('HeaderClass')) {
+              expect(responses?.[200]?.headers?.ResponseHeader).to.not.eq(undefined);
```
WoH, line 442: Let's not just test for existence, but for content.

tests/unit/swagger/schemaDetails.spec.ts
```diff
@@ -183,6 +183,49 @@ describe('Schema details generation', () => {
...
+          const paths = ['SuccessResponseWithHeaderClass', 'SuccessResponseWithObject', 'ResponseWithHeaderClass', 'ResponseWithObject', 'TsoaResponseWithHeaderClass', 'TsoaResponseWithObject'];
+          paths.forEach((path: string) => {
+            const responses = responseSpec.paths[`/ResponseHeader/${path}`].get?.responses;
+
+            expect(responses?.[200]?.headers).to.not.eq(undefined);
+            if (path.includes('HeaderClass')) {
+              expect(responses?.[200]?.headers).to.deep.eq({
+                Link: {
+                  type: 'string',
+                  description: 'a link string',
```
WoH, line 198: Should have `required: true`.

tests/fixtures/controllers/responseHeaderController.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  @Get('TsoaResponseWithHeaderClass')
+  public async handler5(@Res() res: TsoaResponse<200, null, ResponseHeader>): Promise<void> {
+    return;
+  }
+
+  @Get('TsoaResponseWithObject')
+  public async handler6(@Res() res: TsoaResponse<200, null, { linkE: string; linkF: number }>): Promise<void> {
+    return;
+  }
```
WoH, line 52:
```ts
interface ComplexHeader {
  /**
   * Numbers description
   * @example [1,2,3]
   * @default [1,2,3]
   * @isInt
   */
  nums: number[];
}


  @Get('TsoaResponseWithHeaderArray')
  public async handler7(@Res() res: TsoaResponse<200, string, ComplexHeader[]>) {
    return res(200, 'Hello, World', [{ nums: [1, 2, 3] }]);
  }
```

So we can test that all the possible header object fields (using the style + schema, which is what req headers do): 
- description
- required
- ~~deprecated~~ (does not exist on property yet, so omit that for now)
- schema
- style (this should be static for all headers)
- allowReserved this should be static for all headers)
- ~~example(s)~~ (should be in the schema's example prop)

packages/cli/src/swagger/specGenerator3.ts
```diff
@@ -301,6 +301,24 @@ export class SpecGenerator3 extends SpecGenerator {
...
+      if (res.headers) {
+        const headers = {};
+        if (res.headers.dataType === 'refObject') {
+          headers[res.headers.refName] = {
+            schema: this.getSwaggerTypeForReferenceType(res.headers) as Swagger.Schema3,
+            description: res.headers.description,
+          } as Swagger.Header3;
+        } else if (res.headers.dataType === 'nestedObjectLiteral') {
+          res.headers.properties.forEach((each: Tsoa.Property) => {
+            headers[each.name] = {
```
WoH, line 314: Document `required` etc. which exist on the property.

packages/runtime/src/decorators/response.ts
```diff
@@ -1,10 +1,22 @@
...
-export function Response<T>(name: string | number, description?: string, example?: T): Function {
+export function Response(name: string | number, description?: string): Function;
+
+export function Response<U>(name: string | number, description?: string): Function;
+
+export function Response<T>(name: string | number, description?: string, example?: T): Function;
+
+export function Response<T, U>(name: string | number, description?: string, example?: T): Function;
+
+export function Response<T, U>(name: string | number, description?: string, example?: T): Function {
```
WoH, line 19: 1. Prefer more helpful names (Schema, Headers) over generic T, U
2. Add some restrictions for the Header type

packages/runtime/src/decorators/response.ts
```diff
@@ -1,10 +1,22 @@
-export function SuccessResponse(name: string | number, description?: string): Function {
+export function SuccessResponse(name: string | number, description?: string): Function;
+
+export function SuccessResponse<T>(name: string | number, description?: string): Function;
+
+export function SuccessResponse<T>(name: string | number, description?: string): Function {
```
WoH, line 5: See https://github.com/lukeautry/tsoa/pull/857/files#r549009809

packages/cli/src/swagger/specGenerator2.ts
```diff
@@ -184,17 +184,30 @@ export class SpecGenerator2 extends SpecGenerator {
...
+        swaggerResponses[res.name].schema = this.getSwaggerType(res.schema) as Swagger.Schema;
       }
       if (res.examples && res.examples[0]) {
-        swaggerResponses[res.name].examples = { 'application/json': res.examples[0] };
+        swaggerResponses[res.name].examples = { 'application/json': res.examples[0] } as Swagger.Example;
+      }
+
+      if (res.headers) {
+        const headers = {};
+        if (res.headers.dataType === 'refObject' || res.headers.dataType === 'nestedObjectLiteral') {
```
WoH, line 202: Add else branch that throws to prevent silent failures.

packages/cli/src/swagger/specGenerator3.ts
```diff
@@ -301,6 +301,24 @@ export class SpecGenerator3 extends SpecGenerator {
...
+              schema: this.getSwaggerType(each.type),
+              description: each.description,
+            };
+          });
+        }
```
WoH, line 319: Add else to prevent silent failure

packages/runtime/src/decorators/response.ts
WoH, line 19: I think the goal here is to provide options, but I'm afraid these are too many.
This will cover what we need in a way that's not ambiguous:

```ts
export function SuccessResponse<HeaderType>(name: string | number, description?: string): Function {
  return () => {
    return;
  };
}

export function Response<ExampleType, Header = {}>(name: string | number, description?: string, example?: ExampleType): Function {
  return () => {
    return;
  };
}
```

Otherwise, it looks like this:

```ts
export function Response<U>(name: string | number, description?: string): Function;

export function Response<T>(name: string | number, description?: string, example?: T): Function;
```

means the argument can be only the Header type, but the parsing will just fo the wrong thing.

tests/unit/swagger/schemaDetails.spec.ts
jackey8616: There is no `required` field in [HeaderObject at OpenAPI2.0](https://swagger.io/specification/v2#header-object)

packages/runtime/src/decorators/response.ts
jackey8616: @WoH sorry, I don't quite understanding this statements.  
```ts
Header extends keyof Header extends string ? unknown : 'Header Type must only have string keys';
```
Can you provide some example, thank you!

tests/fixtures/controllers/responseHeaderController.ts
jackey8616: @WoH You mean we can make multiple different header class(or field inside header class) to test all of these features with proper test cases?

tests/unit/swagger/schemaDetails.spec.ts
WoH: You're right, this was supposed to a comment related to the OAS3 specs, which are now properly asserting this.

tests/fixtures/controllers/responseHeaderController.ts
WoH: As long as every one of these properties is tested once, we are good imo.
The only remaining properties are `style` and `allowReserved`, which we should set to provide the best spec when it comes to returning header properties of `string[]`.

packages/runtime/src/decorators/response.ts
WoH:
```ts
export type TsoaResponse<T extends HttpStatusCodeLiteral, U, Header extends keyof Header extends string ? unknown : 'Header Type must only have string keys';> = (status: T, data: U, headers?: V) => any;

```

That said, I think it makes sense to also restrict the values to `string | string[]`, having to deal with anything else is a headache.

WoH:
2 thoughts:
- We may want to add type aliases in the future, currently, that will fail
- This relates to both https://github.com/lukeautry/tsoa/pull/850 and #770 which initially suggested the Header docs API: Currently, this PR extends the `@SuccessResponse` decorator arguments in a descriptive fashion. This means, there's no guarantee that the response headers actually match the documentation. #850 changes that for the response status, I'd argue there's a lot of times, where the header could be set using `@SuccessHeader({'Cache-Control', 'none'})`. What do you think? Would also be interested in @devnev 's take here. While I don't think we should do that in this PR, if we anticipate that functionality, we should change the descriptive API to not add an Argument to `@SuccessResponse` and instead add `@SuccessHeader`.

WoH:
@jackey8616 I'd love to add some commits that address some of the things I laid out myself if you don't mind. Would you like to give me commit access to the PR or should I send you diffs?

jackey8616:
@WoH I turned it on!

packages/runtime/src/decorators/response.ts
WoH: Applied this now. While your initial approach would probably work, I looked at the header typings and the consensus (tsoa, express) usually is that values must be strings, so I applied that.

Maybe we can loosen that w/ regards to numbers, i.e. fastify types headers as

```ts
  header(key: string, value: any): FastifyReply<RawServer, RawRequest, RawReply, RouteGeneric, ContextConfig>;
  headers(values: {[key: string]: any}): FastifyReply<RawServer, RawRequest, RawReply, RouteGeneric, ContextConfig>;
  getHeader(key: string): string | undefined;
  getHeaders(): {
    // Node's `getHeaders()` can return numbers and arrays, so they're included here as possible types.
    [key: string]: number | string | string[] | undefined;
  };
```

Since it's all a string after returning, I think a string with pattern probably is a more precise spec. Let me know what you think.

WoH:
Remaining TODOs:

- [ ] OAS3: `explode: true`
- [ ] Clean up tests: Both OAS2 and OAS3 specs should test all the new features, there are discrepancies rn
- [ ] Discuss `@SuccessResponse<Header>()` vs. `SuccessHeader<Header>()`

jackey8616:
@WoH 
> @jackey8616 was there a specific reason why there's a deviation from the API suggested in #770?

IMHO, header must be used along with a response in swagger, thus, they should be good if we let it be optional parameter at `Response`, `SuccessResponse` or `TsoaResponse`.

Since `SuccessResponse` and `SuccessHeader` decorator is implicitly contains a 2xx status.
When programmer not using default status as it decorator gives, these two decorator may let the usage logic become fuzzy.

WoH:
> Since SuccessResponse and SuccessHeader decorator is implicitly contains a 2xx status.
> When programmer not using default status as it decorator gives, these two decorator may let the usage logic become fuzzy.

I don't think I can follow. Can you give me an example?

jackey8616:
> > Since SuccessResponse and SuccessHeader decorator is implicitly contains a 2xx status.
> > When programmer not using default status as it decorator gives, these two decorator may let the usage logic become fuzzy.
> 
> I don't think I can follow. Can you give me an example?

If we want to let success status code to be 204,  
we need to set both in `SuccessResponse` & `SuccessHeader`.  
and it seems a little bit bother to set 204 code twice.  
```ts
@SuccessResponse(204, 'some descriptions doesn\'t want be omit.')
@SuccessHeader<HeaderClass>(204, {
  linkA: 'http://google.com',
})
public async handler(): Promise<void> {
  return;
}
```

WoH:
> > > Since SuccessResponse and SuccessHeader decorator is implicitly contains a 2xx status.
> > > When programmer not using default status as it decorator gives, these two decorator may let the usage logic become fuzzy.
> > 
> > I don't think I can follow. Can you give me an example?
> 
> If we want to let success status code to be 204,  
> we need to set both in `SuccessResponse` & `SuccessHeader`.  
> and it seems a little bit bother to set 204 code twice.  
> ```ts
> @SuccessResponse(204, 'some descriptions doesn\'t want be omit.')
> @SuccessHeader<HeaderClass>(204, {
>   linkA: 'http://google.com',
> })
> public async handler(): Promise<void> {
>   return;
> }
> ```

But do we? What's the problem with 

```ts
​@​SuccessResponse​(​204​,​ ​'some descriptions doesn\'t want be omit.'​)​
@​SuccessHeader​<​Header>({
  ​linkA​: ​'http://google.com'​,​
​}​)​
​public​ ​async​ ​handler​(​)​: ​Promise​<​void​>​ ​{​
  ​return​;​
​}
```

jackey8616:
> But do we? What's the problem with 
> 
> ```ts
> ​@​SuccessResponse​(​204​,​ ​'some descriptions doesn\'t want be omit.'​)​
> @​SuccessHeader​<​Header>({
>   ​linkA​: ​'http://google.com'​,​
> ​}​)​
> ​public​ ​async​ ​handler​(​)​: ​Promise​<​void​>​ ​{​
>   ​return​;​
> ​}
> ```
I think its no problem expect it runs different logic compares with SuccessResponse, which requires pre-check exists response defined by it, further more, it also effect by existence of SuccessResponse decorator, which may cause complex logic of SuccessHeader IMHO.

I've surveyed #723, and I don't see any actual reason to introduce SuccessHeader.

WoH:
> I think its no problem expect it runs different logic compares with SuccessResponse, which requires pre-check exists response defined by it, further more, it also effect by existence of SuccessResponse decorator, which may cause complex logic of SuccessHeader IMHO.
> 
> I've surveyed #723, and I don't see any actual reason to introduce SuccessHeader.


I hope I get where you're coming from now. For this PR, the goal is only to document the headers on the default / success response. Admittedly, you can do that both with the API you implemented and `@SuccessHeaders<Header>()>`.
However, pulling out the header portion into it's own decorator allows us to extend the usage more easily in the future. That requires 2 additional steps, one of which will be already done in #850, because it allows tsoa to know what the default/success status code is. Based on this, we are able to infer when a default response is returned and are able to append the headers as provided in an argument to `@SuccessHeaders({})`.

I'm looking specifically at this: 

```ts
@​SuccessHeader​<​Header>({
  ​linkA​: ​'http://google.com'​,​
​}​)​
​public​ ​async​ ​handler​(​)​: ​Promise​<​void​>​ ​{​
  ​return​;​
​}
```

Here, `Header` is the type used for documentation, the actual argument will be set as headers automatically.
If we use the `@SuccessResponse`, my worry was that this would be something like:

```ts
@​SuccessResponse<​Header>(undefined, undefined, {
  ​linkA​: ​'http://google.com'​,​
​}​)​
​public​ ​async​ ​handler​(​)​: ​Promise​<​void​>​ ​{​
  ​return​;​
​}
```

However, we can obviously do type/arg checks so we can do:

```ts
@​SuccessResponse<​Header>({
  ​linkA​: ​'http://google.com'​,​
​}​)​
​public​ ​async​ ​handler​(​)​: ​Promise​<​void​>​ ​{​
  ​return​;​
​}
```

So it seems to come down to preference. For my personal taste setting the headers as a separate annotation feels more intuitive to me, but if I can't really point a finger as to why, your approach is probably equally fine.

devnev:
Should the solution for success headers generalise to headers with dynamic values?

WoH:
@jackey8616 I rebased and this lgtm now. Can you take a look and confirm?

WoH:
> Should the solution for success headers generalise to headers with dynamic values?

You'd still have to set them using `setHeaders`, but you'd be able to declare them (as optional properties on the Header interface).

PR #866, opened 2020-12-09
Validate path variables. Fixes #865

### LINKED ISSUE #865: Validate that path variables are formatted correctly
## Sorting

- **I'm submitting a ...**

  - [ ] bug report
  - [x] feature request
  - [ ] support request

- I confirm that I

## Expected Behavior
1. Create a route `@Get('/comments}')`
2. Expect tsoa route generation to fail because `}` is an invalid closing of a path parameter

## Current Behavior
- tsoa route generation succeeds and generates (for express):
```ts
app.get('/comments',
```
- the trailing `}` is auto-stripped during route generation here: https://github.com/lukeautry/tsoa/blob/227f73b8131271dac84bd02a88a4a2e66317cc8a/packages/cli/src/module/generate-routes.ts#L25
- swagger generation does _not_ strip the trailing `}`, so swagger generates a different route:
```yaml
paths:
    '/comments}':
        get:
```

## Possible Solution
1. Users could validate this themselves
    - As a tsoa user, I'd prefer this baked in because tsoa is already scanning and processing all annotations
2. In `generate-routes` `pathTransformer`, throw an error if brackets are invalid
    - Could use a library like [parenthesis](https://www.npmjs.com/package/parenthesis), but that's likely overkill

## Steps to Reproduce
- See above

## Context (Environment)

Version of the library: 3.4.0
Version of NodeJS: 12

## Breaking change?
- Yes. tsoa would start throwing for anyone that currently has stray `{` or `}` in their routes

aldenquimby:
### All Submissions:

**Closing issues**
Closes #865

- Breaking change. Would any users rely on `{` or `}` being part of their path?

**Test plan**
- Unit tests show `{param}` -> `:param` replacement still works, and that errant `{` or `}` now throw

### DIFF
--- a/packages/cli/src/module/generate-routes.ts
+++ b/packages/cli/src/module/generate-routes.ts
@@ -4,7 +4,7 @@ import { ExtendedRoutesConfig } from '../cli';
 import { MetadataGenerator } from '../metadataGeneration/metadataGenerator';
 import { Tsoa } from '@tsoa/runtime';
 import { RouteGenerator } from '../routeGeneration/routeGenerator';
-import { convertBracketPathParams } from '../utils/pathUtils';
+import { convertBracesPathParams } from '../utils/pathUtils';
 
 export const generateRoutes = async (
   routesConfig: ExtendedRoutesConfig,
@@ -21,7 +21,7 @@ export const generateRoutes = async (
 
   const routeGenerator = new RouteGenerator(metadata, routesConfig);
 
-  let pathTransformer = convertBracketPathParams;
+  let pathTransformer = convertBracesPathParams;
   let template;
 
   switch (routesConfig.middleware) {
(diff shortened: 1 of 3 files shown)


WoH:
Should route/swagger behave differently? Obviously not. However, I'd say that's for now mainly a bug with the square bracket conversation. 
I'm don't know what each of the possibly underlying frameworks do with square brackets, I'd hope they auto escape/accept the raw and the escaped form. 

The current implementation is already too aggressively replacing things, so I'd rather not do more of that.

aldenquimby:
@WoH good point. I updated the implementation to match swagger. So instead of throwing if there are floating `{` or `}` in a path, route generation just won't replace them.

Note - I also renamed from `brackets []` to `braces {}` because I was using the wrong term

WoH:
LGTM, thanks a lot.

PR #898, opened 2021-01-28
docs(README): fix broken link to API reference

tsimbalar:
it was pointing to `https://tsoa-community.github.io/reference/globals.html` which does not exist

(removed the PR template as this is just a docs update, and does not need tests)


WoH:
Good catch, thanks!

PR #905, opened 2021-02-04
fix(cli): allow returning 0 from controllers

### LINKED ISSUE #889: Wrong return
## Wrong response statusCode and body

- **I'm submitting a ...**

  - [x] bug report
  - [ ] feature request
  - [ ] support request

- I confirm that I

## Expected Behavior
I need to return number 0 (this show count of bonuses on user wallet).
example code in controller: 
const response = 0 
return response

i must get 0, and statusCode 200

## Current Behavior

But now i get empty response with statusCode 204

## Possible Solution

change response generator 
res || "empty response with 204 status"
to
res ?? || "empty response with 204 status"

## Steps to Reproduce

1) create controller
2) create @Get endpoint ''
``` 
@Get('/test')
async getZero(): Promise<number> {
  return 0
} 
```

## Context (Environment)

Version of the library: 
Version of NodeJS: 

fahrradflucht:
Closes #889

Instead of checking for falsyness and guarding for `false` explicitly check if a controller result is `null` or `undefined`, set status code and data based on that. This allows to return `false`, `0` and `''`(empty string).

### DIFF
--- a/packages/cli/src/routeGeneration/templates/express.hbs
+++ b/packages/cli/src/routeGeneration/templates/express.hbs
@@ -174,7 +174,7 @@ export function RegisterRoutes(app: express.Router) {
         });
         if (data && typeof data.pipe === 'function' && data.readable && typeof data._read === 'function') {
             data.pipe(response);
-        } else if (data || data === false) { // === false allows boolean result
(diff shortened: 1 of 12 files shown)


WoH:
LGTM, thank you!

Issue #859, opened 2020-12-02
[BUG] - Header already sent


dobobaie:
Lines to update : 
[packages/cli/src/routeGeneration/templates/express.hbs#L152](https://github.com/lukeautry/tsoa/blob/8606da96ccc71608b0c07c8cd8384fcb7813d820/packages/cli/src/routeGeneration/templates/express.hbs#L152)
[packages/cli/src/routeGeneration/templates/hapi.hbs#L165](https://github.com/lukeautry/tsoa/blob/8606da96ccc71608b0c07c8cd8384fcb7813d820/packages/cli/src/routeGeneration/templates/hapi.hbs#L165)
[packages/cli/src/routeGeneration/templates/koa.hbs#L160](https://github.com/lukeautry/tsoa/blob/8606da96ccc71608b0c07c8cd8384fcb7813d820/packages/cli/src/routeGeneration/templates/koa.hbs#L160)

## Sorting

- **I'm submitting a ...**

  - [x] bug report
  - [ ] feature request
  - [ ] support request

- I confirm that I
  - [x] used the [search](https://github.com/lukeautry/tsoa/search?type=Issues) to make sure that a similar issue hasn't already been submit

## Steps to Reproduce

- Use express as framework
- Install `connect-timeout` or similar package
- Add the middleware with `5s` of timeout
- Add a new controller and put the code bellow in your class
```js
  @Get('list')
  public async retrieveList(): Promise<any> {
    await new Promise((resolve) =>
      setTimeout(() => resolve(), 10000)
    );
    return { status: 'ok' };
  }
```
- Request on the route and watch the console

## Detailed Description

if any external middleware send a response before the request is done then we got the error bellow : 

``` bash
Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client
    at ServerResponse.setHeader (_http_outgoing.js:547:11)
``` 

## Expected Behavior

```bash
Nothing
```

## Current Behavior

``` bash
Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client
    at ServerResponse.setHeader (_http_outgoing.js:547:11)
``` 

## Possible Solution

Add a verification to know is the request is already sent

Maybe
```js
response.on('close', () => ...);
```

dobobaie:
Any update?

WoH:
Feel free to submit a PR that adds this check

dobobaie:
[...]

PR #908, opened 2021-02-09
Don't modify the response after headers were sent

### LINKED ISSUE #904: How to properly set response statusCode for "Typechecked Alternative Responses"?
I am using "Typechecked Alternative Responses" and enabled logging with [morgan](https://github.com/expressjs/morgan) middleware and am struggling to sync the status code observed by the user that receives the response and what I observe in the server logs.
Status code is the same when I don't use any async/await calls, the proper status code that I defined is observed both in the response that the user receives and in the server logs.
As soon as I add any async/await call, the user still receives proper statusCode, but in the server logs, I observe default success response status code (200/204).

## Sorting

- **I'm submitting a ...**

  - [ ] bug report
  - [ ] feature request
  - [x] support request

- I confirm that I

## Expected Behavior

Response status code has the same value in the response sent to the user and logged with the middleware.

## Current Behavior

I observe different statusCode reported to the user and different statusCode in the server logs (using morgan) when the controller contains async/await calls.

## Steps to Reproduce

1. Setup simple server e.g. according to https://tsoa-community.github.io/docs/getting-started.html
2. Add morgan middleware to the app
```ts
import morgan from 'morgan';
/*
...
*/
app.use(morgan('dev'));
```
3. Change `UsersController` to:
```ts
@Route("users")
export class UsersController {
  @Get()
  public async getUser(
    @Res() res: TsoaResponse<400, string>,
  ): Promise<User> {
      return res(400, 'data');
  }
}
```
4. Run the server and send the request, client properly receives response 400 and the same status is logged in the server logs by morgan
5. Now change the controller to
```ts
@Route("users")
export class UsersController {
  @Get()
  public async getUser(
    @Res() res: TsoaResponse<400, string>,
  ): Promise<User> {
      await new Promise(resolve => setTimeout(resolve, 10));
      return res(400, 'data');
  }
}
```
in the controller before the return.
6. Run the server and send the request, client properly receives response 400, but now the logger logs 204 (default success response) in the server logs.

What am I missing here?

## Context (Environment)

Version of the library: 3.5.1
Version of NodeJS: 14.15.0

### LINKED ISSUE #907: Is there a way to handle multiple http methods as one function in tsoa controller?
## Sorting

- **I'm submitting a ...**

  - [ ] bug report
  - [x] feature request
  - [ ] support request

- I confirm that I

## Expected Behavior

I want to write code like:

```ts
@Post("/proxy")
@Get("/proxy")
@Put("/proxy")
@Delete("/proxy")
public async proxy(
  @Request req: Request,
): Promise<any> {
  const modifiedReq = makeModifiedReq(req)
  return proxyToAnotherServer(modifiedReq.url, modifiedReq.method, modifiedReq.query, modifiedReq.body)
}

// or 

@Multiple(["Get","Post","Put","Delete"], "/proxy")
public async proxy(
  @Request req: Request,
): Promise<any> {
  const modifiedReq = makeModifiedReq(req)
  return proxyToAnotherServer(modifiedReq.url, modifiedReq.method, modifiedReq.query, modifiedReq.body)
}
```

## Current Behavior

I have to write repetitive code.

```ts
@Post("/proxy")
public async proxyPost(
  @Request req: Request,
): Promise<any> {
  const modifiedReq = makeModifiedReq(req)
  return proxyToAnotherServer(modifiedReq.url, modifiedReq.method, modifiedReq.query, modifiedReq.body)
}

@Get("/proxy")
public async proxyGet(
  @Request req: Request,
): Promise<any> {
  const modifiedReq = makeModifiedReq(req)
  return proxyToAnotherServer(modifiedReq.url, modifiedReq.method, modifiedReq.query, modifiedReq.body)
}

@Put("/proxy")
public async proxyPut(
  @Request req: Request,
): Promise<any> {
  const modifiedReq = makeModifiedReq(req)
  return proxyToAnotherServer(modifiedReq.url, modifiedReq.method, modifiedReq.query, modifiedReq.body)
}

@Delete("/proxy")
public async proxyDelete(
  @Request req: Request,
): Promise<any> {
  const modifiedReq = makeModifiedReq(req)
  return proxyToAnotherServer(modifiedReq.url, modifiedReq.method, modifiedReq.query, modifiedReq.body)
}

```

## Detailed Description

**Is there a way to process multiple http methods as one function in tsoa controller?**

szymon-planeta:
### All Submissions:

**Closing issues**

closes #904  #859 

enhancement discussed in #907 #859

1. Couldn't find anything similar to `headersSent` for the hapi and wasn't sure how to add same check for this framework. Maybe it's because of the request lifecycle specific for the hapi? Would need someone familiar with hapi to comment.
2. Issues with testing (see below)

**Test plan**

I have added integration test with multiple `@Res`, but it's not really correct in its current state.
- for hapi and koa, the test already succeeds because of the `__isTsoaResponded` checks
- for express I can see the exception ```Error [ERR_HTTP_HEADERS_SENT]``` in the test logs (if the check is not there), but this exception does not cause the test to break (since the client receives proper response). How do I catch that exception in the test and cause it to fail?

### DIFF
--- a/packages/cli/src/routeGeneration/templates/express.hbs
+++ b/packages/cli/src/routeGeneration/templates/express.hbs
@@ -169,6 +169,9 @@ export function RegisterRoutes(app: express.Router) {
     // WARNING: This file was auto-generated with tsoa. Please do not modify it. Re-run tsoa to re-generate this file: https://github.com/lukeautry/tsoa
 
     function returnHandler(response: any, statusCode?: number, data?: any, headers: any = {}) {
+        if (response.headersSent) {
+            return;
+        }
         Object.keys(headers).forEach((name: string) => {
             response.set(name, headers[name]);
         });
--- a/packages/cli/src/routeGeneration/templates/koa.hbs
+++ b/packages/cli/src/routeGeneration/templates/koa.hbs
@@ -181,7 +181,7 @@ export function RegisterRoutes(router: KoaRouter) {
     // WARNING: This file was auto-generated with tsoa. Please do not modify it. Re-run tsoa to re-generate this file: https://github.com/lukeautry/tsoa
 
     function returnHandler(context: any, next: () => any, statusCode?: number, data?: any, headers: any={}) {
-        if (!context.response.__tsoaResponded) {
+        if (!context.headerSent && !context.response.__tsoaResponded) {
             context.set(headers);
 
             if (data !== null && data !== undefined) { 
--- a/tests/fixtures/controllers/getController.ts
+++ b/tests/fixtures/controllers/getController.ts
@@ -225,6 +225,19 @@ export class GetTestController extends Controller {
   public async getRes(@Res() res: TsoaResponse<400, TestModel, { 'custom-header': string }>): Promise<void> {
     res?.(400, new ModelService().getModel(), { 'custom-header': 'hello' });
   }
+
+  /**
+   * @param res The alternate response
+   * @param res Another alternate response
+   */
+  @Get('MultipleRes')
+  public async multipleRes(@Res() res: TsoaResponse<400, TestModel, { 'custom-header': string }>, @Res() anotherRes: TsoaResponse<401, TestModel, { 'custom-header': string }>): Promise<Result> {
+    res?.(400, new ModelService().getModel(), { 'custom-header': 'hello' });
+    anotherRes?.(401, new ModelService().getModel(), { 'custom-header': 'another hello' });
+    return {
+      value: 'success',
+    };
+  }
 }
 
 export interface ErrorResponse {
--- a/tests/integration/express-server.spec.ts
+++ b/tests/integration/express-server.spec.ts
@@ -138,6 +138,18 @@ describe('Express Server', () => {
     );
   });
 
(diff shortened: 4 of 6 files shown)


szymon-planeta:
any ideas on how to proceed with that?

WoH:
You might want to insert a console spy and make sure there was an error before, which does not happen with your PR.

See https://github.com/lukeautry/tsoa/blob/master/tests/unit/swagger/pathGeneration/routeDuplicates.spec.ts

szymon-planeta:
@WoH sorry but I'm still unsure how to test it...
The error happens on the server side - if I'm trying to test it with the integration test I guess I cannot do that, since I'm only verifying what was returned from the server? Even the error itself in the console log is visible after the test has already succeeded.

I think this could be tested with a unit test (just as the example that you linked) for the controller, but as I understand from contribution guidelines, writing a unit test for the generated controller might be troublesome?

WoH:
You're not verifying what the server returns, you're already doing that. What you need to add is a console spy that matches the error message that would show up and makes sure there's no error about the headers being printed to the console.

Issue #730, opened 2020-06-09
Should tsoa require typescript as a runtime (non-dev) dependency?


tavishf:
## Sorting

- **I'm submitting a ...**

  - [ ] bug report
  - [ ] feature request
  - [x] support request

- I confirm that I
  - [x] used the [search](https://github.com/lukeautry/tsoa/search?type=Issues) to make sure that a similar issue hasn't already been submit

## Expected Behavior

A project with tsoa as a dev dependency only should run fine when built via npm ci --production.

_At least, this is what I would intuitively expect, but this might be totally wrong. It might be that it's actually intended for it to be a full runtime dependency, in which case this issue makes no sense. But, it seemed surpising/unusual that it needed to be a runtime dependency to me._

## Current Behavior

App with typescript as a dev dependency only, when built with npm ci --production fails at runtime because typescript-formatter has a peer dependency on typescript, and it's not there.

## Possible Solution

Not sure, don't know enough about why it's a dependency to comment.

## Steps to Reproduce

Not sure if this happens 100% of the time or just with the specific way we're using it - will provide more info if this is in fact not intentional behaviour, otherwise there's no point. :)

## Context (Environment)

Version of the library: 3.1.1
Version of NodeJS: 12.13.1

- Confirm you were using yarn not npm: [ ]

## Detailed Description

Maybe no change, this might be entirely intended!

## Breaking change?

Probably not?

kdeenanauth:
It seems intentional for the time being - see #690

IanStorm:
As the related PR #690 is still under discussion, this issue should not be marked as stale from my perspective.
But I am not sure how this is usually being handled. What do you think?

PR #915, opened 2021-02-17
Fix @tsoa/runtime's package.json for pnpm by moving avoiding mention of ts.CompilerOptions in Config

kibiz0r:
### All Submissions:

  - Related: #730 and #690
  - Sort of (#730), but that issue was closed and it was also made before tsoa was split into separate packages

*Edit: Changed approach to avoid referencing typescript-provided types at all.*

Original approach below:

> Sounds like the explicit goal of #690 was to exclude `typescript` as a prod dependency for the runtime.
>
> But if you use pnpm and TypeScript together, it becomes apparent that the runtime *does* in fact depend on `typescript` for its public interface, so it won't compile without the change in this PR.
>
> npm users sneak past this issue because it doesn't isolate dependencies the way pnpm does, so they're able to leverage their dev dependencies even while doing prod builds, and since this is just some config thing tucked away somewhere instead of a deep dependency on calling a function in `typescript`, probably nobody has seen any runtime errors from this issue.

Relevant code from `@tsoa/runtime`:

```ts
// @tsoa/runtime/dist/config.d.ts
...
export interface Config {
    ...
    compilerOptions?: ts.CompilerOptions;
}
```

Example project demonstrating the error: https://github.com/kibiz0r/tsoa-typescript-dep-bug-example

The example uses regular npm, to prove it's not just pnpm being weird, but that meant I had to reference `@tsoa/runtime` instead of `tsoa` in order to trigger the behavior that pnpm triggers simply by referencing `tsoa`.

```
yarn run v1.22.4
$ tsc
node_modules/@tsoa/runtime/dist/config.d.ts:1:21 - error TS2307: Cannot find module 'typescript' or its corresponding type declarations.

1 import * as ts from 'typescript';
                      ~~~~~~~~~~~~
```

So two options here are:
1. Accept the prod dependency
2. Avoid the reference to typescript in the public API

I assume the authors prefer option 2, so they can indeed omit `typescript` from `@tsoa/runtime` actual (not declared) dependencies.

~~But that's a design change, where this PR is just a change to make the declared dependencies match the actual dependencies, thus making strict package managers like pnpm work with tsoa as it is designed today.~~

Edit: Going with option 2.

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -74,8 +74,8 @@ const getConfig = async (configPath = 'tsoa.json'): Promise<Config> => {
   return config;
 };
 
-const validateCompilerOptions = (config?: ts.CompilerOptions): ts.CompilerOptions => {
-  return config || {};
+const validateCompilerOptions = (config?: Record<string, unknown>): ts.CompilerOptions => {
+  return (config || {}) as ts.CompilerOptions;
 };
 
 export interface ExtendedSpecConfig extends SpecConfig {
--- a/packages/runtime/src/config.ts
+++ b/packages/runtime/src/config.ts
@@ -1,4 +1,3 @@
-import * as ts from 'typescript';
 import { Swagger } from './swagger/swagger';
 
 export interface Config {
@@ -35,10 +34,10 @@ export interface Config {
   /**
    * Typescript CompilerOptions to be used during generation
    *
-   * @type {ts.CompilerOptions}
+   * @type {Record<string, unknown>}
    * @memberof RoutesConfig
    */
-  compilerOptions?: ts.CompilerOptions;
+  compilerOptions?: Record<string, unknown>;
 }
 
 /**


WoH:
I agree that this issue should not be happening, but I'd prefer to not require typescript as a runtime dependency to pull in one interface if we can avoid it.

kibiz0r:
Wanna just make it an `any` then? Simplest path forward.

Also, for anyone who just needs to get past this today, something like this should work (slightly modified from our real code, so I apologize if you have to tweak it a bit):

```js
// save this as pnpmfile.js in your root and then re-create your lockfile
const alreadyWarnedAboutMissingDependencies = [];

const missingDependencyFixes = {
  "@tsoa/runtime": {
    expectedVersion: "3.5.2",
    missingDependencies: {
      "typescript": "^4.1.2"
    },
  }
};

function addMissingDependenciesFor(pkg) {
  const fixes = missingDependencyFixes[pkg.name];

  if (!fixes) {
    return;
  }

  if (fixes.expectedVersion == pkg.version) {
    if (!alreadyWarnedAboutMissingDependencies.includes(pkg.name)) {
      console.warn(
        `adding missing dependencies to ${pkg.name}: ${JSON.stringify(
          fixes.missingDependencies
        )}`
      );
      alreadyWarnedAboutMissingDependencies.push(pkg.name);
    }

    for (missingDependencyName in fixes.missingDependencies) {
      const missingDependencyVersion =
        fixes.missingDependencies[missingDependencyName];
      pkg.dependencies[missingDependencyName] = missingDependencyVersion;
    }
  } else {
    throw new Error(
      `package ${pkg.name} has been updated from ${fixes.expectedVersion} to ${pkg.version}; please check whether dependency fixes in pnpmfile.js are still necessary; if so, update expectedVersion`
    );
  }
}

module.exports = {
  hooks: {
    readPackage(pkg) {
      addMissingDependenciesFor(pkg);
      return pkg;
    },
  },
};
```

WoH:
Let's type it as `Record<string, unknown` and cast it to `CompilerOptions` in `@tsoa/cli`.

kibiz0r:
@WoH: Updated code (and PR description) to use the `Record<string, unknown>` approach instead.

Wish I could change the branch name now. 😆

PR #925, opened 2021-02-26
Added a noWriteIfUnchanged option to avoid writing routes file if content is unchanged

kibiz0r:
This PR adds a `routes.noWriteIfUnchanged` option to `tsoa.json`, which avoids writing to the filesystem (thus changing the mtime) if the `routes.ts` content would remain the same.

### Motivation

Using `nodemon` or `tsc -w` to generate routes when src changes and then compile src when the routes change can get noisy and cause unnecessary compilation.

If you're throwing the routes file in with the rest of your source, you can't even generate the routes and use `tsc -w` side-by-side, because it causes an infinite loop.

Using `nodemon` to compile, `--ignore` avoids that issue, but `--delay` isn't a great way to avoid a second round of compilation because you have to tune it to your generation time -- you're either wasting time waiting too long to avoid a second compile, or wasting time doing a second compile because you didn't wait long enough.

### Example

```json
// tsoa.json
{
  "entryFile": "src/app.ts",
  "noImplicitAdditionalProperties": "throw-on-extras",
  "controllerPathGlobs": ["src/**/*.controller.ts"],
  "spec": {
    "outputDirectory": "spec",
    "specVersion": 3
  },
  "routes": {
    "routesDir": "src",
    "iocModule": "src/di/inversify.config.ts",
    "noWriteIfUnchanged": true
  }
}
```

### All Submissions:

**Closing issues**

**Test plan**

### DIFF
--- a/packages/cli/src/routeGeneration/routeGenerator.ts
+++ b/packages/cli/src/routeGeneration/routeGenerator.ts
@@ -3,7 +3,7 @@ import * as handlebars from 'handlebars';
 import * as path from 'path';
 import { ExtendedRoutesConfig } from '../cli';
 import { Tsoa, TsoaRoute, assertNever } from '@tsoa/runtime';
-import { fsReadFile, fsWriteFile } from '../utils/fs';
+import { fsReadFile, fsWriteFile, fsExists } from '../utils/fs';
 import { isRefType } from '../utils/internalTypeGuards';
 import { normalisePath } from './../utils/pathUtils';
 
@@ -19,6 +19,15 @@ export class RouteGenerator {
 
     const fileName = `${this.options.routesDir}/${this.options.routesFileName || 'routes.ts'}`;
     const content = this.buildContent(middlewareTemplate, pathTransformer);
+
+    if (this.options.noWriteIfUnchanged) {
+      if (await fsExists(fileName)) {
+        const existingContent = (await fsReadFile(fileName)).toString();
+        if (content === existingContent) {
+          return;
+        }
+      }
+    }
 
     await fsWriteFile(fileName, content);
   }
--- a/packages/runtime/src/config.ts
+++ b/packages/runtime/src/config.ts
@@ -177,6 +177,11 @@ export interface RoutesConfig {
   routesFileName?: string;
 
   /**
+   * Avoid writing the generate route file if the existing file is identical (useful to optimize watch processes); false by default
+   */
+  noWriteIfUnchanged?: boolean;
+
+  /**
    * Base API path; e.g. the '/v1' in https://myapi.com/v1
    */
   basePath?: string;
--- a/packages/runtime/src/config.ts
+++ b/packages/runtime/src/config.ts
@@ -177,7 +177,7 @@ export interface RoutesConfig {
   routesFileName?: string;
 
   /**
-   * Avoid writing the generate route file if the existing file is identical (useful to optimize watch processes); false by default
+   * Avoid writing the generated route file if the existing file is identical (useful to optimize watch processes); false by default
    */
   noWriteIfUnchanged?: boolean;
 


WoH:
Initially, my reaction to this feature would be to wrap the exported cli method, but I understand that this method will already write to disk. Therefore, I'd rather export a method extracted from the route generation that returns the content of the routes file
https://github.com/lukeautry/tsoa/blob/bc621b45b1b24bc4a6bb5c86e1e38b54cf8058d2/packages/cli/src/routeGeneration/routeGenerator.ts#L21
and does not write to disk. 

Comparing the entire content also does not seem like a performance award winner.

> Using nodemon to compile, --ignore avoids that issue, but --delay isn't a great way to avoid a second round of compilation because you have to tune it to your generation time -- you're either wasting time waiting too long to avoid a second compile, or wasting time doing a second compile because you didn't wait long enough.

I agree, don't do any of that. If you're looking to save time, build the routes on change and then use tsc.

kibiz0r:
> Comparing the entire content also does not seem like a performance award winner.

Agreed. I want to do something like this:

```ts
const existingContent = await fsReadFile(filename); // => Buffer
if (Buffer.equals(newContent)) {
  ...
}
```

...but if newContent is a string, that doesn't work, and in order to make it a buffer we have to copy the entire string anyway, defeating the point.

kibiz0r:
I tried my best at hand-rolling a string-vs-buffer comparison, and did some performance comparisons:

Times are in seconds, running `time node <script>`, where each script is comparing a 50,000,000-character in-memory string against a 50,000,000-character file.

Some scripts need to do a bit more or less work to prepare the in-memory string, so it's not apples-to-apples across scenarios, but we care about comparing across *approaches*, not scenarios.

Approach | Entirely random content | Entirely same content | Same content up to the 50% mark, then random
--- | --- | --- | ---
Compare string and buffer | 0.361 | 1.109 | 1.260
Copy string to buffer | 0.526 | 0.334 | 0.386
Read buffer to string | 0.395 | 0.275 | 0.299

Source files: https://gist.github.com/kibiz0r/3eca0b3e330aa414dd83288bfa5cab10

Output:

```
# random content
harrinm@C02D11F7MD6T:~/git/tsoa|option-to-avoid-writing-unchanged-routes-file> time node compareStringAndBuffer.js
equal? false
node compareStringAndBuffer.js  0.23s user 0.12s system 100% cpu 0.348 total
harrinm@C02D11F7MD6T:~/git/tsoa|option-to-avoid-writing-unchanged-routes-file> time node copyToBuffer.js
equal? false
node copyToBuffer.js  0.33s user 0.14s system 97% cpu 0.481 total
harrinm@C02D11F7MD6T:~/git/tsoa|option-to-avoid-writing-unchanged-routes-file> time node readToString.js
equal? false
node readToString.js  0.25s user 0.14s system 99% cpu 0.390 total
```

```
# same content
harrinm@C02D11F7MD6T:~/git/tsoa|option-to-avoid-writing-unchanged-routes-file> time node compareStringAndBufferSame.js
equal? true
node compareStringAndBufferSame.js  0.98s user 0.15s system 100% cpu 1.123 total
harrinm@C02D11F7MD6T:~/git/tsoa|option-to-avoid-writing-unchanged-routes-file> time node copyToBufferSame.js
equal? true
node copyToBufferSame.js  0.19s user 0.14s system 100% cpu 0.332 total
harrinm@C02D11F7MD6T:~/git/tsoa|option-to-avoid-writing-unchanged-routes-file> time node readToStringSame.js
equal? true
node readToStringSame.js  0.15s user 0.14s system 101% cpu 0.288 total
```

```
# same content up to 50% mark
harrinm@C02D11F7MD6T:~/git/tsoa|option-to-avoid-writing-unchanged-routes-file> time node compareStringAndBufferSameHalfway.js
equal? false
node compareStringAndBufferSameHalfway.js  0.47s user 0.10s system 99% cpu 0.577 total
harrinm@C02D11F7MD6T:~/git/tsoa|option-to-avoid-writing-unchanged-routes-file> time node copyToBufferSameHalfway.js
equal? false
node copyToBufferSameHalfway.js  0.26s user 0.12s system 99% cpu 0.387 total
harrinm@C02D11F7MD6T:~/git/tsoa|option-to-avoid-writing-unchanged-routes-file> time node readToStringSameHalfway.js
equal? false
node readToStringSameHalfway.js  0.18s user 0.12s system 97% cpu 0.312 total
```

kibiz0r:
Looks like it's probably as good as we're gonna get it, performance-wise, without doing something crazy. You have to opt into this behavior, so it's not like existing users are gonna be paying an overhead in exchange for nothing.

WoH:
I was thinking about md5 hashes

kibiz0r:
Tried with `md5` library: https://www.npmjs.com/package/md5, which was horrible. On the order of tens of seconds instead of hundreds of milliseconds.

Tried again with `hasha`: https://www.npmjs.com/package/hasha, which was much better, but still not a winner.

I think what we're seeing here is that JS just doesn't offer a whole lot of escape hatches where you can get native performance on a call. Blitting and string comparisons are a couple of places where you can avoid VM overhead.

Updated table:

Approach | Entirely random content | Entirely same content | Same content up to the 50% mark, then random
--- | --- | --- | ---
Compare string and buffer | 0.348 | 1.123 | 0.577
Copy string to buffer | 0.481 | 0.332 | 0.387
Read buffer to string | 0.390 | 0.288 | 0.312
`md5` | 17.717 | 13.412 | 13.706
`hasha` | 0.692 | 0.459 | 0.518

```
# hasha
harrinm@C02D11F7MD6T:~/git/tsoa|option-to-avoid-writing-unchanged-routes-file> time node compareMD5.js
equal? false
node compareMD5.js  0.54s user 0.15s system 99% cpu 0.692 total
harrinm@C02D11F7MD6T:~/git/tsoa|option-to-avoid-writing-unchanged-routes-file> time node compareMD5Same.js
equal? true
node compareMD5Same.js  0.32s user 0.14s system 99% cpu 0.459 total
harrinm@C02D11F7MD6T:~/git/tsoa|option-to-avoid-writing-unchanged-routes-file> time node compareMD5SameHalfway.js
equal? false
node compareMD5SameHalfway.js  0.40s user 0.12s system 99% cpu 0.518 total

# md5
harrinm@C02D11F7MD6T:~/git/tsoa|option-to-avoid-writing-unchanged-routes-file> time node compareMD5.js
equal? false
node compareMD5.js  16.20s user 1.86s system 101% cpu 17.717 total
harrinm@C02D11F7MD6T:~/git/tsoa|option-to-avoid-writing-unchanged-routes-file> time node compareMD5Same.js
equal? true
node compareMD5Same.js  12.42s user 1.39s system 102% cpu 13.412 total
harrinm@C02D11F7MD6T:~/git/tsoa|option-to-avoid-writing-unchanged-routes-file> time node compareMD5SameHalfway.js
equal? false
node compareMD5SameHalfway.js  12.74s user 1.35s system 102% cpu 13.706 total
```

kibiz0r:
Correction: The files are 100-million characters, cuz it's 50-million *hex pairs*.

So we're talking about a 100-MB routes.ts.

I also checked to see how much of that is just JS startup. A `doNothing.js` which just loads those requires, defines the comparison function, and does a console.log takes 0.063s. So a lot of the timings here are just sheer JS VM.

And a script which just prepares the in-memory string before doing the file-read and equality-check takes 0.148s.

So if we wanna just talk about the file-read and equality-check, we're talking.... 0.288 - 0.148 = 0.140s?

So 140ms overhead on a 100-MB routes.ts.

kibiz0r:
@WoH: I think this is as optimized as we're likely to get here. How do you feel about going forward with this?

WoH:
> @WoH: I think this is as optimized as we're likely to get here. How do you feel about going forward with this?

I think we can optimize later, that's not my main issue. The reason I have not merged this yet was:

> Initially, my reaction to this feature would be to wrap the exported cli method, but I understand that this method will already write to disk. Therefore, I'd rather export a method extracted from the route generation that returns the content of the routes file.

That said, I've come around to believe it's reasonable to merge this as it stands now, as most of the issues related to the output of the routes file should be solved through a custom template.

rosslavery:
Just came across this project and in getting set up, this one bit us. Because we're using `tsc-watch + node` instead of `ts-node + nodemon`.

Compiling programmatically helped, but now we're in the "infinite loop" scenario that @kibiz0r described.

Any chance you can cut an npm release with this commit in it? Would be a huge help!

PR #932, opened 2021-03-06
Added stricter types for @Response decorator

### LINKED ISSUE #119: @Response creates wrong swagger code

spaguette:
Accepts only valid HTTP status code in number or string format (plus additional supported in OpenAPI) for the @Response decorator

### All Submissions:

Closes #119 

1. Not sure how to add the unit tests for this change. On the other hand, should the TS type checking be enough?


packages/runtime/src/interfaces/response.ts
```diff
@@ -60,4 +60,66 @@ export type HttpStatusCodeLiteral =
   | 510
   | 511;
 
+export type HttpStatusCodeLiteralStr =
```
WoH, line 63: Let's use a template literal type here: 

```ts
type HttpStatusCodeString = `${HttpStatusCode}`;
```

spaguette: Didn't realize it would work like this 😅

packages/runtime/src/interfaces/response.ts
```diff
@@ -60,4 +60,8 @@ export type HttpStatusCodeLiteral =
   | 510
   | 511;
 
+export type HttpStatusCodeLiteralStr = `${HttpStatusCodeLiteral}`;
```
WoH, line 63:
```suggestion
export type HttpStatusCodeStringLiteral = `${HttpStatusCodeLiteral}`;
```

WoH:
LGTM, thanks

Issue #938, opened 2021-03-14
Using 'useSuccessResponseCode' doesn't set response code


brian-pinwheel:
The code change in PR #850 was awesome.  

## Sorting

- **I'm submitting a ...**

  - [ X] bug report
  - [ ] feature request
  - [ ] support request

- I confirm that I
  - [X ] used the [search](https://github.com/lukeautry/tsoa/search?type=Issues) to make sure that a similar issue hasn't already been submit

## Expected Behavior

When using SuccessResponse, it should return the response defined in the SuccessResponse.

## Current Behavior

For performance, I am using a singleton Controller.  My create method needs to return a 201, but since the status is shared across all methods, the 201 was "leaking" to other requests.

When I found the feature to use SuccessResponse, I thought it solved my problems.  Unfortunately, it didn't. When I got it enabled, I saw the routes.js was generated successfully.  Unfortunately, the value was being overwritten by an undefined value.  

Specifically, the if statement for isController is executed because I am using the controller.  Hence, the value from my SuccessResponse is passed into the method correctly, but overwritten by an undefined value.

```
            let statusCode = successStatus;
            let headers;
            if (isController(controllerObj)) {
                headers = controllerObj.getHeaders();
                statusCode = controllerObj.getStatus();
            }
```

## Possible Solution

Don't overwrite the value if the status from the controller.getStatus is undefined.  This makes existing behavior work along with it working as expected (passing the SuccessResponse)

## Context (Environment)

Version of the library: 3.5.2
Version of NodeJS: 12

- Confirm you were using yarn not npm: [ X]

## Breaking change?

Should not break

WoH:
Closed by #939

PR #939, opened 2021-03-14
Don't overwrite status codes from SuccessResponse if the controller's…

brian-pinwheel:
… status is undefined

This is to address Issue #938 

### All Submissions:

- [ X] Have you followed the guidelines in our [Contributing](https://github.com/lukeautry/tsoa/tree/master/docs/CONTRIBUTING.md) document?
- [ X] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/lukeautry/tsoa/pulls) for the same update/change?
- [ X ] This PR is associated with an existing issue?

**Closing issues**

None

### DIFF
--- a/packages/cli/src/routeGeneration/templates/express.hbs
+++ b/packages/cli/src/routeGeneration/templates/express.hbs
@@ -166,7 +166,7 @@ export function RegisterRoutes(app: express.Router) {
                 let headers;
                 if (isController(controllerObj)) {
                     headers = controllerObj.getHeaders();
-                    statusCode = controllerObj.getStatus();
+                    statusCode = controllerObj.getStatus() || statusCode;
                 }
 
                 // WARNING: This file was auto-generated with tsoa. Please do not modify it. Re-run tsoa to re-generate this file: https://github.com/lukeautry/tsoa
--- a/packages/cli/src/routeGeneration/templates/hapi.hbs
+++ b/packages/cli/src/routeGeneration/templates/hapi.hbs
@@ -245,8 +245,7 @@ export function RegisterRoutes(server: any) {
 
                 if (isController(controllerObj)) {
                     headers = controllerObj.getHeaders();
-
-                statusCode = controllerObj.getStatus();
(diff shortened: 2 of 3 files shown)


WoH:
@devnev I think this makes sense, and it doesn't conflict with your interpretation of what should happen in these cases,  right?

devnev:
@WoH Yeah I think this makes sense too.

WoH:
@brian-pinwheel LGTM then, but could you add a test for this?

brian-pinwheel:
> @brian-pinwheel LGTM then, but could you add a test for this?

@WoH I would love to, but honestly, this is my first time using Express and don't know how to write tests.  I come from more of a Java world, which is why TSOA was a natural fit.  Unfortunately, I have other higher priorities at this time that prevent me from taking the time to learn how the tests work. I have to imagine the test wouldn't take very long, but the time taken to debug the issue has set the project back so trying to catch back up to our release date.

WoH:
I added a test on master. Please note that using `@SuccessResponse` on controllers that inherit `Controller` by tsoa will still lead to unforseen consequences if you're using Singleton controllers.
Repro: Handler `handleA` uses `setStatus` to set the status prop (not no longer `undefined`, but now, for example 404. ). Now we call `handlerB`, which uses `@SuccessResponse(418)`. At runtime, the controller is still the singleton created by your Factory/DI with a status prop of 404. The check determines the code should be 404 || 418 -> 404 is returned).

Solutions: 
a) Don't use singleton contollers
b) reset the status code to undefined

devnev:
FYI, we're now avoiding any accidental false sharing by enabling `useSuccessResponseCode` and not inheriting `Controller`, which bypasses the code in the `isController(...)` condition and thus the use of controller state. For the main response from a method, the status comes from `@SuccessResponse`, and when we need an alternative response status, we use an `@Res` callback.

WoH:
Exactly. And the `@SuccessStatus` + `@Res` is the better API that we should encourage. This PR allows for the admittedly dubious `this.setStatus(undefined as number)` escape that you can use to migrate.

PR #965, opened 2021-04-20
expose generateSpecAndRoutes for consumption

### LINKED ISSUE #964: Can generateSpecAndRoutes please be a module export?
Hello, I'm wondering if you would mind exposing/re-exporting the generateSpecAndRoutes function? And also allow me to pass in the configuration object manually at this point (ie, ConfigArgs referenced by generateSpecAndroutes would be of type `string | Config`? I'd like to bundle this in with my existing build system and it would be convenient for this to be directly exposed.

I'm already watching my filesystem, and the rest of my build process and build configuration (using esbuild) is colocated. It would help consolidate both process and documentation for me to be able to codify in the same place as everything else in our build that:

"Whenever these controller files change, run generateSpecAndRoutes". I could just replicate what you're already doing with that function, but it would be nice to gain ongoing compatibility with whatever updates/improvements/changes you end up making down the road.

## Sorting

- **I'm submitting a ...**

  - [ ] bug report
  - [X] feature request
  - [ ] support request

- I confirm that I

## Expected Behavior

```ts
import { generateSpecAndRoutes } from 'tsoa';

generateSpecAndRoutes({
  configuration: { 
    routes: {
      routesDir: "src/server"
      //...etc
     },
    spec: {
      specVersion: 3 
      // ...etc
    }
  }
});
```

## Current Behavior

```ts
import { generateSpecAndRoutes } from 'tsoa' // no exported member
```

## Steps to Reproduce

1. 2. 3. 4.

## Context (Environment)

Version of the library: 
Version of NodeJS: 

mscottnelson:
### All Submissions:

**Closing issues**

closes #964

**Test plan**

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -3,7 +3,7 @@ import * as path from 'path';
 import * as ts from 'typescript';
 import * as YAML from 'yamljs';
 import * as yargs from 'yargs';
-import { Config, RoutesConfig, SpecConfig } from '@tsoa/runtime';
+import { Config, RoutesConfig, SpecConfig, Tsoa } from '@tsoa/runtime';
 import { MetadataGenerator } from './metadataGeneration/metadataGenerator';
 import { generateRoutes } from './module/generate-routes';
 import { generateSpec } from './module/generate-spec';
@@ -72,6 +72,10 @@ const getConfig = async (configPath = 'tsoa.json'): Promise<Config> => {
   }
 
   return config;
+};
+
+const resolveConfig = async (config?: string | Config) => {
+  return typeof config === 'object' ? config : await getConfig(config);
 };
 
 const validateCompilerOptions = (config?: Record<string, unknown>): ts.CompilerOptions => {
@@ -211,7 +215,7 @@ const jsonArgs: yargs.Options = {
 
 export interface ConfigArgs {
   basePath?: string;
-  configuration?: string;
+  configuration?: string | Config;
 }
 
 export interface SwaggerArgs extends ConfigArgs {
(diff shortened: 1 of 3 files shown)


packages/cli/src/cli.ts
```diff
@@ -74,6 +74,10 @@ const getConfig = async (configPath = 'tsoa.json'): Promise<Config> => {
   return config;
 };
 
+const resolveConfig = async (config?: string | Config) => {
+  return typeof config === 'object' ? config : await getConfig(config);
```
WoH, line 78: Don't release Zalgo

mscottnelson: I know about Zalgo but I'm not sure what you're asking for here. This is an async function, and its return value will always be on "a later tick". 

Returning `config` directly will still result in an async function signature / Promise. Adding an await, as in `return await config` *will* further defer by adding an additional Promise resolution, but it doesn't increase determinacy in any way that I can see? For that matter, I guess `getConfig` doesn't need to be awaited, either, and `return (typeof config==="config") ? config : getConfig(config)` would remove an unneeded intermediary Promise resolution.

I added a clarifying return type to the function signature. Let me know if there's something else I am missing here. Thanks for looking this over!

WoH: Missed the async for some reason, your right

WoH:
> If you think it aligns, I'd be happy to add unit tests to cover the various scenarios.

Can you add 1/2 of these?

WoH:
I'll merge this as is, given it affects the API only in places where we explicitly state that we don't follow SemVer there.

mscottnelson:
thanks @WoH, apologies that I didn't get an opportunity to come back around to the tests.

PR #987, opened 2021-05-10
Expand @deprecated / @Deprecated support for OAS 3.0

### LINKED ISSUE #975: Feature: deprecated schemas/fields
Open API 3.0 allows you to [deprecate a schema/field](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.2.md#fixed-fields-20), e.g. 

```json
{
  "newHotness": {
    "type": "string",
  },
  "oldAndBusted": {
    "type": "number",
    "deprecated": true
  }
}
```

While tsoa has a `@Deprecated` decorator for methods, it can't be used for deprecating schemas/fields. Seems like tsoa could look for `@deprecated` jsdoc annotations and deprecate things accordingly in the generated spec.

Happy to put together a PR for this if you're on board.

## Sorting

- **I'm submitting a ...**

  - [ ] bug report
  - [x] feature request
  - [ ] support request

- I confirm that I

jenseng:
Closes #975

OpenAPI 3.0 allows you to mark schemas and parameters as deprecated. See:
* https://swagger.io/specification/#schemaDeprecated
* https://swagger.io/specification/#parameterDeprecated

This PR enhances tsoa to look for `@Deprecated` decorators where TS allows them (on parameters and class properties) as well as `@deprecated` JSDoc annotations (on parameters and class/interface/alias properties) and sets the `deprecated` field accordingly when generating the 3.0 schema.

### All Submissions:

No major issues, though it didn't seem like there was a clear way to support `@deprecated` cascading down through non-trivial mapped types, as there is insufficient info in the AST. I was able to support it for straightforward mapped types that use `keyof`, and this behavior matches intellisense in VS Code, so maybe that's ok? See [this comment](https://github.com/jenseng/tsoa/blob/support-deprecated-schemas/packages/cli/src/metadataGeneration/typeResolver.ts#L185) and [these test fixtures](https://github.com/jenseng/tsoa/blob/support-deprecated-schemas/tests/fixtures/testModel.ts#L151-L159) for more info.

**Test plan**

* Added relevant tests, all passing, specifically covering:
 * `@deprecated` JSDoc annotations on class method parameters (i.e. controller operations)
 * `@deprecated` JSDoc annotations on class properties
 * `@deprecated` JSDoc annotations on class parameter properties
 * `@deprecated` JSDoc annotations on type/interface properties
 * `@deprecated` JSDoc annotations cascading through simple mapped types
 * `@Deprecated` decorators on class method parameters (i.e. controller operations)
 * `@Deprecated` decorators on class properties
 * `@Deprecated` decorators on class properties cascading through simple mapped types
 * `@Deprecated` decorators on class parameter properties
 * Negative tests for params/properties/etc that should not be deprecated
 * Negative tests for deprecated things that should not appear in the spec (e.g. non-public class parameter property) 
 * Negative tests across the board for OAS < 3.0
* Also verified end-to-end in my app and it added the missing `deprecated`s to swagger.json :)


jenseng:
Ok fixed now, added several more tests as well

packages/cli/src/metadataGeneration/parameterGenerator.ts
```diff
@@ -305,6 +318,10 @@ export class ParameterGenerator {
     return undefined;
   }
 
+  private getParameterDeprecation(node: ts.ParameterDeclaration) {
+    return isExistJSDocTag(node, tag => tag.tagName.text === 'deprecated') || isDecorator(node, identifier => identifier.text === 'Deprecated');
```
WoH, line 322: I'm a little torn here, the other (JSDoc) modifiers don't have corresponding Parameter Decorators.

packages/runtime/src/swagger/swagger.ts
```diff
@@ -128,6 +128,7 @@ export namespace Swagger {
   }
 
   export type Parameter = BodyParameter | FormDataParameter | QueryParameter | PathParameter | HeaderParameter;
+  export type Parameter3 = Parameter & { deprecated?: boolean };
```
WoH, line 131: I'd prefer to keep this in Parameter and set `x-deprecated` if we are dealing with v2, wdty?

jenseng: Sure, I can tweak this

packages/cli/src/metadataGeneration/parameterGenerator.ts
jenseng: I don't feel too strongly, I can leave off the decorator logic here if you like. Since TS supports decorating parameters I figured why not give people the option. I do think it's good to keep the JSDoc annotation support here since people already use that in the wild and it's supported by IDEs/etc.

tests/fixtures/testModel.ts
```diff
@@ -136,6 +138,24 @@ export interface TestModel extends Model {
         [name: string]: TypeAliasModel1;
       };
     };
+    /** @deprecated */
+    deprecatedSubProperty?: number;
+  };
+
+  /** not deprecated */
+  notDeprecatedProperty?: number;
+  /** type is deprecated, but this property is not */
```
jenseng, line 147: I'm wondering if i should rethink this, as I'm right in the middle of adding better `deprecated` support to openapi-generator and this seems at odds with that... In OAS terms, this becomes a `ref`, and if that corresponding schema happens to be deprecated, then this property should be considered deprecated, as everything from the ref comes in.

By the same token I need a test explicitly around deprecating a type alias (like line 141, but not a primitive); the ref overrides everything including an explicit `deprecated`. Might need to give that some thought (perhaps emit a warning?)

packages/cli/src/metadataGeneration/parameterGenerator.ts
WoH: I think we keep it, I can't see a case where it's causing confusion especially compared to the issue we'd inevitably get eventually when someone tries to do it using decorators. Intent is clear and it doesn't interfere with anything else.

tests/fixtures/testModel.ts
WoH: Here's what I've previously done, I think it was w/ regards to `descriptions`s along $refs. We still add it to the spec, knowing whoever parses the spec is bound by the spec to treat this as: `Any members [other than "$ref] SHALL be ignored.`. Warning about that seems reasonable, but we don't currently do it.

Seems like was a reasonable choice given that OpenAPI 3.1 allows for this explicitly: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#referenceObject

jenseng:
Rebased against latest master and added a second commit which (I think) addresses the PR feedback...
* Rework Parameter/Parameter2/Parameter3
* Add x-deprecated to OAS 2
* Fix an oversight where deprecated classes/types weren't reflected in the generated spec (see DeprecatedClass and DeprecatedType in the tests)

WoH:
Diffs look good, I'll try to make time to check this out locally tomorrow unless there's anything I should wait for.

jenseng:
I think it’s good to go, but let me know if anything else needs to change. Thanks!

WoH:
LGTM, thanks a lot for this addition!

Issue #659, opened 2020-04-12
Hint from authorizer which error response to return to the user


fantapop:
Our authorizer supports 3 different types of auth. In the case of all auth failing, the promise to resolve last is the one that is returned to the user.  For the express route that happens here:  https://github.com/lukeautry/tsoa/blob/0df645d5db22d6bc9c2b0e2e83b6550171004c86/src/routeGeneration/templates/express.hbs#L112

This is usually the correct error response. I think the one that is actually being attempted will likely get further in the authentication check than auth methods that can be immediately pruned out, However I've seen the wrong error being shown to our users on a few occasions and its very confusing.

## Sorting

* **I'm submitting a ...**
  - [ ] bug report
  - [x] feature request
  - [ ] support request

* I confirm that I
  - [x] used the [search](https://github.com/lukeautry/tsoa/search?type=Issues) to make sure that a similar issue hasn't already been submit

## Expected Behavior
The error related to the intended authentication method will be returned to the user.

## Current Behavior
The auth promise that takes the longest to resolve is returned to the user.

## Possible Solution
My idea is to allow the authorizer to reject the auth method with some additional information which indicates that it was not attempted by the user.  For example if an api supports api key as well as bearer token and a request is made in which the x-api-key header was not passed the authorizer could reject with:
```
{
    error: new Error("X-Api-Key Header was missing"),
    priority: 2,
}
```
and
```
{
    error: new Error("Bearer Token was expired"),
    priority: 1,
}
```

## Context (Environment)
Version of the library: 3.0.4
Version of NodeJS: 12
* Confirm you were using yarn not npm: [x]

## Breaking change?
This could be made in a way that supports the old way and the new way

WoH:
Can you provide some additional info, so we can better understand your use case.
Are you using one Security Decorator here or multiple?
Maybe it's a good idea to have a combined security decorator (if you are using multiple) so you have full control what to return from that one method?

fantapop:
We have 3 security decorators currently. 2 types of oauth implementations
and 1 api key decorator. Our web app has access to a different but
overlapping set of methods as our app. I’ve also written an endpoint to
create a custom swagger spec which is filtered by security type so that we
can generate clients with the subset of methods which are exposed to the
particular client. Currently that mechanism is relying on the security
type. I suppose it would be possible to twist scopes into doing what we
want here. I do think that the current mechanism is a little flawed in how
it works with multiple failures and multiple successes.

fantapop:
[...]

PR #995, opened 2021-05-15
974 koa hapi multi authentication

fantapop:
### All Submissions:

1. This solution is waiting for all authentication promises to resolve before proceeding.  A better solution would utilize Promise.any to all the code to proceed after the first success. The current build does not provide a polyfill for this method.  I'd  appreciate any thoughts on this.
2. I have limited knowledge of the Hapi server although this PR does make changes there.  It seems to me that setting the user via the assign block is redundant now and it seems to simplify the code to set it using one method only.  Let me know if I'm mistaken here.

**Test plan**

I've ported the AND / OR Security method tests from the express tests to Hapi / Koa.  Everything seems to be working from what I can tell.  I've also tested locally with my Koa server to verify the intitial issue I saw is fixed.

### DIFF
--- a/packages/cli/src/routeGeneration/templates/express.hbs
+++ b/packages/cli/src/routeGeneration/templates/express.hbs
@@ -8,6 +8,8 @@ import { {{name}} } from '{{modulePath}}';
 {{/each}}
 {{#if authenticationModule}}
 import { expressAuthentication } from '{{authenticationModule}}';
+// @ts-ignore - no great way to install types from subpackage
+const promiseAny = require('promise.any');
 {{/if}}
 {{#if iocModule}}
 import { iocContainer } from '{{iocModule}}';
@@ -104,51 +106,63 @@ export function RegisterRoutes(app: express.Router) {
     // WARNING: This file was auto-generated with tsoa. Please do not modify it. Re-run tsoa to re-generate this file: https://github.com/lukeautry/tsoa
 
     {{#if useSecurity}}
+
+    // WARNING: This file was auto-generated with tsoa. Please do not modify it. Re-run tsoa to re-generate this file: https://github.com/lukeautry/tsoa
+
     function authenticateMiddleware(security: TsoaRoute.Security[] = []) {
-        return function runAuthenticationMiddleware(request: any, _response: any, next: any) {
-            let responded = 0;
-            let success = false;
-
-            const succeed = function(user: any) {
-                if (!success) {
-                    success = true;
-                    responded++;
-                    request['user'] = user;
-                    next();
-                }
-            }
+        return async function runAuthenticationMiddleware(request: any, _response: any, next: any) {
 
             // WARNING: This file was auto-generated with tsoa. Please do not modify it. Re-run tsoa to re-generate this file: https://github.com/lukeautry/tsoa
 
-            const fail = function(error: any) {
-                responded++;
-                if (responded == security.length && !success) {
-                    error.status = error.status || 401;
-                    next(error)
-                }
-            }
+            // keep track of failed auth attempts so we can hand back the most
+            // recent one.  This behavior was previously existing so preserving it
+            // here
+            const failedAttempts: any[] = [];
+            const pushAndRethrow = (error: any) => {
+                failedAttempts.push(error);
+                throw error;
+            };
 
-            // WARNING: This file was auto-generated with tsoa. Please do not modify it. Re-run tsoa to re-generate this file: https://github.com/lukeautry/tsoa
-
+            const secMethodOrPromises: Promise<any>[] = [];
             for (const secMethod of security) {
                 if (Object.keys(secMethod).length > 1) {
-                    let promises: Promise<any>[] = [];
+                    const secMethodAndPromises: Promise<any>[] = [];
 
                     for (const name in secMethod) {
-                        promises.push(expressAuthentication(request, name, secMethod[name]));
+                        secMethodAndPromises.push(
+                            expressAuthentication(request, name, secMethod[name])
+                                .catch(pushAndRethrow)
+                        );
                     }
 
-                    Promise.all(promises)
-                        .then((users) => { succeed(users[0]); })
-                        .catch(fail);
+                    // WARNING: This file was auto-generated with tsoa. Please do not modify it. Re-run tsoa to re-generate this file: https://github.com/lukeautry/tsoa
+
+                    secMethodOrPromises.push(Promise.all(secMethodAndPromises)
+                        .then(users => { return users[0]; }));
                 } else {
                     for (const name in secMethod) {
-                        expressAuthentication(request, name, secMethod[name])
-                            .then(succeed)
-                            .catch(fail);
+                        secMethodOrPromises.push(
+                            expressAuthentication(request, name, secMethod[name])
+                                .catch(pushAndRethrow)
+                        );
                     }
                 }
             }
+
+            // WARNING: This file was auto-generated with tsoa. Please do not modify it. Re-run tsoa to re-generate this file: https://github.com/lukeautry/tsoa
+
+            try {
+                request['user'] = await promiseAny(secMethodOrPromises);
+                next();
+            }
+            catch(err) {
+                // Show most recent error as response
+                const error = failedAttempts.pop();
+                error.status = error.status || 401;
+                next(error);
+            }
+
+            // WARNING: This file was auto-generated with tsoa. Please do not modify it. Re-run tsoa to re-generate this file: https://github.com/lukeautry/tsoa
         }
     }
     {{/if}}
--- a/packages/cli/src/routeGeneration/templates/hapi.hbs
+++ b/packages/cli/src/routeGeneration/templates/hapi.hbs
@@ -8,6 +8,8 @@ import { {{name}} } from '{{modulePath}}';
 {{/each}}
 {{#if authenticationModule}}
 import { hapiAuthentication } from '{{authenticationModule}}';
+// @ts-ignore - no great way to install types from subpackage
+const promiseAny = require('promise.any');
 {{/if}}
 {{#if iocModule}}
 import { iocContainer } from '{{iocModule}}';
@@ -60,8 +62,7 @@ export function RegisterRoutes(server: any) {
                 pre: [
                     {{#if security.length}}
                     {
-                      method: authenticateMiddleware({{json security}}),
-                      assign: "user"
+                      method: authenticateMiddleware({{json security}})
                     },
                     {{/if}}
                     {{#if uploadFile}}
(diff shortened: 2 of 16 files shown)


fantapop:
@technicallyfeasible I changed the approach with handling errors in the controller after authentication.  Based on the tests, It seems like its still working to me.  If you're still relying on this behavior, I'd appreciate your eyes here.

WoH:
> The current build does not provide a polyfill for this method. I'd appreciate any thoughts on this.

Please elaborate. The polyfill would never have to be provided by tsoa, but that's what I think your saying?

fantapop:
> > The current build does not provide a polyfill for this method. I'd appreciate any thoughts on this.
> 
> Please elaborate. The polyfill would never have to be provided by tsoa, but that's what I think you're saying?

I was assuming we wouldn't want to put the onus of including the polyfill on a user of tsoa so we'd either want to include it in the runtime as a dependency and load it in the routes or inline an implementation in the routes file itself.  I wasn't really sure which to do but after you mentioned this I did some research and found that the polyfill is only a few lines of code so I added it directly to the routes template.

After this, various bugs and misunderstandings of what the code was currently doing, led me to do a bit of a refactor of the authentication code. I added some new tests and kept all the rest of course.  I think Its a bit clearer now and I was able to bring the complicated part of the implementation back into sync across each server routes file. Let me know what you think.

technicallyfeasible:
> @technicallyfeasible I changed the approach with handling errors in the controller after authentication. Based on the tests, It seems like its still working to me. If you're still relying on this behavior, I'd appreciate your eyes here.

thanks for checking, will give this a go and get back to you

WoH:
> I was assuming we wouldn't want to put the onus of including the polyfill on a user of tsoa so we'd either want to include it in the runtime as a dependency and load it in the routes or inline an implementation in the routes file itself. 

When would any transpiler + tsconfig that someone could currently be using that takes the routeTemplate and transforms it to JS not take care of that?

WoH:
> but it would be a breaking change

Yes, if we use Promise.any without fallback.

I am not really familiar with the Hapi template, maybe @icopp can check that?

packages/cli/src/routeGeneration/templates/express.hbs
```diff
@@ -104,51 +104,73 @@ export function RegisterRoutes(app: express.Router) {
...
                 }
             }
+
+            // WARNING: This file was auto-generated with tsoa. Please do not modify it. Re-run tsoa to re-generate this file: https://github.com/lukeautry/tsoa
+
+            try {
+                request['user'] = await promiseAny(secMethodOrPromises);
+                next();
+            }
+            catch(err) {
```
WoH, line 166: Promise.any returns an AggregateError, the current polyfill does not which is worrying to me.

WoH, line 166: Looking at https://github.com/lukeautry/tsoa/issues/659 here.
The Promise.any would throw an AggregateError here, which would be great for determining what the respective issues were.
However the current Polyfill doesn't do that. Wdyt about a spec compliant shim?

WoH: This would be a breaking change, but one that makes sense and allows us to eventually remove witout headaches.

fantapop: I'm all for bringing in a spec compliant shim. It would utilize shared code better by using a single import and it would be easy to pull it out later once Promise.any becomes commonplace.

I'm not convinced that we'd want to return the aggregate error here however.  If we returned the AggregateError we'd still need some code outside of the framework to determine which one is the correct one to show the user.  It feels like integrating that into the authorization code is the way to go since it already exists and its a singular spot to contain crucial details that may not be passed along in the AggregateError.  Here's one example.

say we're using two kinds of auth:
1) bearer token authentication
2) api key

A user with a bearer token which is a JWT attempts to authenticate but the token is expired

the code checks bearer token auth and throws an Error saying "Token Expired"
the code then checks the api key auth and see's that there is no X-Api-Key Header at all so throws an error saying "Missing X-Api-Key Header". 

The current templates would return the error saying "missing X-Api-Key Header" since it was last to fail.  Inside the authorization code we've just checked for the X-Api-Key header, seen that its missing and we have a pretty good idea that this type of authentication is not the intended one.  If we sent both errors inside of an AggregateError we'd still need to do a follow up check on the request object to see that X-Api-Key header is missing.  This bit of code would probably need to do a lot of little checks which would end up looking like the authorization code path already does.

fantapop: I did push a fix up to show what the changes look like to use the polyfill instead.  I ended up using the imported method directly instead of executing any.shim() because typescript still didn't know about `Promise.any` after the shim ran so it would still involve extra changes to suppress this.

WoH: The imported method is perfectly fine.
For your example I'd argue it should be up to the developer to decide which error to show.

> still need some code outside of the framework to determine which one is the correct one to show the user

We already pass down the error, so this would not change. However, we could let the developer decide which Error to show if multiple errors occurred (in your example, we might end up passing down the Error for 2) when the token is expired.

> The current templates would return the error saying "missing X-Api-Key Header" since it was last to fail. 

Which is obviously what I'd want to prevent. Expired Token is the correct issue to face, wouldn't you agree?

Instead of one error you can have an AggregateError you can use to derive what the issues were.

fantapop: I agree we need to have the developer decide which error to show.  I'm seeing a few possibilities for how this could work.

1) In the event that no Authorization methods are accepted, we throw the aggregate error that contains each error.  In order to determine which error to show, the developer would need to add an additional middleware which runs after the generated routes to intercept the error and choose which one to show. In order to figure out the correct error, the developer would need to write a set of checks for each auth type.  For example, If X-api-key header was not passed, we know the authentication isn't any that relied on that header, etc.  These are probably simple checks depending on what the authentication methods are and how many there are.  Adding a new authentication method means updating the authentication.ts file and updating the middleware which is responsible to determine how to pick the correct error out of the AggregateError.  This would be a breaking change for users with multiple types of authentication.

2) We allow the authentication methods for each type of auth pass a special value in the thrown error which helps determine whether this is the correct error to show.  For example, this could be a priority value as mentioned in  #659  or it could be something like a flag which explicitly removes the error from consideration. For example `error.skip = true`.  The routes file then uses these additional flags to figure out the correct error and then falls back to existing behavior if the values are ambiguous or don't exist.  In this case, adding a new authentication method means just updating the authentication.ts file and not a new middleware.  It is not a breaking change since it falls back to existing behavior.  There is an existing pattern of adding special values to the error, see here for customizing status code via the authentication function: https://github.com/lukeautry/tsoa/blob/master/packages/cli/src/routeGeneration/templates/express.hbs#L126

3) we could add an additional optional developer defined function similar to the auth methods in the authentication.ts file.  It could be a specifically named method that is exported from authentication.ts which is specifically for determining which error to choose in this case.  If that method does not exist TSOA could fall back to the existing behavior of choosing the most recently failed.  This would avoid the breaking change associated with option 1.

I personally prefer option 2 because it doesn't require a breaking change and because i think that the code which is required to check whether the auth method is appropriate looks a lot like the code which is already in the authentication implementation. My assumption is that its just a line or two of code for each method to return whether its one you want to show.

Ultimately, I'll leave it up to you @WoH to decide which one you think is best for the framework.  Let me know your thoughts.

fantapop: One more option is keep what I have already which preserves the existing behavior and consider the other changes as part of a separate PR.  That would let us get this landed sooner.

WoH: I hope you're not making this assumption based on my (lack) of quick feedback which should be attributed to my schedule, I am sorry about that.

WoH: If you feel like making a non-breaking change and expanding on it with a (smaller) breaking change, I'm happy to do that, up to you. My thought was to make it easier on you by making only one PR and merging it with 2 other breaking changes that I have on mind for the next release (TS String Literals, node > 12, #997, possibly #1008)

WoH: The problem with 2) is I don't see how the priority would solve the issue. I.e. if there is no Authorization header that's low prio. If the Authorization header has an expired token, that's now very high priority for me, since I'm expecing the client intended to use that token mechanism to authorize. But it's the same token strategy in both cases.

fantapop: If you want to go with option 1 and make the breaking change, it would be good to get it out with  the next release.  Is that coming up soon?

I am thinking that it would be better to make a separate PR for the altering of error return behavior. It will be easier to understand from a historical perspective to keep the 2 issues separate.

Are you okay with this one as is considering it preserves existing behavior?  I've gone ahead and rebased in the fixup to use the external promise definition.

fantapop: Hmm, I just did some local testing using the updated version of this branch and ran into a couple issues.
1) my local project is complaining that the types for promise.any weren't installed. 
2) There's something weird about how the promise.any stuff is exported.  My project is complaining about the import * as promiseAny syntax. I was able to get it working 2 ways, I'm wondering which you recommend or if something else should be done:
These seem to work:
const promiseAny = require('promise.any');  // not ideal because it uses require, do some typescript configs restrict that?

fantapop: @WoH do you think the @types/promise.any would need to be included as a full dependency?  I don't think subpackage devDependencies are installed and if the developer using noImplictAny, they'll get a type error.

WoH: TS-ignore them or cast to explicit any

fantapop: okay, ts-ignore works well for the lack of types.  I ended up having to use the require for this shim because i couldn't get the tests working without it.  Not sure why.

WoH: Can you elaborate? 
```
import * as promiseAny from 'promise.any'
```

should be fine.

packages/cli/src/routeGeneration/templates/express.hbs
```diff
@@ -104,51 +106,63 @@ export function RegisterRoutes(app: express.Router) {
...
+                // Show most recent error as response
+                const error = failedAttempts.pop();
```
WoH, line 160: Instead of manual tracking, err.errors.pop()

WoH: <details>
<summary>Here's the combined diff</summary>

```diff
diff --git a/packages/cli/src/routeGeneration/templates/express.hbs b/packages/cli/src/routeGeneration/templates/express.hbs
index 222b28b..1ae0354 100644
--- a/packages/cli/src/routeGeneration/templates/express.hbs
+++ b/packages/cli/src/routeGeneration/templates/express.hbs
@@ -9,7 +9,7 @@ import { {{name}} } from '{{modulePath}}';
 {{#if authenticationModule}}
 import { expressAuthentication } from '{{authenticationModule}}';
 // @ts-ignore - no great way to install types from subpackage
-const promiseAny = require('promise.any');
+import * as promiseAny from 'promise.any';
 {{/if}}
 {{#if iocModule}}
 import { iocContainer } from '{{iocModule}}';
@@ -114,15 +114,6 @@ export function RegisterRoutes(app: express.Router) {
 
             // WARNING: This file was auto-generated with tsoa. Please do not modify it. Re-run tsoa to re-generate this file: https://github.com/lukeautry/tsoa
 
-            // keep track of failed auth attempts so we can hand back the most
-            // recent one.  This behavior was previously existing so preserving it
-            // here
-            const failedAttempts: any[] = [];
-            const pushAndRethrow = (error: any) => {
-                failedAttempts.push(error);
-                throw error;
-            };
-
             const secMethodOrPromises: Promise<any>[] = [];
             for (const secMethod of security) {
                 if (Object.keys(secMethod).length > 1) {
@@ -131,7 +122,6 @@ export function RegisterRoutes(app: express.Router) {
                     for (const name in secMethod) {
                         secMethodAndPromises.push(
                             expressAuthentication(request, name, secMethod[name])
-                                .catch(pushAndRethrow)
                         );
                     }
 
@@ -143,7 +133,6 @@ export function RegisterRoutes(app: express.Router) {
                     for (const name in secMethod) {
                         secMethodOrPromises.push(
                             expressAuthentication(request, name, secMethod[name])
-                                .catch(pushAndRethrow)
                         );
                     }
                 }
@@ -157,7 +146,7 @@ export function RegisterRoutes(app: express.Router) {
             }
             catch(err) {
                 // Show most recent error as response
-                const error = failedAttempts.pop();
+                const error = err.errors.pop();
                 error.status = error.status || 401;
                 next(error);
             }
diff --git a/packages/cli/src/routeGeneration/templates/hapi.hbs b/packages/cli/src/routeGeneration/templates/hapi.hbs
index ddb3251..c1822b6 100644
--- a/packages/cli/src/routeGeneration/templates/hapi.hbs
+++ b/packages/cli/src/routeGeneration/templates/hapi.hbs
@@ -9,7 +9,7 @@ import { {{name}} } from '{{modulePath}}';
 {{#if authenticationModule}}
 import { hapiAuthentication } from '{{authenticationModule}}';
 // @ts-ignore - no great way to install types from subpackage
-const promiseAny = require('promise.any');
+import * as promiseAny from 'promise.any';
 {{/if}}
 {{#if iocModule}}
 import { iocContainer } from '{{iocModule}}';
@@ -145,15 +145,6 @@ export function RegisterRoutes(server: any) {
 
             // WARNING: This file was auto-generated with tsoa. Please do not modify it. Re-run tsoa to re-generate this file: https://github.com/lukeautry/tsoa
 
-            // keep track of failed auth attempts so we can hand back the most
-            // recent one.  This behavior was previously existing so preserving it
-            // here
-            const failedAttempts: any[] = [];
-            const pushAndRethrow = (error: any) => {
-                failedAttempts.push(error);
-                throw error;
-            };
-
             const secMethodOrPromises: Promise<any>[] = [];
             for (const secMethod of security) {
                 if (Object.keys(secMethod).length > 1) {
@@ -162,7 +153,6 @@ export function RegisterRoutes(server: any) {
                     for (const name in secMethod) {
                         secMethodAndPromises.push(
                             hapiAuthentication(request, name, secMethod[name])
-                                .catch(pushAndRethrow)
                         );
                     }
 
@@ -174,7 +164,6 @@ export function RegisterRoutes(server: any) {
                     for (const name in secMethod) {
                         secMethodOrPromises.push(
                             hapiAuthentication(request, name, secMethod[name])
-                                .catch(pushAndRethrow)
                         );
                     }
                 }
@@ -188,7 +177,7 @@ export function RegisterRoutes(server: any) {
             }
             catch(err) {
                 // Show most recent error as response
-                const error = failedAttempts.pop();
+                const error = err.errors.pop();
                 if (isBoom(error)) {
                     throw error;
                 }
diff --git a/packages/cli/src/routeGeneration/templates/koa.hbs b/packages/cli/src/routeGeneration/templates/koa.hbs
index 5147456..4855634 100644
--- a/packages/cli/src/routeGeneration/templates/koa.hbs
+++ b/packages/cli/src/routeGeneration/templates/koa.hbs
@@ -13,7 +13,7 @@ import { {{name}} } from '{{modulePath}}';
 {{#if authenticationModule}}
 import { koaAuthentication } from '{{authenticationModule}}';
 // @ts-ignore - no great way to install types from subpackage
-const promiseAny = require('promise.any');
+import * as promiseAny from 'promise.any';
 {{/if}}
 {{#if iocModule}}
 import { iocContainer } from '{{iocModule}}';
@@ -114,15 +114,6 @@ export function RegisterRoutes(router: KoaRouter) {
 
             // WARNING: This file was auto-generated with tsoa. Please do not modify it. Re-run tsoa to re-generate this file: https://github.com/lukeautry/tsoa
 
-            // keep track of failed auth attempts so we can hand back the most
-            // recent one.  This behavior was previously existing so preserving it
-            // here
-            const failedAttempts: any[] = [];
-            const pushAndRethrow = (error: any) => {
-                failedAttempts.push(error);
-                throw error;
-            };
-
             const secMethodOrPromises: Promise<any>[] = [];
             for (const secMethod of security) {
                 if (Object.keys(secMethod).length > 1) {
@@ -131,7 +122,6 @@ export function RegisterRoutes(router: KoaRouter) {
                     for (const name in secMethod) {
                         secMethodAndPromises.push(
                             koaAuthentication(context.request, name, secMethod[name])
-                                .catch(pushAndRethrow)
                         );
                     }
 
@@ -141,7 +131,6 @@ export function RegisterRoutes(router: KoaRouter) {
                     for (const name in secMethod) {
                         secMethodOrPromises.push(
                             koaAuthentication(context.request, name, secMethod[name])
-                                .catch(pushAndRethrow)
                         );
                     }
                 }
@@ -157,7 +146,7 @@ export function RegisterRoutes(router: KoaRouter) {
             }
             catch(err) {
                 // Show most recent error as response
-                const error = failedAttempts.pop();
+                const error = err.errors.pop();
                 context.status = error.status || 401;
                 context.throw(context.status, error.message, error);
             }
```
</details>

packages/cli/src/routeGeneration/templates/express.hbs
fantapop: I retried again and it seems to be working fine.  I think my environment was slightly messed up.  I'll push that change back up.

packages/cli/src/routeGeneration/templates/express.hbs
fantapop: if we pop the error off of the aggregated error list, I believe this will always be the error from the last auth decorator.  The failedAttempts array tracks the rejections in the order that they failed rather than the order they were specified.  I do like the simplicity of not worrying about that but it is a breaking change and I've always thought the behavior tended to choose the "right" error to report on in most cases. I could be wrong though. Since you're getting ready to rev the major we could make that change as part of that assuming we also get something in to control the order in a timely manner.

packages/cli/src/routeGeneration/templates/express.hbs
fantapop: actually it doesn't work. This version builds and passes the tests but when I test it with my project I get this error:

`err:  TypeError: promiseAny is not a function`

I looked at the promise.any code and its doing something fancy to make the import work as a function an object. I'm wondering if this is causing it to break using this syntax.  I'm unclear why its not breaking for the tests.  Maybe I've got my project configured to handle the imports differently?

here is the source:
https://github.com/es-shims/Promise.any/blob/master/index.js

packages/cli/src/routeGeneration/templates/express.hbs
WoH: Seems about right, my brain actively tries to block that out as it's just not how I would want this to (have) work(ed).
I'll try to build on your PR and propagate a `TsoaAuthenticationError` that extends `AggregateError` for v4, I think this has to be the long-term plan.
However, I still can't reproduce the tests failing with `import * as`. Any hints there?

PR #1000, opened 2021-05-29
Feat/custom example label

### LINKED ISSUE #970: Custom labels for examples
Currently example labels are Example 1,2,3 .... I would like to customise example labels.
## Sorting

- **I'm submitting a ...**

  - [ ] bug report
  - [x] feature request
  - [ ] support request

- I confirm that I

Yes I tried " Custom example label/name".

## Expected Behavior

It would be nice to have a way to customise example labels. 

```yaml
examples:
  MyCustomExample1:
    - attr1
  MyCustomExample2:
    - attr2
```

```ts
```

## Current Behavior

```yaml
examples:
    Example 1
     - attr1
    Example 2
     - attr2
```

```ts
@example onError {"code":"customErrorCode1","messsage":"Invalid book name"}
@example onError {"code":"customErrorCode2","messsage":"No such book"}
```

## Possible Solution
Maybe adding a dot follow by example label
```
@example onError.Invalid_book_name {"code":"customErrorCode1","message":"Invalid book name"}
@example onError.No_such_book {"code":"customErrorCode2","message":"No such book"}
```

## Steps to Reproduce

```
class Controller(){
  /*
   * @example onError {"code":"customErrorCode1", "message":"Invalid book name"}
   * @example onError {"code":"customErrorCode2", "message":"No such book"}
   * /
  @Get('/book')
  public getBook(@Res() onError: TsoaResponse<400, SomeError>,):string{
    return
  }
}

```
## Context (Environment)

Version of the library:
v3.7.0
Version of NodeJS: 
v12.19.0

davidqqq:
### All Submissions:

**Closing issues**

closes #970 

Current solution has one limitation where space cannot be used because it will disrupt json parsing. 
`@example error.customError` or `@example error.custom_error` but not `@example error.custom error`. 

We may use "_" for space. 

**Test plan**

Test whether label following dot has correctly been included in example object output
Test whether no or empty label has correctly been included in example object output

### DIFF
--- a/packages/runtime/src/metadataGeneration/tsoa.ts
+++ b/packages/runtime/src/metadataGeneration/tsoa.ts
@@ -42,6 +42,7 @@ export namespace Tsoa {
     default?: any;
     validators: Validators;
     deprecated: boolean;
+    exampleLabels?: string[];
   }
 
   export interface ResParameter extends Response, Parameter {
@@ -72,6 +73,7 @@ export namespace Tsoa {
     name: string;
     schema?: Type;
     examples?: unknown[];
+    exampleLabels?: string[];
     headers?: HeaderType;
   }
 
--- a/packages/runtime/src/metadataGeneration/tsoa.ts
+++ b/packages/runtime/src/metadataGeneration/tsoa.ts
@@ -42,7 +42,7 @@ export namespace Tsoa {
     default?: any;
     validators: Validators;
     deprecated: boolean;
-    exampleLabels?: string[];
+    exampleLabels?: Array<string | undefined>;
   }
 
   export interface ResParameter extends Response, Parameter {
@@ -73,7 +73,7 @@ export namespace Tsoa {
     name: string;
     schema?: Type;
     examples?: unknown[];
-    exampleLabels?: string[];
+    exampleLabels?: Array<string | undefined>;
     headers?: HeaderType;
   }
 
--- a/tests/fixtures/controllers/exampleController.ts
+++ b/tests/fixtures/controllers/exampleController.ts
@@ -131,6 +131,8 @@ export class ExampleTestController {
    * @param res The alternate response
    * @example res.NoSuchCountry { "errorMessage":"No such country", "errorCode": 40000 }
    * @example res. { "errorMessage":"No custom label", "errorCode": 40000 }
+   * @example res "Unlabeled 1"
+   * @example res "Another unlabeled one"
    * @example res.NoSuchCity {
    *     "errorMessage":"No such city",
    *     "errorCode": 40000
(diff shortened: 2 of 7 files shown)


tests/fixtures/controllers/exampleController.ts
```diff
@@ -126,4 +126,20 @@ export class ExampleTestController {
     res?.(400, 123, { 'custom-header': 'hello' });
     return 'test 1';
   }
+
+  /**
+   * @param res The alternate response
+   * @example res.NoSuchCountry { "errorMessage":"No such country", "errorCode": 40000 }
```
WoH, line 132: Do we need the `.` separator at all?

Looking at something like
```ts
   @example res NoSuchCountry { "errorMessage":"No such country", "errorCode": 40000 }
```

Is there any ambiguity if we split on `' '` and determine based on the length of the result?

packages/runtime/src/metadataGeneration/tsoa.ts
```diff
@@ -42,6 +42,7 @@ export namespace Tsoa {
     default?: any;
     validators: Validators;
     deprecated: boolean;
+    exampleLabels?: string[];
```
WoH, line 45: Change `example?: unknown[]` to `examples?: Record<string, unknown>` instead of adding labels separately.

tests/unit/swagger/schemaDetails3.spec.ts
```diff
@@ -698,6 +698,30 @@ describe('Definition generation for OpenAPI 3.0.0', () => {
             },
           });
         });
+
+        it('Supports custom example labels', () => {
```
WoH, line 702: Can you please add specs for OpenAPI 2? While that version does not support examples/example, it still should be clear what happens. In other cases, we add this as an `x-example`, I think it would be nice to add the first example without label and possibly warn about the lack of support.

tests/fixtures/controllers/exampleController.ts
```diff
@@ -126,4 +126,20 @@ export class ExampleTestController {
...
+   * @example res. { "errorMessage":"No custom label", "errorCode": 40000 }
```
WoH, line 133:
```suggestion
   * @example res. { "errorMessage":"No custom label", "errorCode": 40000 }
   * @example res "Throw the index off"
```

WoH: Just add one to the middle to make sure they don't throw off the array indices.

tests/unit/swagger/schemaDetails3.spec.ts
davidqqq: I get what you mean in general but not sure about the "example without label". I used an example with label in test and emit a warning if exampleLabels is not empty which should indicate usage of example labels.

WoH:
LGTM, thank you for your contribution! 🚀

PR #1008, opened 2021-06-04
Added readOnly to interface models properties

### LINKED ISSUE #1004: [QUESTION]: How to set model interface field as readOnly: true
I would like to know if it is currently possible to set specific model fields as 'readOnly'.

I believe it should be possible to set a specific field as readOnly following Open API 3.0 standards. I have tried a few variations of the following:

```
export interface Payload {

    /**
     * List of values
     */
    values: any[];

    /**
     * read only field of type string
     * @readonly
     */
    field?: string;
}
```

But had no success.

## Sorting

- **I'm submitting a ...**

  - [ ] bug report
  - [ ] feature request
  - [x] support request

- I confirm that I

## Expected Behavior

The expected behavior should be that the generated swagger doc should contain the information that the field is readOnly:

```
"Payload": {
    "description": "This is a model",
    "properties": {
        "values": {
            "items": {},
            "type": "array",
            "description": "list of values"
        },
        "field": {
            "type": "string",
            "readOnly": true,
            "description": "read only field of type string"
        }
    },
    "required": [
        "values"
    ],
    "type": "object",
    "additionalProperties": false
}
```

## Current Behavior

Currently it generates the swagger spec without the "readOnly: true".

itair-miguel:
### All Submissions:

**Closing issues**

closes #1004 

No potential problems. I have just added the property to interface model properties though.

**Test plan**

The unit tests added follow the same approach as for other model interfaces properties.

### DIFF
--- a/packages/cli/src/metadataGeneration/typeResolver.ts
+++ b/packages/cli/src/metadataGeneration/typeResolver.ts
@@ -890,6 +890,7 @@ export class TypeResolver {
       type: new TypeResolver(propertySignature.type, this.current, propertySignature.type.parent, this.context, propertySignature.type).resolve(),
       validators: getPropertyValidators(propertySignature) || {},
       deprecated: isExistJSDocTag(propertySignature, tag => tag.tagName.text === 'deprecated'),
+      readOnly: this.getNodeReadOnly(propertySignature),
     };
     return property;
   }
@@ -1108,6 +1109,10 @@ export class TypeResolver {
     } else {
       return undefined;
     }
+  }
+
+  private getNodeReadOnly(node: ts.PropertySignature) {
+    return node.modifiers?.some(modifier => modifier.kind === ts.SyntaxKind.ReadonlyKeyword) || isExistJSDocTag(node, tag => tag.tagName.text === 'readonly');
   }
 }
 
--- a/tests/fixtures/testModel.ts
+++ b/tests/fixtures/testModel.ts
@@ -185,6 +185,11 @@ export interface TestModel extends Model {
   inlineMappedTypeRemapped?: {
     [K in keyof ParameterTestModel as `${Capitalize<K>}Prop`]?: string;
   };
+
+  /** @readonly */
+  readOnlyFieldAnnotation?: string;
+  readonly readOnlyFieldKeyword?: string;
+  notReadOnlyField?: string;
 }
 
 /** @deprecated */
--- a/tests/unit/swagger/definitionsGeneration/definitions.spec.ts
+++ b/tests/unit/swagger/definitionsGeneration/definitions.spec.ts
@@ -628,6 +628,15 @@ describe('Definition generation', () => {
               type: 'object',
             });
           },
+          notReadOnlyField: (propertyName, propertySchema) => {
+            expect(propertySchema).not.to.haveOwnProperty('readOnly', `for property ${propertyName}`);
+          },
+          readOnlyFieldAnnotation: (propertyName, propertySchema) => {
+            expect(propertySchema['readOnly']).to.eq(true, `for property ${propertyName}[readOnly]`);
+          },
+          readOnlyFieldKeyword: (propertyName, propertySchema) => {
+            expect(propertySchema['readOnly']).to.eq(true, `for property ${propertyName}[readOnly]`);
+          },
           notDeprecatedProperty: (propertyName, propertySchema) => {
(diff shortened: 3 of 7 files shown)


packages/cli/src/metadataGeneration/typeResolver.ts
```diff
@@ -1109,6 +1110,10 @@ export class TypeResolver {
       return undefined;
     }
   }
+
+  private getNodeReadOnly(node: ts.PropertySignature) {
+    return node.modifiers?.some(modifier => modifier.kind === ts.SyntaxKind.ReadonlyKeyword) || isExistJSDocTag(node, tag => tag.tagName.text === 'readonly');
```
WoH, line 1115:
```suggestion
    return node.modifiers?.some(modifier => modifier.kind === ts.SyntaxKind.ReadonlyKeyword);
```

tests/fixtures/testModel.ts
```diff
@@ -185,6 +185,11 @@ export interface TestModel extends Model {
   inlineMappedTypeRemapped?: {
     [K in keyof ParameterTestModel as `${Capitalize<K>}Prop`]?: string;
   };
+
+  /** @readonly */
+  readOnlyFieldAnnotation?: string;
+  readonly readOnlyFieldKeyword?: string;
+  notReadOnlyField?: string;
```
WoH, line 192:
```suggestion
  readonly readOnlyFieldKeyword?: string;
  notReadOnlyField?: string;
```

tests/unit/swagger/definitionsGeneration/definitions.spec.ts
```diff
@@ -628,6 +628,15 @@ describe('Definition generation', () => {
               type: 'object',
             });
           },
+          notReadOnlyField: (propertyName, propertySchema) => {
+            expect(propertySchema).not.to.haveOwnProperty('readOnly', `for property ${propertyName}`);
+          },
+          readOnlyFieldAnnotation: (propertyName, propertySchema) => {
+            expect(propertySchema['readOnly']).to.eq(true, `for property ${propertyName}[readOnly]`);
+          },
```
WoH, line 636: remove

tests/unit/swagger/schemaDetails3.spec.ts
```diff
@@ -885,6 +885,15 @@ describe('Definition generation for OpenAPI 3.0.0', () => {
               type: 'object',
             });
           },
+          notReadOnlyField: (propertyName, propertySchema) => {
+            expect(propertySchema.readOnly).to.eq(undefined, `for property ${propertyName}.readOnly`);
+          },
+          readOnlyFieldAnnotation: (propertyName, propertySchema) => {
+            expect(propertySchema.readOnly).to.eq(true, `for property ${propertyName}.readOnly`);
+          },
```
WoH, line 893: remove

WoH (commented):
Good start, but there are 3 areas that aren't covered yet:

- PropertyDeclarations on Classes [See this AST](https://ts-ast-viewer.com/#code/MYGwhgzhAECC0G8BQ1XQE4FMwBMD2AdiAJ7QAO6eZAXNBAC7oCWBA5imhVQEy0PNskAXyRJ6xMpmgAlAPIAeACoA+aAF5EHVFlyES0ANoAFaC2gBrTMTwAzaIoC6tRcYcBuYR7ESpAIXUyCrDKQA)
- Mapped types: See the AST and the [typeResolver](https://github.com/lukeautry/tsoa/blob/master/packages/cli/src/metadataGeneration/typeResolver.ts#L175) 
- Validation: I think we should consider not passibg properties that are `readonly`. If  `additionalProperties` is `true`, possibly throwing errors. This way, we could avoid 2 ways which are confusing and instead stick to the TS keyword unless there's a good use case where we need both. Feel free to discuss.

packages/cli/src/metadataGeneration/typeResolver.ts
itair-miguel: Hi @WoH Thanks for your reply! I will look into your suggestions this evening or over the weekend.

In regards to your comment:
> Also, I think 2 ways are confusing, if we can avoid them. Let's stick to the TS keyword unless there's a good use case.

My reasoning is that the keyword `readonly` makes the compiler look for illegal property assignments, this can be a breaking change in case someone just want to make use of the swagger `readOnly` option that simply includes the property in a response but not in a request. Let me know what you think though.

WoH: I agree that this change would be breaking.

PR #1040, opened 2021-07-23
Allow null and primitive values in an Extension

### LINKED ISSUE #958: Extensions don't allow scalar / primitive values
`@Extension('x-internal','true')`

Creates: 

`"x-internal": "true"` as expected.

`@Extension('x-internal',true)` a primitive type like a bool / int won't be serialized

mhyeun:
### All Submissions:

- [ ] Have you written unit tests that cover the negative cases (i.e.: if bad data is submitted, does the library respond properly)? (Already existed as InvalidExtensionControllerGenerator test suite)

**Closing issues**

Closes #958.

**Test plan**

Updated the Extension data being used in `methodController.ts` and updated the tests in `metadata.spec.ts`, `schemaDetails.spec.ts` and `schemaDetails3.spec.ts`. These tests now include null, primitive values, and their usage in objects and arrays.

### DIFF
--- a/packages/cli/src/metadataGeneration/initializer-value.ts
+++ b/packages/cli/src/metadataGeneration/initializer-value.ts
@@ -39,6 +39,8 @@ export const getInitializerValue = (initializer?: ts.Expression, typeChecker?: t
         return dateString;
       }
       return;
+    case ts.SyntaxKind.NullKeyword:
+      return null;
     case ts.SyntaxKind.ObjectLiteralExpression:
       const objectLiteral = initializer as ts.ObjectLiteralExpression;
       const nestedObject: any = {};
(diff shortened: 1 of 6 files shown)


packages/cli/src/metadataGeneration/initializer-value.ts
```diff
@@ -39,6 +39,8 @@ export const getInitializerValue = (initializer?: ts.Expression, typeChecker?: t
         return dateString;
       }
       return;
+    case ts.SyntaxKind.NullKeyword:
+      return null;
```
mhyeun, line 43: Without this case, `null` would fall to the `default` case, being returned as `undefined`.

WoH:
LGTM, congrats and thank you for your first contribution!

PR #1053, opened 2021-08-12
add support for esm

noamokman:
### All Submissions:

- [ ] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/lukeautry/tsoa/pulls) for the same update/change?

**Closing issues**

**Test plan**

Added a small option in the routes config to support esm in newer versions of node.
It is required to have a `.js` extensions on import.
Also added a test for it.

I haven't opened an issue for it because it's a relatively small change.


SharonGrossman (approved):
LGTM 👍

noamokman:
@WoH what do you think?

packages/cli/src/routeGeneration/routeGenerator.ts
```diff
@@ -179,7 +179,7 @@ export class RouteGenerator {
 
   private getRelativeImportPath(fileLocation: string) {
     fileLocation = fileLocation.replace(/.ts$/, ''); // no ts extension in import
-    return `./${path.relative(this.options.routesDir, fileLocation).replace(/\\/g, '/')}`;
+    return `./${path.relative(this.options.routesDir, fileLocation).replace(/\\/g, '/')}${this.options.esm ? '.js' : ''}`;
```
WoH, line 182: We can't make this assumption. May be `.js`, but you can import  `.ts` directly.

tests/unit/templating/routeGenerator.spec.ts
```diff
@@ -96,5 +96,31 @@ describe('RouteGenerator', () => {
 
       expect(models).to.equal('./controllerWith.tsInPath');
     });
+
+    it('adds js for routes if esm is true', () => {
```
WoH, line 100: Moving towards esm compat is a good idea, but while adding the extension is one piece here, there are other pieces we should address (exports, moving away from require in templates)

packages/cli/src/routeGeneration/routeGenerator.ts
noamokman: even if you import a `.ts` file you need to write `.js` 😕

noamokman: https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c#can-i-import-esm-packages-in-my-typescript-project

tests/unit/templating/routeGenerator.spec.ts
WoH: In terms of specs, I'd like to add an ESM project in the tests (see [fixtures](https://github.com/lukeautry/tsoa/tree/master/tests/fixtures)) that has catches some basic regressions.

noamokman: oh yeah let's catch everything.
most of our production services are using tsoa, and I would love to move them to esm

noamokman: we need to think how we want the tests to run, we can't mix the runs of esm and non esm.
because we need to load the files differently.
(we need a package.json with `"type": "module"`)

noamokman:
ok looking at the fixtures, we also need to build the new `esm` fixture with a different typescript config.

tests/unit/templating/routeGenerator.spec.ts
WoH: Yes, unlike the other fixtures, we can't just throw the ts files in there and keep everything the same.
It should have it's own
- tsconfig
- package.json
- files + import

dflupu:
I've been looking forward to esm support from tsoa for a while. Is there anything I can do to help get this merged?

noamokman:
With more and more packages moving to esm only, this is actually a blocker for servers waiting to upgrade other packages.

As I see it, we can choose between:
1. Moving tsoa to support esm only, "rip the Band-Aid" in a new major version for tsoa.
2. Copying the tests and maintain duplicated tests for esm and none esm.
3. Add the minimum amount of code to stop blocking esm projects.

Option 3 is the easiest one by just merging this PR.
But I think that we should push for option 1. esm is compatible with node 12.

WoH:
1)  Won't happen. Too many blockers right now, i.e. APM support / instrumentation ecosystem as a whole.

This PR basically needs one thing, and it would be fine for me. An ESM fixture that verifies that the integration tests run fine, so we don't have any major regressions breaking ESM support without us noticing, which I would assume is more in your interest (as someone who wants to run on ESM) than me.

noamokman:
Sure, I'll try to add it to this PR

simllll:
@noamokman are you still looking into this? Can I help somewhere?

noamokman:
Yeah @simllll I'm still trying to work on this, the "core" functionality should be simple,
But the tests are the problem, I took some time to understand how current tests work,
And now I wanna make these tests check a fixture of an esm express server.

Because of how the tests are written, the test is the part that's "running" the server with the `supertest` module, so it means that we need the test runner itself to run in esm,
It means different `package.json` a new `tsconfig` and some magic to make it all work together.

I'm trying to make it all work and then I'll push it.

noamokman:
OK, I thinks this works now, would love a second look from everyone and if you can, approve running the tests

simllll:
LGTM :)

simllll:
I'm currently stuck with this issue too, can I help somehow to resolve it? :)

noamokman:
OK tests pass 🥳 
Rebased everything with master, separated the esm tests to a separate folder with it's own dependencies.

Please everyone take a look 😄

tests/esm/package.json
```diff
@@ -0,0 +1,46 @@
...
+  },
+  "author": "Luke Autry <lukeautry@gmail.com> (http://www.lukeautry.com)",
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/lukeautry/tsoa.git"
+  },
+  "engines": {
+    "yarn": ">=1.9.4",
+    "node": ">=6.0.0"
```
WoH, line 32: 12

tests/package.json
```diff
@@ -20,7 +20,8 @@
     "clean": "rimraf dist fixtures/*/routes.ts fixtures/*/customRoutes.ts",
     "pretest": "yarn clean && cross-env NODE_ENV=tsoa_test ts-node --require tsconfig-paths/register ./prepare.ts",
     "loading-msg-for-tests": "echo '... starting a test run to ensure quality. Note this takes some time as tests do not run in parallel (by design) ...' ",
-    "test": "node ../ensureYarn.js && tsc -b tsconfig.build.json && yarn loading-msg-for-tests && cross-env NODE_ENV=tsoa_test mocha '**/*.spec.ts' --exit --require ts-node/register --require tsconfig-paths/register --timeout 5000",
+    "test": "node ../ensureYarn.js && tsc -b tsconfig.build.json && yarn loading-msg-for-tests && yarn test-esm && cross-env NODE_ENV=tsoa_test mocha '**/*.spec.ts' --exclude 'esm/**/*' --exit --require ts-node/register --require tsconfig-paths/register --timeout 5000",
+    "test-esm": "yarn --cwd esm test",
```
WoH, line 24: Move to the esm pgk.json

noamokman: What do you mean?
it's running the test command from the esm package.json

WoH: If the workspace is registered properly, lerna should pick it up if you run `test --parallel`.
This is what I wanted to get to, but the top pkg doesn't execute test in all the workspaces, therefore the test command in tests has to call the esm tests.
Instead, change the top level tests to lerna run --parallel test and remove the cross references from here to the esm tests.

noamokman: Oh I understand,
I can't do that with the current implementation because the `pretest` step cleans and builds the required assets for both tests, I don't wanna duplicate the setup and cleanup code.
That means you can't really run it in parallel.

We can copy everything but I'm not sure we should

WoH: I am under the assumption that lerna will respect dependencies when building the execution plan. I'll try to check myself to verify, but assuming this is indeed the case, if set up correctly, the order should be test.pretest before both test commands. The command then should be `--stream`, not `--parallel though` in order to preserve the topology.

WoH: Let's move this out, I don't have time to propose a fix and I don't want delay this, the functionality lgtm.

PR #1061, opened 2021-08-24
feat: Allow to use variables and enums as Path #1060

### LINKED ISSUE #1060: Method descriptors do not support variables
## Sorting

- **I'm submitting a ...**

  - [ ] bug report
  - [x] feature request
  - [ ] support request

- I confirm that I

## Current Behavior

we tried to reuse existing variable/const/enum in Method and Path but path end up as `undefined` in OpenAPI spec

```ts
enum aliases {
  TEST_ALIAS = 'test-alias',
}

@Route('/example/{exampleId}/test')
export class DeliverablesController extends Controller {
  @Put(aliases.TEST_ALIAS)
  public createOrUpdateTest(
  ): any {
    return {};
  }
}
```

## Context (Environment)

Version of the library: 3.9.0
Version of NodeJS: 16

ra100:
### All Submissions:

**Closing issues**

closes #1060

This covers just variables/enums. It doesn't cover string templates (that would be another feature).

**Test plan**

Test cover usage of viariable (string) and enum passed to @Get() decorator, and that the path generated is same as value and not the "name" of a variable or enum.

### DIFF
--- a/packages/cli/src/metadataGeneration/methodGenerator.ts
+++ b/packages/cli/src/metadataGeneration/methodGenerator.ts
@@ -1,7 +1,7 @@
 import * as ts from 'typescript';
 import * as path from 'path';
 import { isVoidType } from '../utils/isVoidType';
-import { getDecorators, getDecoratorValues, getSecurites } from './../utils/decoratorUtils';
+import { getDecorators, getDecoratorValues, getPath, getSecurites } from './../utils/decoratorUtils';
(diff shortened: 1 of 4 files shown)


ra100:
Actually, there are no "negative" unit test, I'm not sure what they could test :) only if we wanted to check that some other types of structure won't work. Something like `@Get('/path' + variable)` `@Get(`path/${variable}`)`

WoH:
LGTM, thank you!

HonzaMac:
I like this update, when it's going to be released new version? cc @WoH

Issue #868, opened 2020-12-13
cli: Use paths from tsconfig


ljwagerfield:
## Sorting

- **I'm submitting a bug report**
- I confirm that I used the [search](https://github.com/lukeautry/tsoa/search?type=Issues) to make sure that a similar issue hasn't already been submit

## Expected Behaviour

- `tsoa routes` should honour the `compilerOptions.paths` section from the project's `tsconfig.json` file.

## Current Behaviour

- `tsoa routes` requires you to duplicate the `compilerOptions` section from your `tsconfig.json` file into the `tsoa.json` file, like so:

**tsoa.json**
```json
{
  "entryFile": "src/handlers/HandleRequest.ts",
  "noImplicitAdditionalProperties": "throw-on-extras",
  "controllerPathGlobs": ["src/**/*Controller.ts"],
  "spec": {
    "outputDirectory": "build",
    "specVersion": 3
  },
  "routes": {
    "routesDir": "build"
  },
  "compilerOptions": {
    "baseUrl": "./",
    "paths": {
      "@foo/bar": ["./src/"],
      "@foo/bar/*": ["./src/*"]
    }
  }
}
```

~~I discovered this trick by viewing [a historical version of the `README.md`](https://github.com/lukeautry/tsoa/tree/9760b9122db3d93f0b478383d30a35be190066d3#path-mapping) which outlines path mapping.~~

~~(Unfortunately, this documentation no-longer exists.)~~

[Documentation now lives here.](https://tsoa-community.github.io/docs/path-mapping.html)

I'm glad I found a working solution for path aliases, but I'm intrigued: shouldn't `tsoa` pick up the configuration from `tsconfig.json` rather than requiring you to duplicate your config into `tsoa.json > compilerOptions`?

## Possible Solution

Automatically lookup `compilerOptions.paths` from `tsconfig.json`.

## Steps to Reproduce

1. Create a project with path mappings, e.g. in `tsconfig.json`:

    ```json
    {
      "compilerOptions": {
        ...
        "baseUrl": "./",
        "paths": {
          "@foo/bar": ["./src/"],
          "@foo/bar/*": ["./src/*"]
        }
      }
    }
    ```
2. Reference a `model` type from your `controller` type using an absolute path, e.g.

    ```typescript
    import Foo from "@foo/bar/Foo"
    ```
3. Run `tsoa routes` and receive the following error:

    ```
    Error: No matching model found for referenced type Foo.
    ```

[...]

Issue #860, opened 2020-12-04
Omit doesn't seem to work


bytenik:
## Sorting

- **I'm submitting a ...**

  - [X] bug report
  - [ ] feature request
  - [ ] support request

- I confirm that I
  - [X] used the [search](https://github.com/lukeautry/tsoa/search?type=Issues) to make sure that a similar issue hasn't already been submit

## Expected Behavior

Omit should be able to be used to create types.

## Current Behavior

An error is generated:

> Error: No matching model found for referenced type Omit. If Omit comes from a dependency, please create an interface in your own code that has the same structure. Tsoa can not utilize interfaces from external dependencies. Read more at https://github.com/lukeautry/tsoa/blob/master/docs/ExternalInterfacesExplanation.MD

## Reproduction Code

```
export interface NewIndicator extends Omit<IndicatorBase, 'formula'> {
    ownerId?: string | null;
    formula?: string;
}
```

## Context (Environment)

Version of the library: 3.4.0
Version of NodeJS: 13.14.0

- Confirm you were using yarn not npm: [X]

bytenik:
Having looked through the code, I don't see how this worked / could have ever worked, but I know stuff was merged in to make it work. So I'm confused. It seems like the models are only searched for in import declarations within the file, and since Omit is not imported but rather ambient, it won't be found.

bytenik:
I'm happy to try to fix this myself, but, I would need some guidance since I'm new to the codebase. Thanks!

WoH:
It will be found if the declaration is included in your program. Make sure the relevant files are not exlcuded in your tsconfig. I don't think there's a bug, if there is, please try to provide a repro.

bytenik:
Omit is built into TypeScript.

bytenik:
Literally any use of Omit will fail:
```
export interface ParentModel { abc: string; def: string; }
export interface ChildModel extends Omit<ParentModel, 'abc'> { ghi: string; }```

bytenik:
@WoH I don't think this should have been closed, as `Omit` is built into TypeScript. Please advise?

WoH:
We use `Omit` a lot throughout our tests, which makes me fairly confident in asserting that it's not us. Everything else you said has nothing to do with the issue.

> Omit is built into TypeScript.

Location is still in `node_modules`, which you probably are still excluding from your program.

My advice would be to not down vote for giving you a hint despite the duplicated issue and instead prove your issue with a repro if you're so confident it's an issue.

[...]

Issue #698, opened 2020-05-14
Multiple matching models found for referenced type Omit


simllll:
## Sorting

- **I'm submitting a ...**

  - [x] bug report
  - [ ] feature request
  - [ ] support request

- I confirm that I
  - [x] used the [search](https://github.com/lukeautry/tsoa/search?type=Issues) to make sure that a similar issue hasn't already been submit

## Expected Behavior

There are some similar issues, but I couldn't find a solution so far. I was working with tsoa 3 already successfully, but now the following error appears and I cannot quite figure out why:
```
Error: Multiple matching models found for referenced type Omit; please make model names unique. Conflicts found: "/home/simon/Dev/hokify/hokify-server/node_modules/@elastic/elasticsearch/lib/Helpers.d.ts"; "/home/simon/Dev/hokify/hokify-server/node_modules/@types/mongodb/index.d.ts"; "/home/simon/Dev/hokify/hokify-server/node_modules/@types/lodash/ts3.1/common/common.d.ts".
    at new GenerateMetadataError (/home/simon/Dev/hokify/hokify-server/node_modules/tsoa/dist/metadataGeneration/exceptions.js:22:28)
    at TypeResolver.getModelTypeDeclaration (/home/simon/Dev/hokify/hokify-server/node_modules/tsoa/dist/metadataGeneration/typeResolver.js:675:19)
    at TypeResolver.typeArgumentsToContext (/home/simon/Dev/hokify/hokify-server/node_modules/tsoa/dist/metadataGeneration/typeResolver.js:775:32)
    at /home/simon/Dev/hokify/hokify-server/node_modules/tsoa/dist/metadataGeneration/typeResolver.js:814:38
    at Array.forEach (<anonymous>)
    at /home/simon/Dev/hokify/hokify-server/node_modules/tsoa/dist/metadataGeneration/typeResolver.js:811:26
    at Array.forEach (<anonymous>)
    at TypeResolver.getModelInheritedProperties (/home/simon/Dev/hokify/hokify-server/node_modules/tsoa/dist/metadataGeneration/typeResolver.js:807:25)
    at TypeResolver.getModelReference (/home/simon/Dev/hokify/hokify-server/node_modules/tsoa/dist/metadataGeneration/typeResolver.js:515:40)
    at TypeResolver.getReferenceType (/home/simon/Dev/hokify/hokify-server/node_modules/tsoa/dist/metadataGeneration/typeResolver.js:477:38)
```

[...]

Issue #681, opened 2020-05-05
Deeply nested wide mapped types on enums do not resolve (sometimes)


E-gy:
Nesting mapped types eventually [deeply] does not correctly resolve the schema - not marking top-level props as `required`; sometimes causing a crash, depending on the breadth of nested types; and in oddly specific configurations, it's weird.

## Sorting

- **I'm submitting a ...**

  - [x] bug report
  - [ ] feature request
  - [ ] support request

- I confirm that I
  - [x] used the [search](https://github.com/lukeautry/tsoa/search?type=Issues) to make sure that a similar issue hasn't already been submitted

## Setup

We'll use the following mapped types setup:
```ts
enum venum { ab = "cd", ef = "gh", ij = "kl"}
type contained = { name: string; val: number; }
type K1 = { [P in venum]: contained }
type K2 = { [P in venum]: { [P in venum]: contained } }
type K3 = { [P in venum]: { [P in venum]: { [P in venum]: contained } } }
type K4 = { [P in venum]: { [P in venum]: { [P in venum]: { [P in venum]: contained } } } }
type J2 = { [P in venum]: K1 }
type J3 = { [P in venum]: J2 }
type J4 = { [P in venum]: J3 }
type J5 = { [P in venum]: J4 }
```
and the extended
```ts
enum venum1 { ab = "cd1", ef = "gh1", ij = "kl1", mn = "op1", qr = "st1", uv = "wx1"}
enum venum2 { ab = "cd2", ef = "gh2", ij = "kl2", mn = "op2", qr = "st2", uv = "wx2"}
enum venum3 { ab = "cd3", ef = "gh3", ij = "kl3", mn = "op3", qr = "st3", uv = "wx3"}
enum venum4 { ab = "cd4", ef = "gh4", ij = "kl4", mn = "op4", qr = "st4", uv = "wx4"}
type contained = { name: string; val: number; }
type K1 = { [P in venum1]: contained }
type K2 = { [P in venum2]: { [P in venum1]: contained } }
type K3 = { [P in venum3]: { [P in venum2]: { [P in venum1]: contained } } }
type K4 = { [P in venum4]: { [P in venum3]: { [P in venum2]: { [P in venum1]: contained } } } }
type J2 = { [P in venum2]: K1 }
type J3 = { [P in venum3]: J2 }
type J4 = { [P in venum4]: J3 }
```

Basically, `K` types are mapped directly (without referring to smaller `K`), `J` are mapped to lower `J`.

[...]

PR #1084, opened 2021-09-17
fix: simplify and improve declaration lookup

simllll:
uses getSymbolAtLocation instead of manual lookup code.
This improves a lot of use cases, even though https://github.com/lukeautry/tsoa/blob/master/docs/ExternalInterfacesExplanation.MD is still true, technically it is not true now anymore. It resolves the nodes like the IDE does with "Jump to declaration" now.

This implements the idea of @WoH stated here: 
https://github.com/lukeautry/tsoa/issues/333#issuecomment-625258178
### All Submissions:

**Closing issues**
there is no exact ticket for this, sorry about this, I can create one or if anything is to discuess regarding this approach, we can discucess it here?

closes https://github.com/lukeautry/tsoa/issues/868

related https://github.com/lukeautry/tsoa/issues/333

also fixes "inaccurate" lookups like:
https://github.com/lukeautry/tsoa/issues/860
https://github.com/lukeautry/tsoa/issues/698

possbile also fixes
https://github.com/lukeautry/tsoa/issues/1033
https://github.com/lukeautry/tsoa/issues/681

**Test plan**

current tests are still working, tested it with 5 tsoa projects on our side. Everything still working like before.

### DIFF
--- a/packages/cli/src/metadataGeneration/metadataGenerator.ts
+++ b/packages/cli/src/metadataGeneration/metadataGenerator.ts
@@ -8,7 +8,7 @@ import { TypeResolver } from './typeResolver';
 import { getDecorators } from '../utils/decoratorUtils';
 
 export class MetadataGenerator {
-  public readonly controllerNodes = new Array<ts.Node>();
+  public readonly controllerNodes = new Array<ts.ClassDeclaration>();
   public readonly typeChecker: ts.TypeChecker;
   private readonly program: ts.Program;
   private referenceTypeMap: Tsoa.ReferenceTypeMap = {};
@@ -59,7 +59,7 @@ export class MetadataGenerator {
       }
 
       ts.forEachChild(sf, node => {
-        if (node.kind === ts.SyntaxKind.ClassDeclaration && this.IsExportedNode(node as ts.ClassDeclaration) && getDecorators(node, identifier => identifier.text === 'Route').length) {
+        if (ts.isClassDeclaration(node) && getDecorators(node, identifier => identifier.text === 'Route').length) {
           this.controllerNodes.push(node);
         }
       });
@@ -225,7 +225,7 @@ export class MetadataGenerator {
 
   private buildControllers() {
     return this.controllerNodes
-      .map((classDeclaration: ts.ClassDeclaration) => new ControllerGenerator(classDeclaration, this))
+      .map(classDeclaration => new ControllerGenerator(classDeclaration, this))
       .filter(generator => generator.IsValid())
       .map(generator => generator.Generate());
   }
--- a/packages/cli/src/metadataGeneration/metadataGenerator.ts
+++ b/packages/cli/src/metadataGeneration/metadataGenerator.ts
@@ -5,9 +5,10 @@ import { ControllerGenerator } from './controllerGenerator';
 import { GenerateMetadataError } from './exceptions';
 import { Tsoa } from '@tsoa/runtime';
 import { TypeResolver } from './typeResolver';
+import { getDecorators } from '../utils/decoratorUtils';
 
 export class MetadataGenerator {
-  public readonly nodes = new Array<ts.Node>();
+  public readonly controllerNodes = new Array<ts.Node>();
   public readonly typeChecker: ts.TypeChecker;
   private readonly program: ts.Program;
   private referenceTypeMap: Tsoa.ReferenceTypeMap = {};
@@ -58,28 +59,9 @@ export class MetadataGenerator {
       }
 
       ts.forEachChild(sf, node => {
-        /**
-         * If we declare a namespace within a module, like we do in `tsoaTestModule.d.ts`,
-         * we need to explicitly get the children of the module declaration
-         * (`declare module 'tsoaTest'`) - which are the moduleBlock statements,
-         * because otherwise our type resolver cannot iterate over namespaces defined in that module.
-         */
-        if (ts.isModuleDeclaration(node)) {
-          /**
-           * For some reason unknown to me, TS resolves both `declare module` and `namespace` to
-           * the same kind (`ModuleDeclaration`). In order to figure out whether it's one or the other,
-           * we check the node flags. They tell us whether it is a namespace or not.
-           */
-          // eslint-disable-next-line no-bitwise
-          if ((node.flags & ts.NodeFlags.Namespace) === 0 && node.body && ts.isModuleBlock(node.body)) {
-            node.body.statements.forEach(statement => {
-              this.nodes.push(statement);
-            });
-            return;
-          }
+        if (node.kind === ts.SyntaxKind.ClassDeclaration && this.IsExportedNode(node as ts.ClassDeclaration) && getDecorators(node, identifier => identifier.text === 'Route').length) {
+          this.controllerNodes.push(node);
         }
-
-        this.nodes.push(node);
       });
     });
   }
@@ -242,8 +224,7 @@ export class MetadataGenerator {
   }
 
   private buildControllers() {
-    return this.nodes
-      .filter(node => node.kind === ts.SyntaxKind.ClassDeclaration && this.IsExportedNode(node as ts.ClassDeclaration))
+    return this.controllerNodes
       .map((classDeclaration: ts.ClassDeclaration) => new ControllerGenerator(classDeclaration, this))
       .filter(generator => generator.IsValid())
       .map(generator => generator.Generate());
--- a/packages/cli/src/metadataGeneration/metadataGenerator.ts
+++ b/packages/cli/src/metadataGeneration/metadataGenerator.ts
@@ -14,10 +14,6 @@ export class MetadataGenerator {
   private referenceTypeMap: Tsoa.ReferenceTypeMap = {};
   private circularDependencyResolvers = new Array<(referenceTypes: Tsoa.ReferenceTypeMap) => void>();
 
-  public IsExportedNode(_node: ts.Node) {
-    return true;
-  }
-
   constructor(entryFile: string, private readonly compilerOptions?: ts.CompilerOptions, private readonly ignorePaths?: string[], controllers?: string[]) {
     TypeResolver.clearCache();
     this.program = !!controllers ? this.setProgramToDynamicControllersFiles(controllers) : ts.createProgram([entryFile], compilerOptions || {});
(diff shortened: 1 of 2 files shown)


simllll:
I'm not quite sure, but I guess TypeResolver.getEnumerateType is another implementation of a "lookup". I didn't run in any troubles during my tests, but as this is the only place left where this.current.nodes is used, I believe we could rewrite this part too and gain some performance for build time. Let me know if you think it's worth it.

Also I stumpled upon getSymbolsInScope, maybe MetadataGenerator.extractNodeFromProgramSourceFiles() could benefit from it too, pseudo code (found at https://github.com/Microsoft/TypeScript/issues/5218#issuecomment-147495613):
```ts
const sourceFile: ts.SourceFile = ...;
const exportedSymbols = typeChecker.getSymbolsInScope(sourceFile, ts.SymbolFlags.Export);

const classSymbol = typeChecker.getSymbolsInScope(exportedSymbols[0].getDeclarations()[0], ts.SymbolFlags.Class)[0];

const className = classSymbol.getName();
```

simllll:
Alright, I looked closer and actually found that in my tests I missed that some enum values have been lost when they are not in the program scope. This is a silent error, which makes it harder to see right away ;).
Anyways, now also enums are resolved via getSymbolAtLocation, this even changes some "anys" to actual types in my outputs ;-)

simllll:
Btw, we could deprecate "@tsoaModel" with this, as I currently see no use case for setting a "preferred" model now anymore.

simllll:
Also refactore dthe metadata generator, and also removed one, now not needed anymore, workaround that got introduced here https://github.com/lukeautry/tsoa/pull/692/files

WoH:
> Btw, we could deprecate "@tsoaModel" with this, as I currently see no use case for setting a "preferred" model now anymore. 

Likely. Only thing I'd like to check are multiple global declarations or use of declaration merging, but I think we tend to just pick the first one anyway so that might be a different kind can of worms all together.

> Also refactored the metadata generator, and also removed one, now not needed anymore, workaround that got introduced here https://github.com/lukeautry/tsoa/pull/692/files

Seems like you already hit most of the locations I thought of when I read the description already. Great job 👍 

I'm on vacation this week and will be busy catching up with work for a few days after, but I'll try to keep the delay as short as possible, but I'll still have to ask you for a bit of patience when it comes to reviewing the details of the code, please take it as a sign of me enjoying my time off, not a lack of interest.

packages/cli/src/metadataGeneration/typeResolver.ts
```diff
@@ -1033,7 +1060,7 @@ export class TypeResolver {
   }
 
   private getNodeDescription(node: UsableDeclaration | ts.PropertyDeclaration | ts.ParameterDeclaration | ts.EnumDeclaration) {
-    const symbol = this.current.typeChecker.getSymbolAtLocation(node.name as ts.Node);
+    const symbol = this.getSymbolAtLocation(node.name as ts.Node);
```
simllll, line 1063: Potentially wrong, as the private method getsymbolatlocation looks up the original Definition, but the node description is probably but the one we want to have from the current one

packages/cli/src/metadataGeneration/metadataGenerator.ts
```diff
@@ -58,28 +59,9 @@ export class MetadataGenerator {
...
-           * we check the node flags. They tell us whether it is a namespace or not.
-           */
-          // eslint-disable-next-line no-bitwise
-          if ((node.flags & ts.NodeFlags.Namespace) === 0 && node.body && ts.isModuleBlock(node.body)) {
-            node.body.statements.forEach(statement => {
-              this.nodes.push(statement);
-            });
-            return;
-          }
+        if (node.kind === ts.SyntaxKind.ClassDeclaration && this.IsExportedNode(node as ts.ClassDeclaration) && getDecorators(node, identifier => identifier.text === 'Route').length) {
```
WoH, line 62:
```suggestion
        if (ts.isClassDeclaration(node) && getDecorators(node, identifier => identifier.text === 'Route').length) {
```

Seems like isExported is basically a noop, also ts now ships with type guards which is a lot easier to read.

simllll: Haha Go and enjoy your holiday 😜 but thanks, i will fix that!

packages/cli/src/metadataGeneration/metadataGenerator.ts
```diff
@@ -5,9 +5,10 @@ import { ControllerGenerator } from './controllerGenerator';
 import { GenerateMetadataError } from './exceptions';
 import { Tsoa } from '@tsoa/runtime';
 import { TypeResolver } from './typeResolver';
+import { getDecorators } from '../utils/decoratorUtils';
 
 export class MetadataGenerator {
-  public readonly nodes = new Array<ts.Node>();
+  public readonly controllerNodes = new Array<ts.Node>();
```
WoH, line 11:
```suggestion
  public readonly controllerNodes = new Array<ts.ClassDeclaration>();
```

packages/cli/src/metadataGeneration/metadataGenerator.ts
```diff
@@ -242,8 +224,7 @@ export class MetadataGenerator {
   }
 
   private buildControllers() {
-    return this.nodes
-      .filter(node => node.kind === ts.SyntaxKind.ClassDeclaration && this.IsExportedNode(node as ts.ClassDeclaration))
+    return this.controllerNodes
       .map((classDeclaration: ts.ClassDeclaration) => new ControllerGenerator(classDeclaration, this))
       .filter(generator => generator.IsValid())
       .map(generator => generator.Generate());
```
WoH, line 230:
```suggestion
    return this.controllerNodes
      .map(classDeclaration => new ControllerGenerator(classDeclaration, this))
      .filter(generator => generator.IsValid())
      .map(generator => generator.Generate());
```

dbryar:
Just a quick heads up, I managed to work through this without editing the code by eliminating the @type modules from the non-TypeScript native dependencies. 

See https://github.com/lukeautry/tsoa/issues/698#issuecomment-923501971

simllll:
> Just a quick heads up, I managed to work through this without editing the code by eliminating the @type modules from the non-TypeScript native dependencies.
> 
> See [#698 (comment)](https://github.com/lukeautry/tsoa/issues/698#issuecomment-923501971)

All of these workarounds should not be needed after this PR anymore :) you could try to make a tsoa cli build with this branch and run your code through it and see if it resolves your issues.
I used npm link for that:.1. checkout this branch 2. yarn install 3. Go to cli package and run sudo npm link 4. Go to your sources and run npm link @tsoa/cli ... 5. Generate routes & definition files ;)

packages/cli/src/metadataGeneration/typeResolver.ts
```diff
@@ -471,27 +471,50 @@ export class TypeResolver {
...
+
+    const allNodes = declaredType.getDeclarations() as ts.DeclarationStatement[];
+
+    let enumNodes = allNodes.filter(node => {
+      if (!this.nodeIsUsable(node) || !this.current.IsExportedNode(node)) {
+        return false;
+      }
+
+      return node.kind === ts.SyntaxKind.EnumDeclaration && node.name?.text === enumName;
+    }) as ts.EnumDeclaration[];
```
WORMSS, line 499:
```suggestion
    let enumNodes = allNodes.filter((node): node is ts.EnumDeclaration => {
      if (!this.nodeIsUsable(node) || !this.current.IsExportedNode(node)) {
        return false;
      }

      return node.kind === ts.SyntaxKind.EnumDeclaration && node.name?.text === enumName;
    });
```

WORMSS (changes requested):

packages/cli/src/metadataGeneration/typeResolver.ts
simllll: mhm I know this is subjective, but I see my versoin more correct, as the node that is passed into the filter function is not for sure a ts.EnumDeclaration yet, we only can say so after the node.kind === ts.SyntaxKind.EnumDeclaration check, but I will look into if I can find some pre defined guard for it

WORMSS: ... I didn't "type" the incoming node,.
That is still its orginal type defined by the inference of the array.

I put the type on the return type as a guard function. 

You can do the same with functions that throw errors if something is not correct. In that case it would be
```ts
assertNode(node): asserts node is ts.EnumDeclation {
  ....
}
```

packages/cli/src/metadataGeneration/typeResolver.ts
```diff
@@ -761,57 +784,15 @@ export class TypeResolver {
...
-    const statements: any[] = this.resolveModelTypeScope(leftmostIdentifier, this.current.nodes);
-
     const typeName = type.kind === ts.SyntaxKind.Identifier ? type.text : type.right.text;
 
-    let modelTypes = statements.filter(node => {
+    const symbol = this.getSymbolAtLocation(type);
+    const declarations = symbol?.getDeclarations();
+
+    let modelTypes: UsableDeclarationWithoutPropertySignature[] = declarations?.filter(node => {
       if (!this.nodeIsUsable(node) || !this.current.IsExportedNode(node)) {
```
WoH, line 796: `nodeIsUsable` should do the type guarding so we can get rid of the `as UsableDeclaration` in line 800

packages/cli/src/metadataGeneration/typeResolver.ts
simllll: ah right, sorry now I get your idea :-)!

packages/cli/src/metadataGeneration/typeResolver.ts
WoH: Do you have an example where we can trigger a difference?

simllll: I actually was playing around with it already and tried to find some difference, I couldn't find any issues. Therefore I think my comment can be silently closed ;-)

simllll: I was wrongly assuming that this could be a problem:
image
but as we are looking for "CommonLink" and not the type of commonlink for the description itself, all is good.

WoH:
LGTM, thanks for woking on this!

simllll:
Do you have any eta in mind for the next release? :)

WoH:
No ETAs, but there are several non-breaking PRs in the pipeline that would make sense, as they're 1 or 2 small changes from merge. Looking at #1100, #1096, #1101 (yet to be reviewed) and maybe #1053 if we can get some kind of regression test added.

PR #1096, opened 2021-09-29
feat(metadataGeneration): add support for imported symbols

### LINKED ISSUE #1079: @Example doesn't read value from an imported variable
## Sorting

- **I'm submitting a ...**

  - [x] bug report
  - [ ] feature request
  - [ ] support request

- I confirm that I

## Expected Behavior

This is a repost of issue https://github.com/lukeautry/tsoa/issues/834 because I don't think it has been addressed. 

`@Example()` should be able to read a variable value imported from another file.

With

```ts
// file a.ts
export const example = {someKey: 'someValue'}
// file b.ts
import example from a
...
@Example(example)
...
```
tsoa should generate the correct response example from the imported variable.

## Current Behavior

It now creates an empty object `{}` if the variable is imported. It works if the example value is directly supplied to the decorator or defined in the same file.

Being forced to give the example in the same file is particularly inconvenient for very long jsons that can make the file quickly unreadable.

must:
closes #1079

### All Submissions:

- [x] Have you written integration tests?

**Test plan**

The integration tests test for an example with an example value from an imported symbol.

### DIFF
--- a/packages/cli/src/metadataGeneration/initializer-value.ts
+++ b/packages/cli/src/metadataGeneration/initializer-value.ts
@@ -1,7 +1,7 @@
 import * as ts from 'typescript';
 import { Tsoa } from '@tsoa/runtime';
 
-export const getInitializerValue = (initializer?: ts.Expression, typeChecker?: ts.TypeChecker, type?: Tsoa.Type) => {
+export const getInitializerValue = (initializer?: ts.Expression | ts.ImportSpecifier, typeChecker?: ts.TypeChecker, type?: Tsoa.Type) => {
   if (!initializer || !typeChecker) {
     return;
   }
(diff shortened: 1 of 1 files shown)


packages/cli/src/metadataGeneration/initializer-value.ts
```diff
@@ -48,10 +48,16 @@ export const getInitializerValue = (initializer?: ts.Expression, typeChecker?: t
...
+    case ts.SyntaxKind.ImportSpecifier:
+      const importSymbol = typeChecker.getSymbolAtLocation((initializer as ts.ImportSpecifier).name)!;
+      const aliasedSymbol = typeChecker.getAliasedSymbol(importSymbol);
+      const dec = aliasedSymbol.getDeclarations()![0] as any;
+      return getInitializerValue(dec.initializer ? dec.initializer : undefined, typeChecker);
     default:
       const symbol = typeChecker.getSymbolAtLocation(initializer);
       const extractedInitializer = symbol && symbol.valueDeclaration && hasInitializer(symbol.valueDeclaration) && (symbol.valueDeclaration.initializer as ts.Expression);
-      return extractedInitializer ? getInitializerValue(extractedInitializer, typeChecker) : undefined;
+      const extractedImportSpecifier = symbol?.declarations && symbol.declarations.length > 0 && ts.isImportSpecifier(symbol.declarations[0]) && (symbol.declarations[0] as ts.ImportSpecifier);
```
WoH, line 59: Can you split up this 1 liner into multiple ones.

WoH (approved):
Approach looks fine, some smaller corrections would be helpful

must:
Just pushed a refactored version that hopefully addresses your feedback. If you got any changes you'd like to see through just let me know!

must:
@WoH a random question, is there a way to run a specific part of the test suit (i.e. specific test or filter by name like jest allows you to do)?

WoH:
> @WoH a random question, is there a way to run a specific part of the test suit (i.e. specific test or filter by name like jest allows you to do)?

Have a look at the vscode debug presets. Basically, you can call mocha and pass the specific file as an argument instead of a wildcard that matches everything.

packages/cli/src/metadataGeneration/initializer-value.ts
```diff
@@ -48,11 +48,20 @@ export const getInitializerValue = (initializer?: ts.Expression, typeChecker?: t
         nestedObject[p.name.text] = getInitializerValue(p.initializer, typeChecker);
       });
       return nestedObject;
+    case ts.SyntaxKind.ImportSpecifier:
+      const importSpecifier = initializer as ts.ImportSpecifier;
+      const importSymbol = typeChecker.getSymbolAtLocation(importSpecifier.name)!;
```
WoH, line 53: Is there a scenario where we have a program that compiles fine but returns `undefined` here?

must: I pushed a quick fix to just return if it's undefined. Not familiar enough with the Typescript compiler API to know how to deal with this better.

WoH: I don't see any new commits, am I missing something? If you're handling `undefined`, I'd expect the assertion to be gone:

```suggestion
      const importSymbol = typeChecker.getSymbolAtLocation(importSpecifier.name);
```

must: Sorry looks like I didn't end up pushing the amended commit when I thought I did

PR #1100, opened 2021-10-01
[Issues #1085] Property level vendor extension

### LINKED ISSUE #1085: Property Level Vendor Extension
## Sorting

- **I'm submitting a ...**

  - [ ] bug report
  - [x] feature request
  - [ ] support request

- I confirm that I

## Expected Behavior
Allow us to generate vendor extension on property's attribute:
```yaml
    User:
      properties:
        name:
          type: string
          x-version: V3.36
```

```ts
export class User {

  /**
   * @example "John Doe"
   * */
   @Extension("x-version", "V3.36.0")
  name: string

}
```

## Current Behavior

Currently the `@Extension` annotation will be ignored and nothing happen
```yaml
    User:
      properties:
        name:
          type: string
```

## Possible Solution
`tsoa` should also consume the `@Extension` in the property model.

I've made an adjustment to `tsoa` so it can support those needs [on my fork](https://github.com/DjakaTechnology/tsoa/tree/feature/property-extension)

## Context (Environment)
Version of the library: v3.11.2

## Detailed Description
Sometime we need to put a vendor extension on the property level, usually its for documentation purpose or internal tools needs. One of the example is [Microsoft's mutability](https://github.com/Azure/autorest/blob/main/docs/extensions/readme.md#x-ms-mutability) where it used to define mutabiltiy rule of the property.

My team currently using standard yaml and considering to migrate to tsoa for spec generation only. Our team require information of `Added since` so we create `x-version` as vendor extension so the documentation will look like this:

image 

but during the assesment it seems `tsoa`  doesn't support property level vendor extension yet. 

## Breaking change?
No

DjakaTechnology:
### All Submissions:

**Closing issues**

closes #1085

**Test plan**

Unit test will generate vendor extension on property and we validate the generation result


packages/cli/src/metadataGeneration/extension.ts
```diff
@@ -32,3 +30,34 @@ export function getExtensions(decorators: ts.Identifier[], metadataGenerator: Me
...
+  return extensions;
+}
+
+function validateExtensionKey(key: string) {
+  if (key.indexOf('x-') !== 0) {
+    throw new Error('Extensions must begin with "x-" to be valid. Please see the following link for more information: https://swagger.io/docs/specification/openapi-extensions/');
+  }
+}
+
+function safeFromJson(json: string) {
```
WoH, line 57: Can you move this to utils and reuse it [here](https://github.com/lukeautry/tsoa/blob/master/packages/cli/src/metadataGeneration/typeResolver.ts#L1091)

packages/cli/src/utils/jsDocUtils.ts
```diff
@@ -18,6 +18,18 @@ export function getJSDocComment(node: ts.Node, tagName: string) {
   return tags[0].comment;
 }
 
+export function getJSDocComments(node: ts.Node, tagName: string) {
```
WoH, line 21: This seems like a generalization of `getJSDocComment`, let's expose `getJSDocComment` as a wrapper of this new method that returns only the first element.

Issue #1080, opened 2021-09-15
Fileupload using octet-stream


dderevjanik:
## Sorting

- **I'm submitting a ...**

  - [ ] bug report
  - [x] feature request
  - [ ] support request

- I confirm that I
  - [x] used the [search](https://github.com/lukeautry/tsoa/search?type=Issues) to make sure that a similar issue hasn't already been submit

## Expected Behavior

Right now, is not possible to upload file using `octet-stream`, only by using `@UploadedFile` which is `multipart-formdata`. Would be great to have option to upload files using `octet-stream` (for big file upload). 

```yaml
requestBody:
  content:
    application/octet-stream:
      schema:
        type: string
        format: binary
```

could be generated from this:

```ts
    @Post("/upload")
    async streamUpload(@Request() req: Req, @Body() body: Readable): Promise<void> {
       return new Promise((resolve, reject) => {
          body.on("data", (chunk) => console.log(chunk));
          body.on("error", (chunk) => reject());
          body.on("end", (chunk) => resolve());
       });
    }
```

TSOA could automatically detects that requesting using `@Body() body: Readable`  should be `octet-stream`.

## Current Behavior

Currently, TSOA will generate (from code above)

```jsonc
"requestBody": {
	"required": true,
	"content": {
		"application/json": {   // this should be "application/octet-stream"
			"schema": {
				"type": "string", 
				"format": "byte"
			}
		}
	}
}
```

falk-stefan:
+1

PR #1115, opened 2021-10-23
Bugfix/fix Proper handling for upload/download using octet-stream

dderevjanik:
Similar to https://github.com/lukeautry/tsoa/issues/1080 but instead of creating `application/octet-stream` for uploading file, this fix is related to downloading file

Currently:

```ts
    @Get("/")
    async download(@Request() req: Req): Promise<Readable> {
       return new Readable();
    }
```

Produces:

```yaml
          content:
            application/json:
              schema:
                type: string
                format: byte
```

Instead of: 

```yaml
          content:
            application/octet-stream:
              schema:
                type: string
                format: binary
```

*Note: using swagger-codegen can produce clients with an incorrect response (e.g. expecting application/json based on spec, but instead getting stream from server)* 

### All Submissions:

**Test plan**


dderevjanik:
@WoH I made several updates to support also uploading files using `octet-stream`. Several notes regarding those changes:

- Now it supports also uploading a file by using `octet-stream` https://github.com/lukeautry/tsoa/issues/1080
- According to [openapi 3 specification](https://swagger.io/docs/specification/describing-request-body/file-upload/), to upload a file, we should use `format: binary` or `format: base64`. Actual TSOA version is using `Buffer` type for `Readable` stream -> I changed it to `binary` to automatically generate `format: binary` instead of `format: byte`. I generated several clients using openapi-codegen utils to make sure that I should use `binary` instead of `byte`. 
- Because of those changes (different TsoaType for Readable stream, different content-type instead of default `application/json`), there could be possible breaking changes for users (hopefully not).
- Still I'm not sure how to handle Swagger 2.0 specification properly, because it seems that Swagger 2.0 supports only `multipart-formdata`, which is completely different case. Any ideas?

WoH:
> * Because of those changes (different TsoaType for Readable stream, different content-type instead of default `application/json`), there could be possible breaking changes for users (hopefully not).

There are other changes that will require a major version bump to be on the safe side (like this one), which I'll merge for an alpha v4, so that's fine.

> * Still I'm not sure how to handle Swagger 2.0 specification properly, because it seems that Swagger 2.0 supports only `multipart-formdata` for uploading a file, which is completely different case. Any ideas?

File Download:

```yaml
  /path:
    get:
      responses:
        "200":
          description: file download
          schema:
            type: file
```

File Upload:
This seems perfectly valid:

```yaml
paths:
  /path:
    post:
      consumes:
      - application/json
      produces:
      - application/json
      parameters:
      - in: body
        name: file
        schema:
          type: string
          format: binary
      responses:
        "200":
          description: Ok
```

falk-stefan:
Is this supported by now? :thinking:

PR #1117, opened 2021-10-31
Fixed usage of correct types in enums

### LINKED ISSUE #1068: Typescript union of `number` generates `string` enum in swagger
## Sorting

- **I'm submitting a ...**

  - [x] bug report
  - [ ] feature request
  - [ ] support request

- I confirm that I

## Expected Behavior

Given a type defined as 

```ts
export interface MyAnswersRequest {
  readonly answers: {
    readonly answer1: AnswerValue;
  };
}

export type AnswerValue = 0 | 1 | 2 | 3;
```

I would expect it to generate OpenApi.yaml schemas such as : 

```yaml
        AnswerValue:
            type: number
            enum:
                - 0
                - 1
                - 2
                - 3
        MyAnswersRequest:
            properties:
                answers:
                    properties:
                        answer1:
                            $ref: '#/components/schemas/AnswerValue'
                    required:
                        - answer1
                    type: object
            required:
                - answers
            type: object
            additionalProperties: false
```

## Current Behavior

Instead, the generated enum members are generated as `string`s (with single quotes)

```yaml
        AnswerValue:
            type: number
            enum:
                - '0'
                - '1'
                - '2'
                - '3'
```

Our OpenApi linter (Spectral) complains (and it seems right) that even though the type of this enum is `number` , the listed acceptable values are `string`s 

Note that this only seems to impact the generated swagger file, but validation of incoming requests work properly.

## Steps to Reproduce

1. 2. 3. 4.

## Context (Environment)

Version of the library: `"@tsoa/runtime": "3.9.0",` / `"@tsoa/cli": "3.9.0",`
Version of NodeJS: `v14.16.0`

## Breaking change?

This might be a breaking change for people who rely on unions of `number` ... but this would only impact the generated swagger file, not the runtime behavior.

FlorianRappl:
### All Submissions:

- [x] Have you written unit tests? (adjusted ex. tests)
- [x] Have you written unit tests that cover the negative cases (i.e.: if bad data is submitted, does the library respond properly)? (adjusted ex. tests)

**Closing issues**

Put `closes #1068` in your comment to auto-close the issue that your PR fixes.

(in some sense it yes, so I'll fill this one out, too)

Only targets OAS 3, but this seems to be the target of the issue (and is the current / default spec). It might be "breaking", but then again, it's only the OAS output and not the validator. Since this one even follows OAS more closely than beforehand I doubt it should have any negative impact though. Tooling etc. would like the "new style" even more than the old one.

**Test plan**

Some existing tests only build upon strings, essentially dealing with it the same way that the bug appeared in the first place. I've adjusted them to the desired behavior.


packages/cli/src/swagger/specGenerator3.ts
```diff
@@ -639,7 +652,7 @@ export class SpecGenerator3 extends SpecGenerator {
     if (types.size === 1) {
       const type = types.values().next().value;
       const nullable = enumType.enums.includes(null) ? true : false;
-      return { type, enum: enumType.enums.map(member => (member === null ? null : String(member))), nullable };
+      return { type, enum: enumType.enums.map(member => (member === null ? null : getValue(type, member))), nullable };
```
WoH, line 655:
```suggestion
      return { type, enum: enumType.enums.map(member => getValue(type, member)), nullable };
```

packages/cli/src/swagger/specGenerator3.ts
```diff
@@ -5,6 +5,19 @@ import { convertColonPathParams, normalisePath } from './../utils/pathUtils';
 import { SpecGenerator } from './specGenerator';
 import { UnspecifiedObject } from '../utils/unspecifiedObject';
 
+function getValue(type: 'string' | 'number' | 'integer' | 'boolean', member: any) {
```
WoH, line 8: Handle null

WoH (commented):
Small changes, but mainly one request: There's a similar enum handling happening for OpenAPI 2 !(https://github.com/lukeautry/tsoa/blob/master/packages/cli/src/swagger/specGenerator2.ts#L465)[here]. Please make sure we don't diverge when we don't have to.

WoH:
Please rebase so the CI fixes on the main branch are applied.

packages/cli/src/swagger/specGenerator3.ts
FlorianRappl: Isn't `null` handled implicitly already? -> The `getValue` cannot be called with `null`.

packages/cli/src/swagger/specGenerator3.ts
FlorianRappl: Hm my main objective was to fix the issue - any reason for putting the `null` handling in the `getValue`? We can do it, just wanted to know "why" (and why in the context of that issue).

FlorianRappl:
> Small changes, but mainly one request: There's a similar enum handling happening for OpenAPI 2 here. Please make sure we don't diverge when we don't have to.

I am not sure about OAS 2 - the issue was placed for OAS 3 and I am not sure how the enum handling is supposed to look for OAS 2. If its similar then we can apply it there.

packages/cli/src/swagger/specGenerator3.ts
WoH: I think it's better to handle that functionality in one place if possible and move the conditional out of the already pretty lone one-liner. Not a cause for rejection, but it'd be preferable.

WoH:
> I am not sure about OAS 2 - the issue was placed for OAS 3 and I am not sure how the enum handling is supposed to look for OAS 2. If its similar then we can apply it there.

Should follow the same logic

FlorianRappl:
Alright, great - all done @WoH ! Thanks for the suggestions :rocket:

PR #1123, opened 2021-11-04
allow custom middlewares for controllers and methods

### LINKED ISSUE #47: Middleware decorators
Hello,

it would be great to have some kind of Middleware decorators, where you can give middlewares to run before (or even after) a controller method. Middlewares could be on a separate layer or even special controller methods.

e.g.

```typescript
@Route('Users')
export class UsersController {

    @Middleware
     public validate(request: express.Request, response: express.Request, next: function) {
        // TODO: implement a middleware
    }

    @Middleware
     public render(request: express.Request, response: express.Request, next: function) {
        // TODO: implement a rendering middleware
    }

    @MiddlewareBefore(UserController.validate)
    @MiddlewareAfter(UserController.render)
    @Get('{id}')
    public async getUser(id: number, @Request() request: express.Request): Promise<User> {
        // TODO: implement some code that uses request as well
    }
}
```

With this feature, existing Node.js projects would be much easeier to migrate into a tsoa project and also middleware pattern is essential part of Node.js applications.

It's could be an issue how to handle middleware chains, so e.g. `MiddlewareBefore` could contain an array of middlewares, or multiple decorators could be attached.

If we discuss the details I can help with implementation.
Thanks!

### LINKED ISSUE #62: Use external middlewares in tsoa
I wonder if there is some simple way to use external libraries (such as `passport.js`) with `tsoa`. I saw injection of `req`, but often we need `res` and `next` as well. The ideal situation for me is to pass `req res next` to some middleware (in i.e. `/services`) that would implement logic of the endpoint.

### LINKED ISSUE #624: Custom middleware injection
## How to inject custom middleware for certain routes ?
https://github.com/tarakeshp/stripe-node/blob/master/examples/webhook-signing/typescript-node-express/express-ts.ts
In the line number 34,  bodyParser.raw({type: 'application/json'}) is a middleware injected in the request pipleline.

app.post(
  '/webhook',  **bodyParser.raw({type: 'application/json'}),**
  (req: express.Request, res: express.Response): void => {
....
}

Likewise how can I pass to tsoa controller/action
@Post("subscriptions/webhook")
  public async SubscriptionWebhook(@Request() httpContext: HttpContext): Promise<IBillingResponse> {
......
}

### LINKED ISSUE #948: How to use custom middleware?
Currently, we are moving a lot of non business logic in middleware
and using them per route, for example:

```
router.post('/some/path', accountValidator, userValidator, prometheusMetrics, someOtherMiddleware, myController.getSomethingUseful);
```
How can i add custom middlewares to automatically generated routes.ts?
Are there any other ways like custom decorators or base controller that I should consider using with tsoa?
I think I'm fine migrating custom middlewares to something else but I don't understand what is an alternative tsoa way of doing this?

## Sorting

- **I'm submitting a ...**

  - [ ] bug report
  - [x] feature request
  - [x] support request

- I confirm that I

## Expected Behavior

I expected to find some 
`@Middleware` decorator or a guide how to create custom decorators

## Current Behavior

@Route decorator doesn't have options parameter where i could add functions or middleware to execute before or after main haindler

## Possible Solution

* add an options to @Route ('/my/path', { before: promMetrics } )
* add a custom @Middleware decorator
* add a documentation how to add custom shareable code

- Confirm you were using yarn not npm: [ npm]

cheng81:
# Middleware decorators

This PR introduces a bunch of new decorators (and small changes to the route templates) that allows the usage of `express`, `koa`, and `hapi` middlewares at the route and individual method level.
Different middlewares can be specified depending on the server technology used (not completely sure this is actually useful, but it helps `type`-wise).

Example:

```typescript
import { Route, Get, Middlewares, ExpressMiddlewares } from '@tsoa/runtime';
@Middlewares({
  express: [
    middleware1,
    middleware2,
  ],
  koa: [
    koaMiddleware1,
  ],
})
@Route('Foo')
export class MiddlewareTestController {
  @ExpressMiddlewares(
    middleware3,
    middleware4,
  )
  @Get('/foo')
  public async foo(): Promise<void> {
    return;
  }
}
```

### All Submissions:

- [ ] Have you written unit tests that cover the negative cases (i.e.: if bad data is submitted, does the library respond properly)? **no, but all the other Controllers do not have middlewares, and they continue to work**

**Closing issues**

closes #948 
closes #624 
closes #62
closes #47 

I have some quirks with the current solution:
1. It's not elegant - `middlewares` are "stored" in controllers constructor and method functions, and while this works (and has been used somewhat in `tsoa` already - I'm looking at you `request.user` defined in the `authentication` middleware), it also feels like an hack rather than a proper solution. If people have a better idea in mind, please do share!
2. The various `middleware` types are super vague (e.g. `ExpressMiddleware` is defined as `(req: any, res: any, next: any) => Promise<any>;`) since the various server libraries are not dependencies of `tsoa`. While I don't think it'll cause problems in practice, I would love if there was a way to make them more precise (perhaps requiring just the `types` libraries?).
3. I also have a bit of a problem with ergonomics: in theory I could define a single `Middleware` decorator that accepts `...middlewares: any[])` and be done with that: in practice I'm pretty sure that all controllers are created for a single server technology, so having `@Middlewares({express: [...], koa: [...]})` is really there just for the benefit of the tests.
4. finally, perhaps middlewares should really be a configuration of other decorators: `Route` and the various `http` methods (`Get`, `Post`, and so on).

**Test plan**

I added a `MiddlewareTest` controller with fake middlewares and check in the tests that they are indeed executed.


packages/runtime/src/decorators/middlewares.ts
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * Install `express`, `koa` and `hapi` middlewares
+ * to the Controller or a specific method.
+ * @param middlewares
+ * @returns
+ */
+export function Middlewares(middlewares: Middlewares) {
+  return decorator(target => {
+    target._expressMiddlewares = middlewares.express || [];
```
WoH, line 45: I'd prefer using the reflect API here.
It seems like you're reimplementing some of the concepts that API offers manually in the template etc. Is there a reason why your approach is preferable that I'm missing?

cheng81: Hi, I'll try to take a look at the reflect API and let you know.
I'm not sure what I am reimplementing, if there's already a way to set custom middlewares for the various servers please let me know.
I should add that I'd really prefer to not use a custom routes template, and this feature seems to have been requested a number of times, so it seems to me that it's something the `tsoa` users would like.

WoH: What this PR is doing is patching the target with some semi-transparent properties.
However, there's an API that works well and hides all of that patching: `Reflect.defineMetadata` and `Reflect.getMetadata`.

https://rbuckton.github.io/reflect-metadata/

cheng81: @WoH now I see what you mean, I didn't know about `reflect-metadata`, I updated the code to make use of it.

packages/runtime/src/decorators/middlewares.ts
```diff
@@ -0,0 +1,125 @@
+import 'reflect-metadata';
```
WoH, line 1:
```suggestion
```

This should be at the top of the index, at this point it may be too late

packages/runtime/src/decorators/middlewares.ts
```diff
@@ -0,0 +1,125 @@
...
+export type ExpressMiddleware = (req: any, res: any, next: any) => Promise<any>;
+export type KoaMiddleware = (ctx: any, next: any) => Promise<any>;
+export type HapiMiddlewareBase = (request: any, h: any) => Promise<any>;
+export type HapiMiddlewareSimple = HapiMiddlewareBase | { method: HapiMiddlewareBase; assign?: string; failAction?: HapiMiddlewareBase | string };
```
WoH, line 5: Would prefer if we could try to type this properly as they are visible

packages/runtime/src/decorators/middlewares.ts
```diff
@@ -0,0 +1,125 @@
...
+/**
+ * Helper function to create a decorator
+ * that can act as a class and method decorator.
+ * @param fn a callback function that accepts
+ *           the subject of the decorator
+ *           either the constructor or the
+ *           method
+ * @returns
+ */
+function decorator(fn: (value: any) => void) {
```
WoH, line 27: TS has a MethodDecorator and ParameterDecorator type.
While we have historically not been very diligent, I think it makes sense to type the returns if it's not too complicated

WoH (commented):
Mostly ready, I have some minor inconveniences when it comes to maintenance and UX for the end users. If you could try to tighten the typing a bit more, that'd be very appreciated.

packages/runtime/src/decorators/middlewares.ts
cheng81: I'll try to do it, though that means I'll have to add other dependencies (the typing of `express`, `koa` and `hapi`) to the project.
While this means we'll need to keep up-to-date with those dependencies too, at least we'll know when/if they introduce non-retro compatible changes in their middleware definitions (not that I expect those to happen, but in theory they could do something very asinine like, e.g., putting `next` as 1st parameter instead of the `3rd` for the `express` middleware)

packages/runtime/src/decorators/middlewares.ts
cheng81: 👍  actually I just typed `decorator` to return `ClassDecorator & MethodDecorator` and it works (not sure how `typescript` is able to understand that, though)

packages/runtime/src/decorators/middlewares.ts
WoH: That's probably worth the
a) DX
b) as you mentioned, the chance to detect potential issues earlier

dderevjanik: TBH, I wouldn't mind to have middlewares not typed (using `any`). I think it's better solution than having dependencies on express, koa and hapi types and maintaining those dependencies...

dderevjanik:
@cheng81 Sorry for asking, but what's reason behind using framework specific decorator-middlewares? 
Do you need that kind of information in api-generation or in runtime, or its only because of type inference?

AFAIK, for type inference you can use generics in your `@Middleware` decorator. e.g. for express:

```ts
import { RequestHandler } from 'express'; // Importing type from express instead of TSOA
import { Route, Get, Middlewares } from '@tsoa/runtime';

@Middlewares<RequestHandler>(middleware1)
@Route('Foo')
export class MiddlewareTestController {

  @Middlewares(middleware2, middleware3)
  @Get('/foo')
  public async foo(): Promise<void> {
    return;
  }
}
```

cheng81:
> @cheng81 Sorry for asking, but what's reason behind using framework specific decorator-middlewares? Do you need that kind of information in api-generation or in runtime, or its only because of type inference?
> 
> AFAIK, for type inference you can use generics in your `@Middleware` decorator. e.g. for express:
> 
> ```ts
> import { RequestHandler } from 'express'; // Importing type from express instead of TSOA
> import { Route, Get, Middlewares } from '@tsoa/runtime';
> 
> // Using generic to determine correct type of middleware
> @Middlewares<RequestHandler>(middleware1)
> @Route('Foo')
> export class MiddlewareTestController {
> 
>   @Middlewares<RequestHandler>(middleware2, middleware3)
>   @Get('/foo')
>   public async foo(): Promise<void> {
>     return;
>   }
> }
> ```

@dderevjanik no worries, you shouldn't be sorry for asking!
So the main reason is basically security (so yes, type checking more than type inference), although I do see that in practice that would be probably a non-existent problem: in theory, by using a single `Middlewares` decorator, it would be possible to add, say, an `express` middleware and generate the routes for, say, `koa`. Of course this would be quickly caught by the devs.

Even using a single middleware with generics would be not ideal (in my head, at least): nothing prevents someone from, say, using `@Middlewares<string>` instead of the right type - again, this is stupid and no one in their sane mind would do that, but the fact that the system _allows_ that kinda bothers me.

Perhaps there's a way to define `Middleware<T: ExpressMiddleware | KoaMiddleware | HapiMiddleware>`  but we go back to square one: I need those types defined in `tsoa`.

If we follow the same patter `tsoa` has taken so far, then I'd just define them as `any` (since the compiled `routes` files has a tendency of using `any` everywhere), but honestly I'm not too much happy with it.

Also as I mentioned in the previous comment, in some way, the version of the framework used is already fixed: the code used in the route templates relies on the version it has been used, and should a major version change some of the code the `routes.ts` file uses, it would lead to nasty surprises.

Finally, of course I can adjust and implement a solution everybody can agree on, I just wanted to present all the reasoning behind what I came up with.

dderevjanik:
@cheng81 I see and completely understand what are you trying to achieve. I really like idea of having custom middlewares in TSOA (thank you opening PR 🚀 ). My opinions are highly subjective.

Personally, for me, its a little bit confusing to have defined property `express` inside all middleware decorators like 
`@Middlewares({ express: [ ] })` or having decorators only for specific framework like `@ExpressMiddlewares([])`. It seems that there's a lot of complexity behind so many specific middlewares and generic one. 

Some pros/cons with current approach: 
+ Correctly typed middlewares for different frameworks
- Confusing to have `@Middlewares({ express: [] })` and `@ExpressMiddlewares([])`. Why not only one?
- Having type dependencies (internal/external) on `Express`, `Koa` and `Hapi` in TSOA
- A lot more complexity in TSOA code (than having only one generic middleware)

Don't get me wrong, but I would like to suggest to split this PR into 2 iteration, or separated PRs.  
- First PR: Having generic `@Middlewares<F extends Function>([])` decorator seems like first iteration, easy to implement, easy to understand and super easy to use 👍 . Not type-safe (depends on user) 👎 
- Second PR: Having type-safe `@Middlewares<`, which could provide type safety besides easy to use approach 👍 . We can have further research or discussion.

cheng81:
@dderevjanik @WoH this might be overkilling, but what about having different packages (e.g. `@tsoa/middleware-express`, etc) that all exports a single `Middleware` decorator with the right `typing`? As I said, probably overkilling, but just an idea.

I'll possibly implement for now what @dderevjanik suggested (though it'll need to be something like `@Middlewares<T extends Function | Object>` since `hapi` middlewares are.. well, a bit peculiar).

WoH:
I'd rather see if we can use declaration merging in that case (not having all 3 types)

cheng81:
> I'd rather see if we can use declaration merging in that case (not having all 3 types)

@WoH  I tried to look at declaration merging, but I'm failing to see how it would help here, could you perhaps elaborate a bit?

WoH:
> > I'd rather see if we can use declaration merging in that case (not having all 3 types)
> 
> @WoH I tried to look at declaration merging, but I'm failing to see how it would help here, could you perhaps elaborate a bit?

This may not work because of the non-ambient vs. ambient context as is, but the idea is roughly to produce:
```ts
export declare function ExpressMiddlewares(...middlewares: any[]): ClassDecorator & MethodDecorator;
```

and, as a user, I do:

```ts
declare module '@tsoa/runtime' {
  export function ExpressMiddlewares(...middlewares: Array<import('express').RequestHandler>): ClassDecorator & MethodDecorator;
}
```

cheng81:
@WoH yeah, I tried and no matter what, it kept complaining about ambient vs non-ambient errors.
In principle we could also generate an `Middlewares` function in the `routes.ts`, though that has some "chicken and egg" problem.
Anyhow, I changed the PR and removed the types dependencies, and simplified the whole thing by having a single `Middlewares<T>` function, let me know what you and @dderevjanik think!

packages/runtime/package.json
```diff
@@ -26,6 +26,7 @@
   "license": "MIT",
   "dependencies": {
     "promise.any": "^2.0.2",
+    "reflect-metadata": "^0.1.13",
```
WoH, line 29:
```suggestion
```

packages/runtime/src/index.ts
```diff
@@ -1,3 +1,4 @@
+import 'reflect-metadata';
```
WoH, line 1:
```suggestion
```

tests/prepare.ts
```diff
@@ -110,9 +110,6 @@ const log = async <T>(label: string, fn: () => Promise<T>) => {
           middleware: 'koa',
           routesDir: './fixtures/koa',
         },
-        undefined,
-        undefined,
-        metadata,
```
WoH, line 115: What makes this change necessary?

cheng81: For some reason, now that I used a different controller for each framework, the `koa` generated `routes.ts` file kept having the `MiddlewareExpressController` imported instead of `MiddlewareKoaController`.
I figured it must be something about those `metadata` (the `tsoa.json` file used to generate those is configured for the `express` server), so I tried removing those, and the tests suddenly passed..

packages/runtime/src/index.ts
cheng81: heh, without that my editor complains that `Reflect.defineMetadata` and `Reflect.getMetadata` do not exists

WoH: this line should be at the beginning of the entry point of a program, this import should at best be useless.

cheng81: I'm not sure I understand why that should be the case (even if the library in `node_modules`, if I remove the line, the polyfill is not loaded so the build fails), but I'm more than willing to fix it if it's a problem, so where should I put it instead?

WoH (approved):

dderevjanik:
@cheng81 It looks amazing! Thank you 👍

kshay:
Hi, looking forward to using this—just wondering if it will be included in a 3.x release or is being held until 4.0? Thanks!

michaelvalasanyan:
I'm not able to get this in latest npm package ... is there a timeline for releasing it ?

WoH:
https://github.com/lukeautry/tsoa/releases/tag/v4.0.0-alpha.0

saqibarfeen:
I installed it via yarn add tsoa@next , but my middleware is not getting called, am I missing something?
```
@Route("ping")
@Tags("Ping")
export class PingController extends Controller {
    @Get()
    @Middlewares(async (req, res, next) => {
        console.log("before mw"); // this doesn't get called before the getPing controller
        next();
    })
    public async getPing(@Request() request: express.Request): Promise<any> {
        const resp: Response = { errors: [] };
        this.setStatus(200);
        return {};
    }

**Comment by RevanthGovindan** [comment]
@saqibarfeen same issue for me as well, middleware doesn't called.

**Comment by WoH** [comment]
@RevanthGovindan @saqibarfeen Please make sure you have Reflect API in scope and `experimentalDecorators` on.
If you can post a reproduction case, please open up an issue.

**Comment by RevanthGovindan** [comment]
@WoH 
  "experimentalDecorators": true in tsconfig.json and import 'reflect-metadata'; in index file, still middleware doesn't called

**Comment by curtisblanchette** [comment]
After installing 4.0.0-rc.0 and generating routes it appears that the swagger type definition file has not received some template literal replacements.

``` bash 
Error: node_modules/@tsoa/runtime/dist/swagger/swagger.d.ts(203,10): error TS1023: An index signature parameter type must be either 'string' or 'number'.
Error: node_modules/@tsoa/runtime/dist/swagger/swagger.d.ts(262,10): error TS1023: An index signature parameter type must be either 'string' or 'number'.
```

The interfaces in question are `BaseSchema` and `Operation3`  L195 and L248
``` typescript

  ...
  [ext: `x-${string}`]: unknown;
}
```

If these can be safely removed let me know and I will submit a PR.

LouisCharles70:
Hi @cheng81, do you have a working example with the feature you just added ? Can't find a way to make it work somehow...

kempsteven:
in what version is this in?

AnthonySLWhite:
Do we have a fix on this issue yet? :arrow_up:

WoH:
Any repro yet?

PR #1127, opened 2021-11-08
Support for undefined type alone and in union

### LINKED ISSUE #1031: Support UndefinedKeyword
## Sorting

- **I'm submitting a ...**

  - [ ] bug report
  - [x] feature request
  - [ ] support request

- I confirm that I
  - [ ] used the [search](https://github.com/lukeautry/tsoa/search?type=Issues) to make sure that a similar issue hasn't already been submit

## Expected Behavior

Using `@types/node` 12.20.16 should still permit tsoa to generate `routes.ts` when using a custom error as below:

```ts
export interface CustomError extends Error {
  message: string;
  status: number;
}

@Route('GetTest')
export class GetTestController extends Controller {
  @Get('ThrowsError')
  @Response<CustomError>()
  public async getThrowsError(): Promise<TestModel> {
    throw {
      message: 'error thrown',
      status: 400,
    };
  }
}
```

## Current Behavior

```bash
harrinm@C02D11F7MD6T:~/git/tsoa|master> yarn test
yarn run v1.22.4
$ yarn --cwd tests test
$ yarn clean && cross-env NODE_ENV=tsoa_test ts-node --require tsconfig-paths/register ./prepare.ts
$ rimraf dist fixtures/*/routes.ts fixtures/*/customRoutes.ts
↻ Starting Swagger Spec Generation...
There was a problem resolving type of 'Error'.
There was a problem resolving type of 'CustomError'.
Generate routes error.
 GenerateMetadataError: Unknown type: UndefinedKeyword
At: /Users/harrinm/git/tsoa/tests/node_modules/@types/node/globals.d.ts:127:127.
This was caused by 'string | undefined'
    at TypeResolver.resolve (/Users/harrinm/git/tsoa/packages/cli/src/metadataGeneration/typeResolver.ts:309:13)
    at /Users/harrinm/git/tsoa/packages/cli/src/metadataGeneration/typeResolver.ts:62:84
    at Array.map (<anonymous>)
    at TypeResolver.resolve (/Users/harrinm/git/tsoa/packages/cli/src/metadataGeneration/typeResolver.ts:61:41)
    at TypeResolver.propertyFromSignature (/Users/harrinm/git/tsoa/packages/cli/src/metadataGeneration/typeResolver.ts:890:137)
    at /Users/harrinm/git/tsoa/packages/cli/src/metadataGeneration/typeResolver.ts:848:141
    at Array.map (<anonymous>)
    at TypeResolver.getModelProperties (/Users/harrinm/git/tsoa/packages/cli/src/metadataGeneration/typeResolver.ts:848:98)
    at TypeResolver.getModelReference (/Users/harrinm/git/tsoa/packages/cli/src/metadataGeneration/typeResolver.ts:622:29)
    at TypeResolver.getReferenceType (/Users/harrinm/git/tsoa/packages/cli/src/metadataGeneration/typeResolver.ts:563:30)
```

## Possible Solution

Not sure.

I'll at least be providing a PR to demo a failing test for it though.

## Steps to Reproduce

- Update `tests/package.json` to `"@types/node": "^12.20.16"`
- Add `@Response<CustomError>()` to `tests/fixtures/controllers/getController.ts`
- `yarn test`

...or run the PR I'll be putting up shortly to contrib a failing test demoing the problem.

## Context (Environment)

Version of the library: master
Version of NodeJS: 12.21.0

## Detailed Description

In `@types/node` 12.20.16, DefinitelyTyped's [added `| undefined` to a lot of fields](https://github.com/DefinitelyTyped/DefinitelyTyped/commit/d75a4e3dd11692cd81bb06cd6aaa8baadc68d22e), to prepare for the upcoming `exactOptionalPropertyTypes` setting in TypeScript 4.4.

An unfortunate side-effect is that now tsoa has trouble handling a number of built-in types, including `Error`, which is a problem for us when we use it like so: `@Response<MyCustomError>()`

Hi-Fi:
### All Submissions:

**Closing issues**

closes #1031 

Might have some corner cases that are not thought, sharing code as I'm currently a bit puzzled why GET tests are failing (all others are passing).

**Test plan**

Added new element validation added to model, and changed counts to realize that model has one more field than generated spec (because of `undefined` typed value is dropped).

### DIFF
--- a/packages/cli/src/metadataGeneration/typeResolver.ts
+++ b/packages/cli/src/metadataGeneration/typeResolver.ts
@@ -52,6 +52,13 @@ export class TypeResolver {
       return enumType;
     }
 
+    if (this.typeNode.kind === ts.SyntaxKind.UndefinedKeyword) {
+      const undefinedType: Tsoa.UndefinedType = {
+        dataType: 'undefined',
+      };
+      return undefinedType;
+    }
+
     if (ts.isArrayTypeNode(this.typeNode)) {
       const arrayMetaType: Tsoa.ArrayType = {
         dataType: 'array',
@@ -478,6 +485,10 @@ export class TypeResolver {
       return {
         dataType: 'void',
       };
+    } else if (resolution.resolvedType === 'undefined') {
+      return {
+        dataType: 'undefined',
+      };
     } else {
       return assertNever(resolution.resolvedType);
     }
@@ -755,6 +766,11 @@ export class TypeResolver {
         foundMatch: true,
         resolvedType: 'void',
       };
+    } else if (syntaxKind === ts.SyntaxKind.UndefinedKeyword) {
+      return {
+        foundMatch: true,
+        resolvedType: 'undefined',
+      };
     } else {
       return {
         foundMatch: false,
@@ -876,8 +892,7 @@ export class TypeResolver {
 
   private getModelProperties(node: ts.InterfaceDeclaration | ts.ClassDeclaration, overrideToken?: OverrideToken): Tsoa.Property[] {
     const isIgnored = (e: ts.TypeElement | ts.ClassElement) => {
-      const ignore = isExistJSDocTag(e, tag => tag.tagName.text === 'ignore');
-      return ignore;
+      return isExistJSDocTag(e, tag => tag.tagName.text === 'ignore');
     };
 
     // Interface model
@@ -1161,7 +1176,7 @@ export class TypeResolver {
 
 interface ResolvesToPrimitive {
   foundMatch: true;
-  resolvedType: 'number' | 'string' | 'boolean' | 'void';
+  resolvedType: 'number' | 'string' | 'boolean' | 'void' | 'undefined';
 }
 interface DoesNotResolveToPrimitive {
   foundMatch: false;
--- a/packages/cli/src/swagger/specGenerator.ts
+++ b/packages/cli/src/swagger/specGenerator.ts
@@ -40,7 +40,8 @@ export abstract class SpecGenerator {
       guiltyUntilInnocent === 'file' ||
       guiltyUntilInnocent === 'number' ||
       guiltyUntilInnocent === 'object' ||
-      guiltyUntilInnocent === 'string'
+      guiltyUntilInnocent === 'string' ||
+      guiltyUntilInnocent === 'undefined'
     ) {
       return guiltyUntilInnocent;
     } else {
@@ -49,7 +50,7 @@ export abstract class SpecGenerator {
   }
 
   protected getSwaggerType(type: Tsoa.Type): Swagger.Schema | Swagger.BaseSchema {
-    if (type.dataType === 'void') {
+    if (type.dataType === 'void' || type.dataType === 'undefined') {
       return this.getSwaggerTypeForVoid(type.dataType);
     } else if (type.dataType === 'refEnum' || type.dataType === 'refObject' || type.dataType === 'refAlias') {
       return this.getSwaggerTypeForReferenceType(type);
@@ -96,7 +97,7 @@ export abstract class SpecGenerator {
 
     const additionalProperties = objectLiteral.additionalProperties && this.getSwaggerType(objectLiteral.additionalProperties);
 
-    const required = objectLiteral.properties.filter(prop => prop.required).map(prop => prop.name);
+    const required = objectLiteral.properties.filter(prop => prop.required && !this.hasUndefined(prop)).map(prop => prop.name);
 
     // An empty list required: [] is not valid.
     // If all properties are optional, do not specify the required keyword.
@@ -114,7 +115,7 @@ export abstract class SpecGenerator {
     };
   }
 
-  protected getSwaggerTypeForVoid(_dataType: 'void'): Swagger.BaseSchema {
+  protected getSwaggerTypeForVoid(_dataType: 'void' | 'undefined'): Swagger.BaseSchema {
     // Described here: https://swagger.io/docs/specification/describing-responses/#empty
     const voidSchema = {
       // isn't allowed to have additionalProperties at all (meaning not a boolean or object)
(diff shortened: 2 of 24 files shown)


WoH:
Following up on the textual description, here are some pointers to code:
1) `UndefinedType` should probably be very similar to the `VoidType`, which you can find [here](https://github.com/lukeautry/tsoa/blob/master/packages/runtime/src/metadataGeneration/tsoa.ts#L126)
2. The handling (extracting it from a union and marking the prop as optional) should happen in the Spec Generation, see [here](https://github.com/lukeautry/tsoa/blob/master/packages/cli/src/swagger/specGenerator3.ts#L589)
3. `undefined` as a return should be handled as isVoidType (cli/src/utils/isVoidType) 
4. `undefined` as a Parameter should mark is as optional, but set it to `undedined` during validation (we can defer that for now and get the property and response part working)

Hi-Fi:
> Following up on the textual description, here are some pointers to code:
> 
>     1. `UndefinedType` should probably be very similar to the `VoidType`, which you can find [here](https://github.com/lukeautry/tsoa/blob/master/packages/runtime/src/metadataGeneration/tsoa.ts#L126)
> 
> 
>     1. The handling (extracting it from a union and marking the prop as optional) should happen in the Spec Generation, see [here](https://github.com/lukeautry/tsoa/blob/master/packages/cli/src/swagger/specGenerator3.ts#L589)
> 
>     2. `undefined` as a return should be handled as isVoidType (cli/src/utils/isVoidType)
> 
>     3. `undefined` as a Parameter should mark is as optional, but set it to `undedined` during validation (we can defer that for now and get the property and response part working)

Thank you, I started from master again with a kind of approach that you described, but those pointers given would help me a lot. Thank you for those.

WoH:
In the SpecGenerators, there's a buildSchema/buildDefinitions method, you want to take a look there. Also the buildParameter() in the same class

Hi-Fi:
@WoH  Now just 2 test fails :) Does the approach look better now for other parts?

```
  1271 passing (42s)
  2 failing

  1) Express Server
       Parameter data
         can get request with generic type:
     Uncaught TypeError: Cannot read properties of undefined (reading 'error')
      at Test.<anonymous> (integration/dynamic-controllers-express-server.spec.ts:1131:31)
      at Test.assert (/Users/juhosaarinen/git/tsoa/node_modules/supertest/lib/test.js:181:6)
      at Server.localAssert (/Users/juhosaarinen/git/tsoa/node_modules/supertest/lib/test.js:131:12)
      at Object.onceWrapper (node:events:509:28)
      at Server.emit (node:events:390:28)
      at emitCloseNT (node:net:1681:8)
      at processTicksAndRejections (node:internal/process/task_queues:82:21)

  2) GET route generation
       should generate example from example decorator:
     Error: No examples.
      at Context.<anonymous> (unit/swagger/pathGeneration/getRoutes.spec.ts:232:13)
      at processImmediate (node:internal/timers:464:21)
```

WoH:
This is very close to what I had in mind, with a small difference, which is why there are some issues with type handling we should tweak.
In the type resolver, if we try to filter the `undefined` types, we lose information we will later need (for validation/coercion).
Generally speaking, tsoa is trying to be close to TS validation. Therefore, I'd want to cover these things:

1) If TS says there's an optional property (i.e. `propA?: string`) and the client doesn't send anything, the prop should *NOT* be undefined (instead not have the prop).

2) If TS says there's a property that can be `undefined` (`propA: string | undefined`) and the client doesn't send anything, we should make sure we actually hand off `propA: undefined` to the user code.

But that means we need to keep the info around and only have the specGenerator "rewrite" the unions with `undefined` to an optional property.

That would also mean we should be able to revert the `Partial<TestModel>` work around.

Hi-Fi:
> This is very close to what I had in mind, with a small difference, which is why there are some issues with type handling we should tweak. In the type resolver, if we try to filter the `undefined` types, we lose information we will later need (for validation/coercion). Generally speaking, tsoa is trying to be close to TS validation. Therefore, I'd want to cover these things:
> 
> 1. If TS says there's an optional property (i.e. `propA?: string`) and the client doesn't send anything, the prop should _NOT_ be `undefined` (after validation).
> 2. If TS says there's a property that can be `undefined` (`propA: string | undefined`) and the client doesn't send anything, we should make sure we actually hand off `propA: undefined` to the user code.
> 
> But that means we need to keep the info around and only have the specGenerator "rewrite" the unions with `undefined` to an optional property.
> 
> That would also mean we should be able to revert the `Partial<TestModel>` work around.

So models in test should include `undefinedValue` and `undefineableUnionPrimitiveType`, even though those are not something that user would need to give (if spec says that those are optional or not even in spec)? And those have to be in spec, as (at least when model is used to define input values) those are required by validation? 

I'll have to tinker around the code then a bit more, I mostly focused on produced swagger.yaml and compiling/passing tests.

Hi-Fi:
Related to `Partial<TestModel>`: I'll add test models those 2 new values with undefined, if I don't fix validation, too. As now with fixes those validation tests fail again with HTTP code 400.

Hi-Fi:
@WoH  I left `Partial<TestModel>` change in, as that would test same time that values that are marked mandatory in model are handled correctly in code and not complaining that those are missing.

I think that that explicit returning of the value from validation when value is not set by user is not working now, or then some other part of code might still be problematic and dropping out undefined value.

Hi-Fi:
@WoH  any updates/fixes needed?

WoH:
> This is very close to what I had in mind, with a small difference, which is why there are some issues with type handling we should tweak. In the type resolver, if we try to filter the `undefined` types, we lose information we will later need (for validation/coercion). Generally speaking, tsoa is trying to be close to TS validation. Therefore, I'd want to cover these things:
> 
> 1. If TS says there's an optional property (i.e. `propA?: string`) and the client doesn't send anything, the prop should _NOT_ be `undefined` (after validation).
> 2. If TS says there's a property that can be `undefined` (`propA: string | undefined`) and the client doesn't send anything, we should make sure we actually hand off `propA: undefined` to the user code.
> 
> But that means we need to keep the info around and only have the specGenerator "rewrite" the unions with `undefined` to an optional property.
> 
> That would also mean we should be able to revert the `Partial<TestModel>` work around.

Integration specs for 1. and 2.

Hi-Fi:
@WoH  Now I think that the value handling works as intended, and it's also tested in tests.

Hi-Fi:
@WoH Still something to do? Tests added to check that mentioned setting of undefined, and functionality changed accordingly.

WoH:
LGTM, I had to find time to take a look at the changes.

Issue #968, opened 2021-04-27
Support for vendor media types


pittenga:
OpenAPI provides support for [vendor media types](https://swagger.io/docs/specification/media-types/). Is there a way to use the middleware to define that request body media type in the swagger output? Or maybe use a tag for it?

## Sorting

- **I'm submitting a ...**

  - [ ] bug report
  - [x] feature request
  - [x] support request

- I confirm that I
  - [x] used the [search](https://github.com/lukeautry/tsoa/search?type=Issues) to make sure that a similar issue hasn't already been submit

## Expected Behavior
Show the content type in the request body of a vendor specific media type.

```js
"requestBody": {
      "content": {
            "application/vnd.mycompany-v1.0+json": {
```

## Current Behavior

```js
"requestBody": {
      "content": {
            "application/json": {
```

## Possible Solution

```ts
private buildRequestBody(controllerName: string, method: Tsoa.Method, parameter: Tsoa.Parameter): Swagger.RequestBody {
    const mediaType = this.buildMediaType(controllerName, method, parameter); 
    const mediaTypeName: string = this.buildMediaTypeNameFromMiddleware(); <-- add something like this

    const requestBody: Swagger.RequestBody = {
      description: parameter.description,
      required: parameter.required,
      content: {
        mediaTypeName: mediaType,
      },
    };

    return requestBody;
  }
```
 Or if we can't tap into the middleware like that, maybe we can add a tag for it?
```ts
@Route('Controller')
@Response<ErrorResponseModel>('401', 'Unauthorized')
export class HiddenMethodController extends Controller {
  @Get('protectedGetMethod')
  public async protectedGetMethod(): Promise<TestModel> {
    return Promise.resolve(new ModelService().getModel());
  }
  @MediaType("application/vnd.mycompany-v1.0+json");
  @Post('protectedPostMethod')
  public async protectedPostMethod(): Promise<TestModel> {
    return Promise.resolve(new ModelService().getModel());
  }
}
```
## Steps to Reproduce
  
In the middleware (I'm using express), add something like this:
```ts
app.use(bodyParser.json({ type: "application/vnd.mycompany-v1.0+json" }));
```
And that comes out in the swagger output

## Context (Environment)

Version of the library: 3.6.1
Version of NodeJS: 12.18.2

- Confirm you were using yarn not npm: [ ]

mrl5:
if I understand correctly @pittenga description is about **request** body media type.

[...]

Issue #44, opened 2017-01-16
Send file / return raw binary "application/octect-stream" or "image/jpeg" ...


AntoineABI:
Hello,

I am struggling on something I wan't to do and it's seems not possible to do it with the tsoa framework.
**Returning raw data or file** is not allowed by using a standard controller. Obviously, I can define my own express route and custom controller but I can't add the **spec** to **swagger.json automatically** because when I fill the spec field in **tsoa.json** it is overriding every paths !

Proposal:
 - **merge** tsoa.json "swagger"."spec" **recursively** instead of overriding at the first level.
 - allow **more customization of controller** and extend the router/promiseHandler.

My solution: 
 - Defining custom route with express
 - Custom tsoa.spec.json merged with my own script
(not yet done wanna see if you guys have another way around or if I missed something?)

Regards.

PS: It is whats looks like my tsoa.json if it can help
```
{
    "swagger": {
        "outputDirectory": "./dist",
        "entryFile": "./src/server.ts",
        "host": "localhost:3000",
        "basePath": "/v1",
        "spec": {
            "paths": {
                "/Data/GetUserProfilePicture/{userId}": {
                    "get": {
                        "operationId": "getUserProfilePicture",
                        "produces": [
                            "image/jpeg"
                        ],
                        "responses": {
                            "200": {
                                "description": "",
                                "examples": {},
                                "schema": {
                                    "type": "binary"
                                }
                            }
                        },
                        "description": "Get an user profile picture",
                        "parameters": [
                            {
                                "description": "",
                                "in": "path",
                                "name": "userId",
                                "required": true,
                                "type": "integer",
                                "format": "int64"
                            }
                        ]
                    }
                }
            }
        }
    },
    "routes": {
        "basePath": "/v1",
        "entryFile": "./src/server.ts",
        "routesDir": "./src",
        "middleware": "express"
    }
}
```

lukeautry:
@AntoineABI I think the solution here is going to be to improve the merge logic between custom/"provided" spec and the generated spec; ideally you should be able to override the generated spec when you need to.

If you're inclined to take a stab at a solution, the merge is happening here: https://github.com/lukeautry/tsoa/blob/master/src/swagger/specGenerator.ts#L47

We'd just need to do something custom there instead of using Object.assign, or it may be as easy as swapping the source/destination.

AntoineABI:
That's whats I though ! We could use something like https://www.npmjs.com/package/merge which do exactly what I was expected.

[...]

PR #1129, opened 2021-11-08
feat(spec): allow defining custom media type of responses + fix(koa): dont overwrite content-type of response

### LINKED ISSUE #1126: support: how to set correct content type in generated oas3
Endpoint returning `text/plain` has a corresponding tsoa generated oas3 `content` set as `application/json`.

Is there any way to have a accurate `content` in generated `swagger.json`?

## Sorting

- **I'm submitting a ...**

  - [ ] bug report
  - [ ] feature request
  - [x] support request

- I confirm that I

## Expected Behavior

```diff
                                        "200": {
                                                "description": "Ok",
                                                "content": {
-                                                       "application/json": {
+                                                       "text/plain": {
                                                                "schema": {
-                                                                       "type": "number",
-                                                                       "enum": [
-                                                                               null
-                                                                       ],
-                                                                       "nullable": true
+                                                                       "type": "string"
                                                                }
                                                        }
                                                }
```

## Current Behavior

```diff
                        }
+               },
+               "/.well-known/security.txt": {
+                       "get": {
+                               "operationId": "GetSecurityTxt",
+                               "responses": {
+                                       "200": {
+                                               "description": "Ok",
+                                               "content": {
+                                                       "application/json": {
+                                                               "schema": {
+                                                                       "type": "number",
+                                                                       "enum": [
+                                                                               null
+                                                                       ],
+                                                                       "nullable": true
+                                                               }
+                                                       }
+                                               }
+                                       }
+                               },
+                               "security": [],
+                               "parameters": []
+                       }
                }
```

## Possible Solution

No idea

## Steps to Reproduce

```diff
--- /dev/null
+++ b/src/well-known/wellKnownController.ts
@@ -0,0 +1,16 @@
+import { Response as ExResponse, Request as ExRequest } from 'express';
+import { Controller, Get, Request, Route } from 'tsoa';
+import { WellKnownService } from './wellKnownService';
+
+@Route('/.well-known')
+export class WellKnownController extends Controller {
+    @Get('/security.txt')
+    async getSecurityTxt(@Request() request: ExRequest): Promise<null> {
+        const securityTxt = await new WellKnownService().getSecurityTxt();
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        const res = (<any>request).res as ExResponse;
+        res.type('.txt');
+        res.send(securityTxt);
+        return null;
+    }
+}
```

## Context (Environment)

Version of the library: `3.14.0`
Version of NodeJS: `v14.17.6`

- Confirm you were using yarn not npm:
  - [x] I confirm

## Detailed Description

**use case 1** As a security researcher I must know how to securely disclose security vulnerabilities

**use case 2** as a developer that's creating JSON REST API I wan't to serve `security.txt` as specified in https://securitytxt.org/ and mentioned in https://cheatsheetseries.owasp.org/cheatsheets/Vulnerability_Disclosure_Cheat_Sheet.html#finding-contact-details

specification of `security.txt` says this:
> The security.txt file should have an Internet Media Type of text/plain and must be served over HTTPS.

mrl5:
problem description: currently `application/json` is hardcoded into generated spec
    
proposed solution: allow specifying custom media type in generated spec
on both controller level by:
  - using new `@Produces` decorator

and on method level by:
  - using new `@Produces` decorator
  - introducing new optional arg for `@SuccessResponse` decorator
  - introducing new optional arg for `@Response` decorator
  - setting `Content-Type` header in `@Res` decorator

impact:
  - cli: controller generator
  - cli: method generator
  - cli: parameter generator
  - cli: spec2 generator
  - cli: spec3 generator
  - runtime: response decorators

    
Closes #1126
Related #968

### All Submissions:

**Closing issues**

- [x] Put `closes #XXXX` (where XXXX is the issue number) in your comment to auto-close the issue that your PR fixes.

1. Someone could think that defining custom media type will also override `content-type` of actual response (which is not a case). However I don't think it's a problem -> if you check the issues from the past you will see that writing custom methods was encouraged - for example #44 #224
2. OAS2 is not affected by this changes
3. ~~I wasn't able to introduce new arg for setting custom media type for `@Res` decorator~~

**Test plan**

1. I've tried to show that it is possible to set `@Produces` on controller level
2. I've tried to show that even when `@Produces` is set on controller level it is possible to set it also for specific method
3. I've tried to show that OAS3 media type of responses is possible to be changed by passing new optional arg in `@Response` and `@SuccessResponse` decorators
4. I've tried to show that OAS3 media type of responses is possible to be changed by defining headers for `@Res` decorator

### DIFF
--- a/packages/cli/src/swagger/specGenerator2.ts
+++ b/packages/cli/src/swagger/specGenerator2.ts
@@ -2,7 +2,7 @@ import { ExtendedSpecConfig } from '../cli';
 import { Tsoa, assertNever, Swagger } from '@tsoa/runtime';
 import { isVoidType } from '../utils/isVoidType';
 import { convertColonPathParams, normalisePath } from './../utils/pathUtils';
-import { getValue } from './../utils/swaggerUtils';
+import { DEFAULT_REQUEST_MEDIA_TYPE, DEFAULT_RESPONSE_MEDIA_TYPE, getValue } from './../utils/swaggerUtils';
 import { SpecGenerator } from './specGenerator';
 import { UnspecifiedObject } from '../utils/unspecifiedObject';
 
@@ -14,13 +14,13 @@ export class SpecGenerator2 extends SpecGenerator {
   public GetSpec() {
     let spec: Swagger.Spec2 = {
       basePath: normalisePath(this.config.basePath as string, '/', undefined, false),
-      consumes: ['application/json'],
+      consumes: [DEFAULT_REQUEST_MEDIA_TYPE],
       definitions: this.buildDefinitions(),
       info: {
         title: '',
       },
       paths: this.buildPaths(),
-      produces: ['application/json'],
+      produces: [DEFAULT_RESPONSE_MEDIA_TYPE],
       swagger: '2.0',
     };
 
@@ -146,15 +146,15 @@ export class SpecGenerator2 extends SpecGenerator {
           let path = normalisePath(`${normalisedControllerPath}${normalisedMethodPath}`, '/', '', false);
           path = convertColonPathParams(path);
           paths[path] = paths[path] || {};
-          this.buildMethod(controller.name, method, paths[path]);
+          this.buildMethod(controller.name, method, paths[path], controller.produces);
         });
     });
 
     return paths;
   }
 
-  private buildMethod(controllerName: string, method: Tsoa.Method, pathObject: any) {
-    const pathMethod: Swagger.Operation = (pathObject[method.method] = this.buildOperation(controllerName, method));
+  private buildMethod(controllerName: string, method: Tsoa.Method, pathObject: any, defaultProduces?: string) {
+    const pathMethod: Swagger.Operation = (pathObject[method.method] = this.buildOperation(controllerName, method, defaultProduces));
     pathMethod.description = method.description;
     pathMethod.summary = method.summary;
     pathMethod.tags = method.tags;
@@ -187,21 +187,23 @@ export class SpecGenerator2 extends SpecGenerator {
     method.extensions.forEach(ext => (pathMethod[ext.key] = ext.value));
   }
 
-  protected buildOperation(controllerName: string, method: Tsoa.Method): Swagger.Operation {
+  protected buildOperation(controllerName: string, method: Tsoa.Method, defaultProduces?: string): Swagger.Operation {
     const swaggerResponses: { [name: string]: Swagger.Response } = {};
 
+    let produces: Array<string | undefined> = [];
     method.responses.forEach((res: Tsoa.Response) => {
       swaggerResponses[res.name] = {
         description: res.description,
       };
       if (res.schema && !isVoidType(res.schema)) {
+        produces.push(res.produces);
         swaggerResponses[res.name].schema = this.getSwaggerType(res.schema) as Swagger.Schema;
       }
       if (res.examples && res.examples[0]) {
         if ((res.exampleLabels?.filter(e => e).length || 0) > 0) {
           console.warn('Example labels are not supported in OpenAPI 2');
         }
-        swaggerResponses[res.name].examples = { 'application/json': res.examples[0] } as Swagger.Example;
+        swaggerResponses[res.name].examples = { [DEFAULT_RESPONSE_MEDIA_TYPE]: res.examples[0] } as Swagger.Example;
       }
 
       if (res.headers) {
@@ -220,9 +222,14 @@ export class SpecGenerator2 extends SpecGenerator {
       }
     });
 
+    produces = Array.from(new Set(produces.filter(p => p)));
+    if (produces.length === 0) {
+      produces = [defaultProduces || DEFAULT_RESPONSE_MEDIA_TYPE];
+    }
+
     const operation: Swagger.Operation = {
       operationId: this.getOperationId(method.name),
-      produces: ['application/json'],
+      produces: produces as string[],
       responses: swaggerResponses,
     };
 
--- a/packages/cli/src/swagger/specGenerator2.ts
+++ b/packages/cli/src/swagger/specGenerator2.ts
@@ -222,7 +222,7 @@ export class SpecGenerator2 extends SpecGenerator {
       }
     });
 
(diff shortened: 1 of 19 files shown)


packages/runtime/src/decorators/response.ts
```diff
@@ -23,6 +24,18 @@ export function Response<ExampleType, HeaderType extends IsValidHeader<HeaderTyp
  * The type of the responder function should be annotated `TsoaResponse<Status, Data, Headers>` in order to support OpenAPI documentation.
  */
 export function Res(): Function {
+  // todo: allow setting media type
```
WoH, line 27: I think it'd be better to check the Header type (3rd arg to `TsoaResponse`) over adding an arg to `@Res`.

dderevjanik:
@mrl5 very nice PR! 
I'm just curious, what's the reason behind `@MediaType` decorator? I mean, most frameworks (jersey, spring, asp.net, nestjs, etc.) are using `@Produces` or `@ProducesSomething` wording  - so you can distinguish between `@Consumes`. 

In case that TSOA will support `@Consumes` in future, IMHO it could be confusing to have something like this in controller:

```ts
 @Get('process')
 @Consumes('image/png')  // Consumes MediaType
 @MediaType('text/plain') // Produces MediaType
  public async processImage(@Request() request: ExRequest): Promise<string> {
    // code
    return '';
  }
```

mrl5:
thanks for code review @dderevjanik - I agree that `@Produces` and `@Consumes` are more accurate. I will change that

> I'm just curious, what's the reason behind @MediaType decorator?
I was inspired by the code snippets from #968

@WoH is it ok to `git commit --amend` not to produce commit flood on every improvement or you prefer to keep it separated?

WoH:
Amending and rebasing is fine

mrl5:
1. pushed improvements suggested by @WoH and @dderevjanik
2. updated PR description
3. added possibility to define `@Responses`` on controller level

tests/fixtures/controllers/mediaTypeController.ts
```diff
@@ -0,0 +1,45 @@
...
+    @Body() model: UserRequestModel,
+    @Res() conflictRes: TsoaResponse<409, { message: string }, { 'content-type': 'application/vnd.mycompany.myapp.v2+json' }>,
+  ): Promise<UserResponseModel> {
+    if (model.name === 'bar') {
+      return conflictRes?.(409, { message: 'Conflict' });
+    }
+
+    const body = { id: model.name.length, name: model.name };
+    this.setStatus(202);
+    this.setHeader('content-type', 'application/vnd.mycompany.myapp.v2+json'); // express and hapi returns it, koa ignores
```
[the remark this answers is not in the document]
mrl5: subject of #1134

mrl5:
pushed additional commit that:
1. fixes issue #1134 
2. introduces integration tests checking `content-type`

packages/cli/src/swagger/specGenerator2.ts
```diff
@@ -14,13 +14,13 @@ export class SpecGenerator2 extends SpecGenerator {
...
       basePath: normalisePath(this.config.basePath as string, '/', undefined, false),
-      consumes: ['application/json'],
+      consumes: [DEFAULT_REQUEST_MEDIA_TYPE],
       definitions: this.buildDefinitions(),
       info: {
         title: '',
       },
       paths: this.buildPaths(),
-      produces: ['application/json'],
+      produces: [DEFAULT_RESPONSE_MEDIA_TYPE], // todo: Tsoa.Response.produces
```
WoH, line 23: > A list of MIME types the APIs can produce.
This is global to all APIs but can be overridden on specific API calls.
Value MUST be as described under Mime Types.

Maybe I'm interpreting too much, but we should be able to leave this as is, remove the todo and override on path.method.produces?

packages/cli/src/swagger/specGenerator2.ts
```diff
@@ -222,7 +222,7 @@ export class SpecGenerator2 extends SpecGenerator {
 
     const operation: Swagger.Operation = {
       operationId: this.getOperationId(method.name),
-      produces: ['application/json'],
+      produces: [DEFAULT_RESPONSE_MEDIA_TYPE], // todo: Tsoa.Response.produces
```
WoH, line 225: We should fill in the `produces` here, right?

mrl5: > We should fill in the `produces` here, right?

as opposed to `packages/cli/src/swagger/specGenerator3.ts` where there are `buildPaths()` and `buildOperation()` this would require major refactor.

unfortunately I don't want to do it in this PR

mrl5: > We should fill in the `produces` here, right?

indeed, will do

mrl5:
1. amended with changes from code review
2. updated PR description

thanks @WoH for detailed CR :) I hope we're good to go now

packages/cli/src/swagger/specGenerator2.ts
```diff
@@ -220,9 +222,14 @@ export class SpecGenerator2 extends SpecGenerator {
       }
     });
 
+    produces = Array.from(new Set(produces.filter(p => p)));
```
WoH, line 225: `filter(p => p)` ?

WoH (approved):

WoH:
Thanks for this PR and your patience during the review, much appreciated. Great Work!

alberto-rubio:
Please, @mrl5  and @WoH, is this working in version 4.0.0-alpha.2? I'm adding the decorator:
`@Produces('text/event-stream')`
at method level, but it continues producing "application/json" in the generated openapi yaml definition.

Should I do any additional thing to see "text/event-stream" in the generate openapi?

Thanks and best regards!!

mrl5:
hi @alberto-rubio - I think that this changes are already included in this version. It's hard to help you w/o seeing any example code that is producing unexpected results

if you need example you can study this file that is used in unit tests to ensure that feature introduced in this PR works: https://github.com/lukeautry/tsoa/blob/6a289a7ea0576835cf0803db9905a8f39cab0107/tests/fixtures/controllers/mediaTypeController.ts#L27-L28

PR #1140, opened 2021-11-20
feat(spec): allow defining multiple media types per response, custom media type of requestBody 

### LINKED ISSUE #968: Support for vendor media types

mrl5:
Changes:

1) extend commit 323f6776b0
From swagger media types specification: "Web service operations can
accept and return data in different formats, the most common being JSON,
XML and images. You specify the media type in request and response
definitions."

https://swagger.io/docs/specification/media-types/

Related #1126

2) problem description: currently `application/json` is hardcoded into
generated spec

proposed solution: allow specifying custom media type in generated spec
on method level by:
  - using new `@Consumes` decorator

impact:
  - cli: method generator
  - cli: spec2 generator
  - cli: spec3 generator
  - runtime: response decorators

Closes #968

### All Submissions:

**Closing issues**

1. described in detail in this comment: https://github.com/lukeautry/tsoa/pull/1140#issuecomment-974628592
 2. it is possible to define only one `@Consumes` per method

**Test plan**

unit & integration tests

### DIFF
--- a/packages/cli/src/metadataGeneration/methodGenerator.ts
+++ b/packages/cli/src/metadataGeneration/methodGenerator.ts
@@ -1,7 +1,7 @@
 import * as ts from 'typescript';
 import * as path from 'path';
 import { isVoidType } from '../utils/isVoidType';
-import { getDecorators, getDecoratorValues, getPath, getSecurites } from './../utils/decoratorUtils';
+import { getDecorators, getDecoratorValues, getPath, getProduces, getSecurites } from './../utils/decoratorUtils';
 import { getJSDocComment, getJSDocDescription, isExistJSDocTag } from './../utils/jsDocUtils';
 import { getExtensions } from './extension';
 import { GenerateMetadataError } from './exceptions';
@@ -15,7 +15,7 @@ import { getHeaderType } from '../utils/headerTypeHelpers';
 export class MethodGenerator {
   private method: 'options' | 'get' | 'post' | 'put' | 'patch' | 'delete' | 'head';
   private path: string;
-  private produces?: string;
+  private produces?: string[];
 
   constructor(
     private readonly node: ts.MethodDeclaration,
@@ -139,19 +139,9 @@ export class MethodGenerator {
     this.produces = this.getProduces();
   }
 
-  private getProduces(): string | undefined {
-    const producesDecorators = this.getDecoratorsByIdentifier(this.node, 'Produces');
-
-    if (!producesDecorators || !producesDecorators.length) {
-      return;
-    }
-    if (producesDecorators.length > 1) {
-      throw new GenerateMetadataError(`Only one Produces decorator in '${this.getCurrentLocation()}' method, Found: ${producesDecorators.map(d => d.text).join(', ')}`);
-    }
-
-    const [decorator] = producesDecorators;
-    const [produces] = getDecoratorValues(decorator, this.current.typeChecker);
-    return produces;
+  private getProduces(): string[] | undefined {
+    const produces = getProduces(this.node, this.current.typeChecker);
+    return produces.length ? produces : undefined;
   }
 
   private getMethodResponses(): Tsoa.Response[] {
@@ -169,7 +159,7 @@ export class MethodGenerator {
         description: description || '',
         examples: example === undefined ? undefined : [example],
         name: name || '200',
-        produces,
+        produces: this.getProducesAdapter(produces),
         schema: expression.typeArguments && expression.typeArguments.length > 0 ? new TypeResolver(expression.typeArguments[0], this.current).resolve() : undefined,
         headers: getHeaderType(expression.typeArguments, 1, this.current),
       } as Tsoa.Response;
@@ -206,7 +196,7 @@ export class MethodGenerator {
         description: description || '',
         examples,
         name: name || '200',
-        produces,
+        produces: this.getProducesAdapter(produces),
         schema: type,
         headers,
       },
@@ -315,5 +305,14 @@ export class MethodGenerator {
 
   private getDecoratorsByIdentifier(node: ts.Node, id: string) {
     return getDecorators(node, identifier => identifier.text === id);
+  }
+
+  private getProducesAdapter(produces?: string[] | string): string[] | undefined {
+    if (Array.isArray(produces)) {
+      return produces;
+    } else if (typeof produces === 'string') {
+      return [produces];
+    }
+    return;
   }
 }
--- a/packages/cli/src/metadataGeneration/methodGenerator.ts
+++ b/packages/cli/src/metadataGeneration/methodGenerator.ts
@@ -16,6 +16,7 @@ export class MethodGenerator {
   private method: 'options' | 'get' | 'post' | 'put' | 'patch' | 'delete' | 'head';
   private path: string;
   private produces?: string[];
+  private consumes?: string;
 
   constructor(
     private readonly node: ts.MethodDeclaration,
@@ -64,6 +65,7 @@ export class MethodGenerator {
       parameters,
       path: this.path,
       produces: this.produces,
+      consumes: this.consumes,
       responses,
       successStatus: successStatus,
       security: this.getSecurity(),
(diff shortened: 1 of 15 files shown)


mrl5:
created DRAFT PR because it's easier to discuss certain things when you see the code and tests as an example

packages/cli/src/swagger/specGenerator3.ts
```diff
@@ -298,12 +298,17 @@ export class SpecGenerator3 extends SpecGenerator {
...
       if (res.schema && !isVoidType(res.schema)) {
-        const produces = res.produces || defaultProduces || DEFAULT_RESPONSE_MEDIA_TYPE;
-        swaggerResponses[res.name].content = {
-          [produces]: {
-            schema: this.getSwaggerType(res.schema),
-          } as Swagger.Schema3,
-        };
+        swaggerResponses[res.name].content = {};
+        const produces: string[] = res.produces || defaultProduces || [DEFAULT_RESPONSE_MEDIA_TYPE];
+        for (const p of produces) {
```
mrl5, line 303: normally I would do it simpler by using `.map()` but this is the only working way where typescript compiler didnt complain

packages/cli/src/metadataGeneration/methodGenerator.ts
```diff
@@ -139,19 +139,9 @@ export class MethodGenerator {
     this.produces = this.getProduces();
   }
 
-  private getProduces(): string | undefined {
-    const producesDecorators = this.getDecoratorsByIdentifier(this.node, 'Produces');
-
-    if (!producesDecorators || !producesDecorators.length) {
-      return;
-    }
-    if (producesDecorators.length > 1) {
```
mrl5, line 148: ok @WoH so here is the deal: currently we allow only one `@Produces`. Also it's possible to put only one media type to `SuccessResponse` and `Response` - however according to swagger specification docs it's possible to specify multiple media types for given response.

mrl5:
so this is a working Draft but I'm not sure if it's worth introducing those changes:
1. on one hand swagger allows defining multiple media types per response (I've described [in this comment](https://github.com/lukeautry/tsoa/pull/1140#pullrequestreview-811818057) in more detail)
2. on the other hand if you have multiple produces and multiple response interfaces [like here](https://github.com/lukeautry/tsoa/blob/cafd5a5f8552cbe3465192df1f87be0b1a391e17/tests/fixtures/controllers/requestExpressController.ts#L52-L57) then you get specification like this with this ugly `anyOf` schema :/:

```
{
  "get": {
    "operationId": "GetMultiProduces",
    "responses": {
      "200": {
        "description": "Ok",
        "content": {
          "application/vnd.mycompany.myapp+json": {
            "schema": {
              "anyOf": [
                {
                  "$ref": "#/components/schemas/UserResponseModel"
                },
                {
                  "$ref": "#/components/schemas/UserResponseV2Model"
                },
                {
                  "$ref": "#/components/schemas/UserResponseV3Model"
                },
                {
                  "$ref": "#/components/schemas/UserResponseV4Model"
                }
              ]
            }
          },
          "application/vnd.mycompany.myapp.v2+json": {
            "schema": {
              "anyOf": [
                {
                  "$ref": "#/components/schemas/UserResponseModel"
                },
                {
                  "$ref": "#/components/schemas/UserResponseV2Model"
                },
                {
                  "$ref": "#/components/schemas/UserResponseV3Model"
                },
                {
                  "$ref": "#/components/schemas/UserResponseV4Model"
                }
              ]
            }
          },
          "application/vnd.mycompany.myapp.v3+json": {
            "schema": {
              "anyOf": [
                {
                  "$ref": "#/components/schemas/UserResponseModel"
                },
                {
                  "$ref": "#/components/schemas/UserResponseV2Model"
                },
                {
                  "$ref": "#/components/schemas/UserResponseV3Model"
                },
                {
                  "$ref": "#/components/schemas/UserResponseV4Model"
                }
              ]
            }
          },
          "application/vnd.mycompany.myapp.v4+json": {
            "schema": {
              "anyOf": [
                {
                  "$ref": "#/components/schemas/UserResponseModel"
                },
                {
                  "$ref": "#/components/schemas/UserResponseV2Model"
                },
                {
                  "$ref": "#/components/schemas/UserResponseV3Model"
                },
                {
                  "$ref": "#/components/schemas/UserResponseV4Model"
                }
              ]
            }
          }
        }
      }
    }
  }
}
```

that's why I'm not sure if it's worth introducing this changes because imo there is no easy way to define only one schema per media type

packages/cli/src/swagger/specGenerator3.ts
```diff
@@ -298,12 +298,17 @@ export class SpecGenerator3 extends SpecGenerator {
...
+          const { content } = swaggerResponses[res.name];
+          swaggerResponses[res.name].content = {
+            ...content,
```
mrl5, line 306: normally I would do it simpler but this is the only working way where typescript compiler didn't complain

mrl5:
there is also this brilliant blog post about API versioning: https://apisyouwonthate.com/blog/api-versioning-has-no-right-way - it actually mentions ia.:
    - mime-based resource versioning (example from this PR [here](https://github.com/lukeautry/tsoa/blob/cafd5a5f8552cbe3465192df1f87be0b1a391e17/tests/fixtures/controllers/requestExpressController.ts#L58-L82))
    - Deprecating Fields with OpenAPI v3.0 - which we already support: https://tsoa-community.github.io/docs/decorators.html#deprecated

so @WoH maybe we should not support multiple produces (which is the subject of this DRAFT) and instead encourage usage of `@Deprecated` decorator?

WoH:
I don't think we should encourage either approach. However, multiple produces make sense imo, not just for versioning.

mrl5:
pushed additonal commit that closes #968 by introducing `@Consumes` decorator - however only one `@Consumes` per method is allowed

packages/cli/src/metadataGeneration/methodGenerator.ts
```diff
@@ -137,21 +139,27 @@ export class MethodGenerator {
...
+    const consumesDecorators = this.getDecoratorsByIdentifier(this.node, 'Consumes');
 
-    if (!producesDecorators || !producesDecorators.length) {
+    if (!consumesDecorators || !consumesDecorators.length) {
       return;
     }
-    if (producesDecorators.length > 1) {
-      throw new GenerateMetadataError(`Only one Produces decorator in '${this.getCurrentLocation()}' method, Found: ${producesDecorators.map(d => d.text).join(', ')}`);
+    if (consumesDecorators.length > 1) {
+      throw new GenerateMetadataError(`Only one Consumes decorator in '${this.getCurrentLocation()}' method, Found: ${consumesDecorators.map(d => d.text).join(', ')}`);
```
mrl5, line 157: we allow only one `@Body` per method so I guess it's ok to also allow only one `@Consumes` per method

WoH:
This should be good to go, right? Because it pretty much LGTM

mrl5:
> This should be good to go, right? Because it pretty much LGTM

yup, from my point of view this PR is final and ready to be merged

PR #1144, opened 2021-11-27
feat: Add support for js config file (closes #745)

### LINKED ISSUE #745: Support configuration using JS
Minor improvement, but could you support a tsoa.config.js instead of a tsoa.json so we don't have to duplicate path alias from tsconfig? This would just be a simple function that returns the same json object but from a js script so we could import tsconfig and use data from that to prevent having to duplicate code.

Thanks

### LINKED ISSUE #754: Is there a way to directly specify or return success but leave the method type as it should be ?
## Sorting

- **I'm submitting a ...**

  - [ ] bug report
  - [ ] feature request
  - [X] support request

- I confirm that I
  - [ ] used the [search](https://github.com/lukeautry/tsoa/search?type=Issues) to make sure that a similar issue hasn't already been submit

## Expected Behavior

```yaml

```

```ts
```

## Current Behavior

```yaml

```

```ts
```

## Steps to Reproduce

1. 2. 3. 4.

## Context (Environment)

Version of the library: 
Version of NodeJS: 

AudunWA:
### All Submissions:

**Closing issues**
closes #745

The issue was marked with `help wanted`.

This is a very simple implementation, it doesn't support mjs or TypeScript configs. I'm not sure what the best way to add support for these would be – an option for TS is to use ts-node.

I'd be happy to add some tests checking that the config gets loaded correctly, but I'd need a pointer on where I should put the test to best fit in with the existing ones.

**Test plan**

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -47,10 +47,12 @@ const authorInformation: Promise<
 
 const getConfig = async (configPath = 'tsoa.json'): Promise<Config> => {
   let config: Config;
+  const ext = path.extname(configPath);
   try {
-    const ext = path.extname(configPath);
     if (ext === '.yaml' || ext === '.yml') {
       config = YAML.load(configPath);
+    } else if (ext === '.js') {
+      config = await import(`${workingDir}/${configPath}`);
     } else {
       const configRaw = await fsReadFile(`${workingDir}/${configPath}`);
       config = JSON.parse(configRaw.toString('utf8'));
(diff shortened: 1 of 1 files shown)


packages/cli/src/cli.ts
```diff
@@ -47,10 +47,12 @@ const authorInformation: Promise<
 
 const getConfig = async (configPath = 'tsoa.json'): Promise<Config> => {
   let config: Config;
+  const ext = path.extname(configPath);
   try {
-    const ext = path.extname(configPath);
```
mrl5, line 51: if you would leave the assignment `ext = path.extname(configPath);` inside try-catch then it can be catched in line `61`: https://github.com/lukeautry/tsoa/blob/dcc225fd265fdacaab686d4ac359dde640663590/packages/cli/src/cli.ts#L61-L62

packages/cli/src/cli.ts
```diff
@@ -47,10 +47,12 @@ const authorInformation: Promise<
 
 const getConfig = async (configPath = 'tsoa.json'): Promise<Config> => {
   let config: Config;
+  const ext = path.extname(configPath);
```
mrl5, line 50: `let ext;`

mrl5:
hey @AudunWA, imo having feature described in #745 would be nice :)

I suggested one improvement to your code that would prevent crashing in case  `configPath` does not exist in the file system

packages/cli/src/cli.ts
AudunWA: `path.extname` doesn't actually try to find the file, and the only error it can throw is if you pass a non-string argument to it. (See https://nodejs.org/api/path.html#pathextnamepath).

Issue #1122, opened 2021-11-03
Error: Unknown type: IndexedAccessType


ngaer:
## Sorting

- **I'm submitting a ...**

  - [x] bug report
  - [ ] feature request
  - [x] support request

- I confirm that I
  - [x] used the [search](https://github.com/lukeautry/tsoa/search?type=Issues) to make sure that a similar issue hasn't already been submit

## Expected Behavior

The problem already was discussed in #862. Type resolver throws this error when you use a type like this:

```ts
export const MessageState: {
  pending: 'pending',
  available: 'available'
};

export type MessageState = (typeof MessageState)[keyof typeof MessageState]
```
```
Generate routes error.
 Error: Unknown type: IndexedAccessType
At: controllers/Controller.ts:16:16.
This was caused by 'type MessageState = (typeof MessageState)[keyof typeof MessageState]'
    at new GenerateMetadataError (/my_project/node_modules/@tsoa/cli/src/metadataGeneration/exceptions.ts:6:5)
    at TypeResolver.resolve (/my_project/node_modules/@tsoa/cli/src/metadataGeneration/typeResolver.ts:318:13)
    at TypeResolver.getTypeAliasReference (/my_project/node_modules/@tsoa/cli/src/metadataGeneration/typeResolver.ts:617:120)
    at TypeResolver.getReferenceType (/my_project/node_modules/@tsoa/cli/src/metadataGeneration/typeResolver.ts:585:30)
    at TypeResolver.resolve (/my_project/node_modules/@tsoa/cli/src/metadataGeneration/typeResolver.ts:359:32)
    at /my_project/node_modules/@tsoa/cli/src/metadataGeneration/typeResolver.ts:107:127
    at Array.reduce (<anonymous>)
    at TypeResolver.resolve (/my_project/node_modules/@tsoa/cli/src/metadataGeneration/typeResolver.ts:106:10)
    at TypeResolver.getTypeAliasReference (/my_project/node_modules/@tsoa/cli/src/metadataGeneration/typeResolver.ts:617:120)
    at TypeResolver.getReferenceType (/my_project/node_modules/@tsoa/cli/src/metadataGeneration/typeResolver.ts:585:30)
```

## Possible Solution

Solution suggested in #862:
```ts
// file: packages/cli/src/metadataGeneration/typeResolver.ts
// ↓added
if (ts.isIndexedAccessTypeNode(this.typeNode)) {
[...]

Issue #862, opened 2020-12-09
Generate routes error (IndexedAccessType + TypeOperator)


m-doi2:
## Sorting

- **I'm submitting a ...**

  - [x] bug report
  - [ ] feature request
  - [x] support request

- I confirm that I
  - [x] used the [search](https://github.com/lukeautry/tsoa/search?type=Issues) to make sure that a similar issue hasn't already been submit

## Current Behavior

This is a problem similar to the Enum problem that was improved with this pull request.
https://github.com/lukeautry/tsoa/pull/636

For example, if you use the `KeyOfKeyword` type generated using the object's key value as the schema, the Generate routes error will occur.

```ts
const genderCds: {
    readonly male: "male";
    readonly female: "female";
};

type GenderCd = typeof genderCds[keyof typeof genderCds];

export interface TestModel {
  genderCd: GenderCd;
}
```

```sh
Generate routes error.
 Error: Unknown type: IndexedAccessType
At: /mypath/to/xxx/index.d.ts:8:8.
This was caused by 'type GenderCd = typeof genderCds[keyof typeof genderCds];'
    at new GenerateMetadataError (/mypath/to/node_modules/@tsoa/cli/dist/metadataGeneration/exceptions.js:22:28)
    at TypeResolver.resolve (/mypath/to/node_modules/@tsoa/cli/dist/metadataGeneration/typeResolver.js:308:19)
    at TypeResolver.getTypeAliasReference (/mypath/to/node_modules/@tsoa/cli/dist/metadataGeneration/typeResolver.js:529:353)
    at TypeResolver.getReferenceType (/mypath/to/node_modules/@tsoa/cli/dist/metadataGeneration/typeResolver.js:505:38)
    at TypeResolver.resolve (/mypath/to/node_modules/@tsoa/cli/dist/metadataGeneration/typeResolver.js:341:34)
    at TypeResolver.resolve (/mypath/to/node_modules/@tsoa/cli/dist/metadataGeneration/typeResolver.js:294:131)
    at TypeResolver.propertyFromSignature (/mypath/to/node_modules/@tsoa/cli/dist/metadataGeneration/typeResolver.js:761:143)
    at /mypath/to/node_modules/@tsoa/cli/dist/metadataGeneration/typeResolver.js:727:161
    at Array.map (<anonymous>)
    at TypeResolver.getModelProperties (/mypath/to/node_modules/@tsoa/cli/dist/metadataGeneration/typeResolver.js:727:124)
```

[...]

PR #1160, opened 2021-12-21
fix(cli): support types indexed by their own keys

fgnass:
This PR addresses #1122 to support types indexed by their own keys, as they are generated by [Prisma](https://prisma.io) when a model uses enums.

In contrast to the solution proposed in #862, this approach does not rely on the values to equal their keys.

*NOTE:* It only handles the case `(typeof exp)[keyof typeof exp]`, i.e. when both type queries refer to the same expression. So types like `(typeof foo)[keyof typeof bar]` or `Foo[keyof Bar]` are still unsupported and will result in an error just as before.

### All Submissions:

### DIFF
--- a/packages/cli/src/metadataGeneration/typeResolver.ts
+++ b/packages/cli/src/metadataGeneration/typeResolver.ts
@@ -299,18 +299,16 @@ export class TypeResolver {
       }
     }
 
-    // Indexed by keyof typeof self
-    if (
-      ts.isIndexedAccessTypeNode(this.typeNode) &&
-      ts.isTypeOperatorNode(this.typeNode.indexType) &&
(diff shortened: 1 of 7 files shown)


fgnass:
I added a second commit to the PR since Prisma uses parenthesis around the `typeof value` expression, which got stripped by prettier in my test fixture. I added a second test with parenthesis and put a `//prettier-ignore` comment in front to keep them in place. I tested the patched version in a real-world Prisma project to make sure everything works as expected.

WoH:
Looks good so far, can you make sure (add a test case) `(typeof exp1)[keyof typeof exp2]` still throws?

fgnass:
@WoH ✅ I added a test for that case.

WoH:
LGTM, thanks for contributing!

PR #1163, opened 2021-12-28
fix(resolver): Resolve `keyof` using the type checker

### LINKED ISSUE #1162: regression: `keyof X` on `type X = { a: unknown; b: unknown }` produces `dataType: any`
## Sorting

- **I'm submitting a ...**

  - [x] bug report
  - [ ] feature request
  - [ ] support request

- I confirm that I

## Expected Behavior

```ts
export interface File {
  type: Types;
}

export type Types = keyof Items;

export type Items = {
  type1: unknown;
  type2: unknown;
};

@Route("foo")
export class FooController {
  @Get("bar")
  public async bar(): Promise<{ file: File }> {
    return { file: { type: "type1" } };
  }
}
```

The resulting routes and spec should be something like:

```ts
const models: TsoaRoute.Models = {
    "Types": {
        "dataType": "refAlias",
        "type": {"dataType":"enum","enums":["type1","type2"],"validators":{}},
    },
    "File": {
        "dataType": "refObject",
        "properties": {
            "type": {"ref":"Types","required":true},
        },
        "additionalProperties": false,
    },
};
const validationService = new ValidationService(models);
```

## Current Behavior

It produces `{"dataType":"any"}` in the routes as follows, and has a similar result in the spec.

```ts
const models: TsoaRoute.Models = {
    "Types": {
        "dataType": "refAlias",
        "type": {"dataType":"any","validators":{}},
    },
    "File": {
        "dataType": "refObject",
        "properties": {
            "type": {"ref":"Types","required":true},
        },
        "additionalProperties": false,
    },
};
const validationService = new ValidationService(models);
```

## When keyof works

Note that removing one of the two fields `type1` or `type2` from the `Items` type results in the expected output being generated:

```ts
export interface File {
  type: Types;
}

export type Types = keyof Items;

export type Items = {
  type1: unknown;
};

@Route("foo")
export class FooController {
  @Get("bar")
  public async bar(): Promise<{ file: File }> {
    return { file: { type: "type1" } };
  }
}
```

produces

```ts
const models: TsoaRoute.Models = {
    "Types": {
        "dataType": "refAlias",
        "type": {"dataType":"enum","enums":["type1"],"validators":{}},
    },
    "File": {
        "dataType": "refObject",
        "properties": {
            "type": {"ref":"Types","required":true},
        },
        "additionalProperties": false,
    },
};
const validationService = new ValidationService(models);
```

## Relation to 4.x

Although not strictly a regression in 4.x - I could reproduce with both 3.14.1 and 4.0.0-alpha.1 - I only noticed this bug when checking for regressions with the upgrade, and found that a model that previously produced the proper enum list is now producing the `any` dataType as shown above. The model in question is a little more complex and I have not yet narrowed down the case where the regression occurs, and if its caused by the TSOA upgrade or the related Typescript upgrade. However the model in question does use a union of two `keyof` types, so the changes to union handling could have affected the likelihood of encountering this bug.

## Steps to Reproduce

1. Set up a standard tsoa project.
2. Copy in controller and model as in the `expected behaviour` section
3. Run `yarn run tsoa routes`
4. Check generated routes file. 

## Context (Environment)

Version of the library: 3.14.1 and 4.0.0-alpha.1
Version of NodeJS: v14.18.2

- Confirm you were using yarn not npm: [✓]

WoH:
The old approach was to transform the node into a type and back into a
node to construct a simplified version. However, that approach can break
down silently and without warning.

Now, we try to explicitly get string enums and warn if we are unable to
proceed.

Fixes #1162

### All Submissions:

**Closing issues**

Closes #1162

### DIFF
--- a/packages/cli/src/metadataGeneration/exceptions.ts
+++ b/packages/cli/src/metadataGeneration/exceptions.ts
@@ -10,6 +10,14 @@ export class GenerateMetadataError extends Error {
   }
 }
 
+export class GenerateMetaDataWarning {
+  constructor(private message: string, private node: Node | TypeNode, private onlyCurrent = false) {}
+
+  toString() {
+    return `Warning: ${this.message}\n${prettyLocationOfNode(this.node)}\n${prettyTroubleCause(this.node, this.onlyCurrent)}`;
+  }
+}
+
 export function prettyLocationOfNode(node: Node | TypeNode) {
   const sourceFile = node.getSourceFile();
   const token = node.getFirstToken() || node.parent.getFirstToken();
--- a/packages/cli/src/metadataGeneration/typeResolver.ts
+++ b/packages/cli/src/metadataGeneration/typeResolver.ts
@@ -2,7 +2,7 @@ import * as ts from 'typescript';
 import { getJSDocComment, getJSDocComments, getJSDocTagNames, isExistJSDocTag } from './../utils/jsDocUtils';
 import { getDecorators, getNodeFirstDecoratorValue, isDecorator } from './../utils/decoratorUtils';
 import { getPropertyValidators } from './../utils/validatorUtils';
-import { GenerateMetadataError } from './exceptions';
+import { GenerateMetadataError, GenerateMetaDataWarning } from './exceptions';
 import { getInitializerValue } from './initializer-value';
 import { MetadataGenerator } from './metadataGenerator';
 import { Tsoa, assertNever } from '@tsoa/runtime';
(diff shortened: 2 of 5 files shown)


WoH:
If there is a way to construct something that is not a literal or a union of literals that I should be aware of, I'll add cases for that as well.

@devnev Please check if this does solve any `keyof` related issues.

devnev:
@WoH Upgrading to 4.0.0-alpha.2 resolved the issue I was seeing, thanks.

PR #1167, opened 2021-12-31
TS 4.5 Compiler API, strict tsconfig in test packages, chores

WoH:
Bit of a workaround to only check the test files in strict mode without affecting the path imports.

The main noticeable change is the treatment of remapped properties.

```ts
interface A {
  /**
   * @deprecated
   */
  one: string
}

type B = {
    [K in keyof A as `${K}Prop`]: boolean;
}
```

no longer marks `B.oneProp` as deprecated. This seems (?) to align with TypeScript's behavior.

@jenseng is this a regression or an acceptable breaking change in your opinion?


jenseng:
@WoH yeah that seems like a reasonable change, I feel like the old behavior was a more a side effect of the implementation than desirable behavior

PR #1169, opened 2022-01-07
Allow json form field

### LINKED ISSUE #1027: Is it possible to validate JSON.stringfied data sent in FormData, using interfaces? As if @Body()
Generally, I'm using @Body() to receive data. But if I try to FormData [@UplodedFile() and @FormField() together ], I'm losing validation with interfaces, which makes development a lot easier. 

I spend some time looking into source code to find a way to validate data manually, in the controller.
All I want is to parse received JSON string and validate it using one of the interface, if it is possible. Thanks.

## Sorting

- **I'm submitting a ...**

  - [ ] bug report
  - [ ] feature request
  - [X] support request

- I confirm that I

## Expected Behavior

TSOA runs good. This is just a question.

## Context (Environment)

Version of the library: ^3.6.1
Version of NodeJS: v14.17.1

josselinbuils:
It seems to be a quite common use case to send JSON metadata in a form field when uploading a file.

The users will still have to parse the JSON themselves but TSOA will be flexible enough to validate it if necessary.

Fixes #1027

### All Submissions:

- [X] Have you written unit tests? **Except for HAPI framework, I don't know how to test it.**
- [ ] Have you written unit tests that cover the negative cases (i.e.: if bad data is submitted, does the library respond properly)? **Normally not necessary, the open API validation being already tested.**
- [X] This PR is associated with an existing issue? #1027

**Closing issues**

- [X] Has the issue had a maintainer respond to the issue and clarify that the feature is something that aligns with the [goals](https://github.com/lukeautry/tsoa#goal) and [philosophy](https://github.com/lukeautry/tsoa#philosophy) of the project?

The chosen approach is quite light so the implementation did not add any complex logic to TSOA.

The counterpart is that users will have to parse themselves JSON and they won't be able to use `@UploadedFile` and `@UploadedFiles` annotations.

Indeed, those annotations add a middleware to parse multipart data and so they will conflict with the middleware added by the users to parse JSON in multipart requests.

The good news is that users will still be able to retrieve uploaded file quite easily by using the request annotation.

**Test plan**

I added integration tests for express and koa frameworks to test that JSON form fields pass the OpenAPI validation.

I don't know how to implement the parsing middleware on hapi framework, help is welcome.

### DIFF
--- a/packages/cli/src/metadataGeneration/parameterGenerator.ts
+++ b/packages/cli/src/metadataGeneration/parameterGenerator.ts
@@ -224,11 +224,7 @@ export class ParameterGenerator {
 
   private getFormFieldParameter(parameter: ts.ParameterDeclaration): Tsoa.Parameter {
     const parameterName = (parameter.name as ts.Identifier).text;
-    const type: Tsoa.Type = { dataType: 'string' };
-
-    if (!this.supportPathDataType(type)) {
-      throw new GenerateMetadataError(`Parameter '${parameterName}:${type.dataType}' can't be passed as form field parameter in '${this.method.toUpperCase()}'.`, parameter);
-    }
+    const type = this.getValidatedType(parameter);
 
     return {
       description: this.getParameterDescription(parameter),
--- a/tests/fixtures/controllers/postController.ts
+++ b/tests/fixtures/controllers/postController.ts
@@ -1,4 +1,5 @@
-import { Body, Patch, Post, Query, Route, File, UploadedFile, UploadedFiles, FormField } from '@tsoa/runtime';
+import { Body, Patch, Post, Query, Route, File, UploadedFile, UploadedFiles, FormField, Request } from '@tsoa/runtime';
+import { Request as ExpressRequest } from 'express';
 import { ModelService } from '../services/modelService';
 import { GenericRequest, TestClassModel, TestModel } from '../testModel';
 
@@ -65,6 +66,11 @@ export class PostTestController {
   @Post('ManyFilesAndFormFields')
   public async postWithFiles(@UploadedFiles('someFiles') files: File[], @FormField('a') a: string, @FormField('c') c: string): Promise<File[]> {
     return files;
+  }
+
+  @Post('FileAndJsonFormField')
+  public async postWithFileAndJsonFormField(@Request() req: ExpressRequest, @FormField('json') json: { property: string }): Promise<{ hasFile: boolean; json: { property: string } }> {
+    return { hasFile: !!req.file, json };
   }
 
   @Post('Location')
--- a/tests/fixtures/express/server.ts
+++ b/tests/fixtures/express/server.ts
@@ -1,6 +1,7 @@
 import * as bodyParser from 'body-parser';
 import * as express from 'express';
 import * as methodOverride from 'method-override';
+import * as multer from 'multer';
 import '../controllers/rootController';
 
 import '../controllers/optionsController';
@@ -36,6 +37,13 @@ app.use((req: any, res: any, next: any) => {
   req.stringValue = 'fancyStringForContext';
   next();
 });
+app.post('/v1/PostTest/FileAndJsonFormField', (req: express.Request, res: express.Response, next: express.NextFunction) => {
+  multer().single('file')(req, res, () => {
+    req.body.json = JSON.parse(req.body.json);
+    next();
+  });
+});
+
 RegisterRoutes(app);
 
 // It's important that this come after the main routes are registered
--- a/tests/fixtures/koa/server.ts
+++ b/tests/fixtures/koa/server.ts
@@ -1,4 +1,5 @@
 import * as Koa from 'koa';
+import * as multer from '@koa/multer';
 import * as KoaRouter from '@koa/router';
 import '../controllers/rootController';
 
@@ -28,6 +29,13 @@ const app = new Koa();
 app.use(bodyParser());
 
 const router = new KoaRouter();
+
+router.post('/v1/PostTest/FileAndJsonFormField', async (ctx, next) => {
+  await multer().single('file')(ctx, () => {
+    ctx.request.body.json = JSON.parse(ctx.request.body.json);
+  });
+  await next();
+});
 
 RegisterRoutes(router);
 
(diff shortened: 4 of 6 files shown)


packages/cli/src/metadataGeneration/parameterGenerator.ts
```diff
@@ -224,11 +224,7 @@ export class ParameterGenerator {
 
   private getFormFieldParameter(parameter: ts.ParameterDeclaration): Tsoa.Parameter {
     const parameterName = (parameter.name as ts.Identifier).text;
-    const type: Tsoa.Type = { dataType: 'string' };
-
-    if (!this.supportPathDataType(type)) {
-      throw new GenerateMetadataError(`Parameter '${parameterName}:${type.dataType}' can't be passed as form field parameter in '${this.method.toUpperCase()}'.`, parameter);
-    }
```
WoH, line 231: I am hesitant to commit to this as is. While it enables a specific use case it's very likely to be a pitfall more often than not. We should at least warn that the user has to take additional steps here.

WoH: I think we should use this instead of the error: https://github.com/lukeautry/tsoa/blob/master/packages/cli/src/metadataGeneration/exceptions.ts#L13

josselinbuils: Hello, I'm a bit lost about of that check: the `supportPathDataType` function accepts a lot for types that are not really managed by form fields no?

Regarding the check, is it possible to do it in a place where we can check if a JSON related type has been used without parsing the data?

WoH: `supportPathDataType` basically rejects types that are serialized into objects.
Initially that was intended for path data parsing, but the assumption for form fields was similar: By default, form fields values will be `string`s, which can trivially be coerced into simple types, but not into arrays, or objects, or a mix of both.
Since I don't think middleware will try to parse form field values into objects, the check makes sense (with a warning), unless we commit to trying to coerce a form field's string value as a json object during validation.

Jule-:
@WoH I am not aware of all the implications, but couldn't we consider that form fields are all supported path data types, and when it comes to objects and arrays, they must be JSON serialized?

Thus, as you mentioned, can't we try to unserialize during validation (catching errors as failure) and run validation like for body properties?

Because right now, what are our options to document an API like this:
```ts
public async myMethod(
  @FormField() ids: string[]
): Promise<void> {
  ...
}
```
If there is no change, the parameter is set to `string`, and the users have no insight into what should be inside unless we write some cumbersome description.

Maybe at least shouldn't we generate the documentation correctly as specified by the spec ([Multipart Requests](https://swagger.io/docs/specification/describing-request-body/multipart-requests/)) and eventually let the validation aside, like it is already the case... From an API user point of view, `string[]` in a form field without other explanation is necessarily a JSON serialized array. Am I right?
But I wouldn't recommend not converting the string correctly to the expected type since this will introduce unpredictable behaviours in endpoint implementations without the convertion+validation code on every parameter.

Jule-:
Can we re-open this one, or should we continue this conversation in the linked issue for more visibility?

Jule-:
@WoH, can you give your quick thoughts on this topic or even just acknowledge you read this and don't have time to answer? It would be amazing, thanks! :slightly_smiling_face:

WoH:
Iirc the comments were never addressed. They outline my issues. When it comes to the philosophy, see the referenced issue.

Jule-:
@WoH thank you for your response. :pray: 

I am not sure about what you are suggesting to me. Do you want me to continue this topic on the referenced issue and maybe set up a new PR to try something else? My previous comments try to elaborate on your comments (and I don't want to bloat the referenced issue unnecessarily), especially when you say:
> https://github.com/lukeautry/tsoa/pull/1169#discussion_r810508587
Since I don't think middleware will try to parse form field values into objects, the check makes sense (with a warning), unless we commit to trying to coerce a form field's string value as a json object during validation.

I can do something, but I don't want to go in a way you won't merge. What are you suggesting? As I have already exposed (https://github.com/lukeautry/tsoa/pull/1169#issuecomment-1327714804), I think we should commit to "coerce a form field's string value as a json object during validation".

> Thus, as you mentioned, can't we try to unserialize during validation (catching errors as failure) and run validation like for body properties?

WoH:
> Do you want me to continue this topic on the referenced issue and maybe set up a new PR to try something else?

Up to you. Both ways are fine here imo.

> Since I don't think middleware will try to parse form field values into objects, the check makes sense (with a warning), unless we commit to trying to coerce a form field's string value as a json object during validation.

We can try that, but will have to be careful with security then. We have traditionally not done this, the best example is the body parser which is not part of what we do, but leave to the developers. I'd consider this more plain old deserialization than coercion.

WoH:
Also: How do we distinguish this from urlencoded forms using the qs format?

Jule-:
Good point!

I can see 2 things, `form` is submitted as a query string or as a `multipart/form-data` (potentially with qs object-like encoded format), the latest being the one we are talking about here with `@FormField()`.

## For the query string:
Handle `@Query("user")` with qs object parsing (with [`query-string`](https://www.npmjs.com/package/query-string) or [`qs`](https://www.npmjs.com/package/qs) package) if the associated type is an object, e.g.:
```ts
@Query("user") user: { name: string }
```
But maybe this should be considered as another feature?

## For the multipart:
Use `multer` (if not already used in that case, potentially with `multer.none()`) that should handle this use case correctly and should transform
```http
Content-Disposition: form-data; name="user[name]"

John
```
to pass
```js
assert.deepEqual(req.body, { user: { name: "John" } });
```

In the end, we just need to check if the declared parameter is a `string` while it should be an `object` in order to try to JSON parse the field. If it is already an `object` then `multer` should have done its job on the qs object-like format. If not this should be because we received stringified JSON like that:
```http
Content-Disposition: form-data; name="user"

{"name":"John"}
```

### Alternative way

I am thinking about another approach by handling `Content-Type: application/json` in some way, but I don't know how `multer` is handling this kind of input.

```http
Content-Disposition: form-data; name="user"; filename="blob"
Content-Type: application/json

{"name":"John"}
```

But for this input, you have to append JSON as blob on the client side which is a bit unusual regarding the DX.

```js
formData.append('user', new Blob([ JSON.stringify({ name: "John" }) ], { type: "application/json" }));
```

-----

What do you think? Do you see some gaps or drawbacks?

I am open to suggestions. :slightly_smiling_face:

PR #1179, opened 2022-01-26
Fix issue/516

### LINKED ISSUE #516: specGenerator3 does not generate output for @BodyParam
When placing @BodyProp in a function's params, the params get outputted in Swagger v2 output mode but not in OpenAPI v3 mode. This is due to a lack of implementation in specGenerator3 as compared with specGenerator2. 

## Sorting

* **I'm submitting a ...**
  - [X] bug report
  - [ ] feature request
  - [ ] support request

* I confirm that I

## Expected Behavior
@BodyProp should make its way into the params in OpenAPI v3 output

## Current Behavior
No parameter or schema is emitted into the outputted OpenAPI v3 for @BodyProp entries

## Possible Solution
The code in specGenerator2 which emits BodyProps is missing in specGenerator3

pkotov:
### All Submissions:

**Closing issues**

Closes #516

### DIFF
--- a/packages/cli/src/swagger/specGenerator3.ts
+++ b/packages/cli/src/swagger/specGenerator3.ts
@@ -263,8 +263,9 @@ export class SpecGenerator3 extends SpecGenerator {
       pathMethod.security = method.security as any[];
     }
 
-    const bodyParams = method.parameters.filter(p => p.in === 'body');
-    const formParams = method.parameters.filter(p => p.in === 'formData');
+    const bodyParams: Tsoa.Parameter[] = method.parameters.filter(p => p.in === 'body');
+    const bodyPropParams: Tsoa.Parameter[] = method.parameters.filter(p => p.in === 'body-prop');
+    const formParams: Tsoa.Parameter[] = method.parameters.filter(p => p.in === 'formData');
 
     pathMethod.parameters = method.parameters
       .filter(p => {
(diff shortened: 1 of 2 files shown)


packages/cli/src/swagger/specGenerator3.ts
```diff
@@ -280,6 +281,28 @@ export class SpecGenerator3 extends SpecGenerator {
...
+          } as Tsoa.NestedObjectLiteralType,
+          validators: {},
+          deprecated: false,
+        });
+      }
+
+      const type: Tsoa.NestedObjectLiteralType = bodyParams[0].type as Tsoa.NestedObjectLiteralType;
+      bodyPropParams.forEach((bodyParam: Tsoa.Parameter) => {
+        type.properties.push(bodyParam as Tsoa.Property);
+      });
```
WoH, line 303: This should break in scenarios where we have a ref type as a body.

pkotov: Body and BodyProps couldn't be in same place. It will through exception. You have tests, just launch them....

WoH: The tests are executed automatically in CI, but thanks for the tip :)
But you are somewhat correct, we don't have any tests that have controller methods expecting both `Body` and `BodyProp`. Bit of an undocumented caveat, but I don't think we can expect that within this PR.

PR #1196, opened 2022-02-20
Implement `operationIdTemplate` config parameter.

### LINKED ISSUE #1005: Generated OpenAPI definitions include duplicate operationIds, which violates the OpenAPI spec
## Sorting

- **I'm submitting a ...**

  - [X] bug report
  - [ ] feature request
  - [ ] support request

- I confirm that I

NOTE: This is almost certainly a dup of the closed issue #183, however since there have been several unanswered requests to re-open that issue, I'm going ahead and filing a new one (since it seems those requests were missed).

## Expected Behavior

As required by OpenAPI, I expect `tsoa spec` to generate paths with operationIds that are unique across the entire API definition. However, if two different controllers have the same method name, `tsoa spec` will generate the same `operationId` for both, even though they are two distinct paths. Here's a trivial example which breaks today:

```ts
@Route('one')
class FirstController extends Controller {
    @Get()
    async get(): Promise<void> {}
}

@Route('two')
class SecondController extends Controller {
    @Get()
    async get(): Promise<void> {}
}
```

## Current Behavior

Both of the aforementioned methods are given the same operationId of `Get` in the generated swagger.json:

```json
...
	"paths": {
		"/one": {
			"get": {
				"operationId": "Get",
...
			}
		},
		"/two": {
			"get": {
				"operationId": "Get",
...
			}
		},
	},
...
```

This violates [the OpenAPI spec](https://swagger.io/docs/specification/paths-and-operations/), which says the following about operationIds (emphasis added):

> operationId is an optional unique string used to identify an operation. If provided, _these IDs must be unique among all operations described in your API._

## Possible Solution

I expect each `get()` method in the TypeScript source to be given a unique operationId which includes the route or controller name, for example "get_one" or "one_get" or something similar (preferably something that is moderately human-readable).

## Steps to Reproduce

As above

## Context (Environment)

Version of the library: 3.8.0
Version of NodeJS: 15.14.0

- Confirm you were using yarn not npm: [ ] <-- we use npm exclusively; that's irrelevant in this case though

## Detailed Description

Change the default method of generating `operationId`s in OpenAPI definitions so that they are unique (and still human-understandable) across the entire generated API definition. This could be accomplished by adding the class name or (preferably) a sanitized form of the route path as a prefix. The `operationId` should probably conform to standard variable-naming syntax (i.e. avoid special characters like `.` and `-`) so it's easily consumable by code generators for clients.

Another option might be to simply not emit `operationId`s at all by default, and only emit `operationId`s for methods with an explicit `@OperationId` decorator.

## Breaking change?

Yes, the proposed change may potentially cause breakage for clients which are doing codegen from OpenAPI definitions generated by Tsoa, since some will rely on the `operationId` to choose function/method names, and those names would therefore change.

However, this could be avoided by including a backward-compatibility flag in `tsoa.json` or as an option to `tsoa spec`. It could also be mitigated manually by adding `@OperationId`s to all affected methods in application server code.

BradenM:
### All Submissions:

**Closing issues**

Closes #1005

I personally ran into issues stemming from #1005 (namely, OAS3 client code-gen tools) and just applied a quick n' dirty patch (w/ patch-package) as a temporary workaround for a while.
However, after seeing the recent rc0 tag and realizing I would have to go update the patch (:disappointed: ), figured I would just go ahead and push up what I had though would be a good solution to the issue.

I thought it would make the most sense to expose a config parameter to the end user that would allow for defining a simple handlebars template for generating operation ids.

In a nutshell:

 - Solves #1005
 - Avoids "forcing" a different naming scheme that (while maybe solving the problem for most) causes other issues.
 - Provides the most flexibility while still keeping the config as simple as possible.
 - By using the current naming scheme as the default template, its a **non-breaking** change (:tada:).
 - Currently, just the controller name and method are exposed to the handlebars template (which I believe is all anyone would really need?).

**Test plan**

Additional tests have been added for the oas2/3 unit test suites for the default/current behavior and 'customized' behavior.

### DIFF
--- a/packages/cli/src/swagger/specGenerator.ts
+++ b/packages/cli/src/swagger/specGenerator.ts
@@ -1,5 +1,6 @@
 import { ExtendedSpecConfig } from '../cli';
 import { Tsoa, assertNever, Swagger } from '@tsoa/runtime';
+import * as handlebars from 'handlebars';
 
 export abstract class SpecGenerator {
   constructor(protected readonly metadata: Tsoa.Metadata, protected readonly config: ExtendedSpecConfig) {}
@@ -8,8 +9,17 @@ export abstract class SpecGenerator {
     return this.getSwaggerType(type);
   }
 
-  protected getOperationId(methodName: string) {
-    return methodName.charAt(0).toUpperCase() + methodName.substr(1);
+  protected buildOperationIdTemplate(inlineTemplate: string) {
+    handlebars.registerHelper('titleCase', (value: string) => (value ? value.charAt(0).toUpperCase() + value.slice(1) : value));
+    handlebars.registerHelper('replace', (subject: string, searchValue: string, withValue = '') => (subject ? subject.replace(searchValue, withValue) : subject));
+    return handlebars.compile(inlineTemplate, { noEscape: true });
+  }
+
+  protected getOperationId(controllerName: string, method: Tsoa.Method) {
+    return this.buildOperationIdTemplate(this.config.operationIdTemplate ?? '{{titleCase method.name}}')({
+      method,
+      controllerName,
+    });
   }
 
   public throwIfNotDataFormat(strToTest: string): Swagger.DataFormat {
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -121,6 +121,9 @@ export const validateSpecConfig = async (config: Config): Promise<ExtendedSpecCo
   config.spec.description = config.spec.description || (await descriptionDefault());
   config.spec.license = config.spec.license || (await licenseDefault());
   config.spec.basePath = config.spec.basePath || '/';
+  // defaults to template that may generate non-unique operation ids.
+  // @see https://github.com/lukeautry/tsoa/issues/1005
+  config.spec.operationIdTemplate = config.spec.operationIdTemplate || '{{titleCase method.name}}';
 
   if (!config.spec.contact) {
     config.spec.contact = {};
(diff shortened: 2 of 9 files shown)


BradenM:
@WoH @lukeautry 

No rush or anything :smile: -- but I figured I should mention that I intend to update the existing documentation and such regarding this PR. Still, before doing so, I want to make sure this is an 'accepted' solution (and make any potential code changes). Thanks!

WoH:
LGTM, thanks! 🥳

packages/cli/src/swagger/specGenerator.ts
```diff
@@ -8,8 +9,17 @@ export abstract class SpecGenerator {
     return this.getSwaggerType(type);
   }
 
-  protected getOperationId(methodName: string) {
-    return methodName.charAt(0).toUpperCase() + methodName.substr(1);
+  protected buildOperationIdTemplate(inlineTemplate: string) {
+    handlebars.registerHelper('titleCase', (value: string) => (value ? value.charAt(0).toUpperCase() + value.slice(1) : value));
+    handlebars.registerHelper('replace', (subject: string, searchValue: string, withValue = '') => (subject ? subject.replace(searchValue, withValue) : subject));
```
WoH, line 14: Ideally, we should doi this once, not every time we call `buildOperationIdTemplate`

mrgoonie:
I think it's not working anymore

SebastienGllmt:
@mrgoonie works for me. Make sure `operationIdTemplate` is inside the `spec` section and not a top-level config

PR #1221, opened 2022-03-26
fix(cli): indexed types by their own keys when using TypeReferences

### LINKED ISSUE #1220: Unknown type: IndexedAccessType with TypeReference
## Sorting

- **I'm submitting a ...**

  - [x] bug report
  - [ ] feature request
  - [ ] support request

- I confirm that I

## Expected Behavior

With #1122 and the related PR I expected tsoa@next (4.0.0-rc.1 was used testing this) to generate the same Signature for all 4 Methods on this Controller as all four use the feature added in the Issue at core.

```ts
import { Controller, Get, Route } from '@tsoa/runtime';

type X<T> = T[keyof T];

const b = { a: 'A', b: 'B', c: 'C' } as const;
type B = { readonly a: 'A'; readonly b: 'B'; readonly c: 'C' };

@Route('/')
export class TestController extends Controller {
  @Get('/test')
  test(): typeof b[keyof typeof b] {
    return 'C';
  }
  @Get('/test2')
  test2(): B[keyof B] {
    return 'B';
  }
  @Get('/test3')
  test3(): X<B> {
    return 'C';
  }
  @Get('/test4')
  test4(): X<typeof b> {
    return 'A';
  }
}
```

## Current Behavior

Currently only Case 1 works. As currently only cases where objectType and indexType are both TypeQueries work.

## Possible Solution

Fairly simple fix, instead of just checking for TypeQuery a second check for TypeReference should work.
This would still not support cases where object and index are different objects.

I will make a PR later today.

marudor:
### All Submissions:

- [ ] Have you written unit tests that cover the negative cases (i.e.: if bad data is submitted, does the library respond properly)? - I guess it's not applicable. No Idea how to find a negative case here

**Closing issues**

closes #1220

**Test plan**

I looked at the tests for the core feature and just added Reference and Generic usage.

### DIFF
--- a/packages/cli/src/metadataGeneration/typeResolver.ts
+++ b/packages/cli/src/metadataGeneration/typeResolver.ts
@@ -327,7 +327,13 @@ export class TypeResolver {
       }
       const declaration = this.current.typeChecker.getTypeOfSymbolAtLocation(symbol, this.typeNode.objectType);
       try {
-        return new TypeResolver(this.current.typeChecker.typeToTypeNode(declaration, undefined, ts.NodeBuilderFlags.NoTruncation)!, this.current, this.typeNode, this.context, this.referencer).resolve();
+        return new TypeResolver(
+          this.current.typeChecker.typeToTypeNode(declaration, undefined, ts.NodeBuilderFlags.NoTruncation)!,
+          this.current,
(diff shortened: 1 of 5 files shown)


WoH:
Thanks!

PR #1227, opened 2022-04-08
Add support for `termsOfService` Spec field

smkoyan:
At this moment `tsoa` don't adds `termsOfService` Spec info to the outputting `swagger.json`
This pull request add support for it.


smkoyan:
@lukeautry please take a look

WoH:
Assuming any review will do here ;)
For OpenAPI 3, we should validate the URL format, in both cases I'd prefer we make sure this actually is a string at runtime, else throw.

As a first step this LGTM.

smkoyan:
yeah I can add validation for url format checking

PR #1244, opened 2022-04-25
feat: add title to interface to inline requestBody and responses

### LINKED ISSUE #1243: add title to interface to inline requestBody and responses
## Sorting

- **I'm submitting a ...**

  - [ ] bug report
  - [x] feature request
  - [ ] support request

- I confirm that I

## Expected Behavior

Add title to all inline schemas (requestbody and responses) 
https://swagger.io/specification/#schema-object
to improve the code generation based on the yaml files.
see also https://github.com/OpenAPITools/openapi-generator/pull/12203

Right now the problem is that inline objects have an "undefined" behaviour when clients are generated, by adding a title to this kind of schemas, the output is more consistent when generating client code.

This is will be a non breaking change.

The title tag can be based on the route name or route path and just prefixes it with e.g. "VerifyUserResponse" or "VerifyUserRequestBody"

## Current Behavior

Right now there is no title tag added.

## Possible Solution

Add a title to the generated yaml output.

simllll:
### All Submissions:

**Closing issues**

closes #1243

This feature allows more consistent client generation as the body and response objects where without a name before and therefore client genreation lead to "inline objects" naming without a qualified name.

before: 
image
after: 
image

as this is an additional optional field, I think there are no potential problems

### DIFF
--- a/packages/runtime/src/swagger/swagger.ts
+++ b/packages/runtime/src/swagger/swagger.ts
@@ -217,7 +217,7 @@ export namespace Swagger {
 
   export interface Response3 {
     description: string;
-    content?: { [name: string]: Schema & Example };
+    content?: { [name: string]: Schema3 & { schema?: Schema } & Example };
     headers?: { [name: string]: Header3 };
   }
 
--- a/packages/cli/src/swagger/specGenerator.ts
+++ b/packages/cli/src/swagger/specGenerator.ts
@@ -112,7 +112,7 @@ export abstract class SpecGenerator {
     // An empty list required: [] is not valid.
     // If all properties are optional, do not specify the required keyword.
     return {
-      title,
+      ...(title && { title }),
       properties,
       ...(additionalProperties && { additionalProperties }),
       ...(required && required.length && { required }),
--- a/packages/cli/src/swagger/specGenerator.ts
+++ b/packages/cli/src/swagger/specGenerator.ts
@@ -59,7 +59,7 @@ export abstract class SpecGenerator {
     }
   }
 
-  protected getSwaggerType(type: Tsoa.Type): Swagger.Schema | Swagger.BaseSchema {
+  protected getSwaggerType(type: Tsoa.Type, title?: string): Swagger.Schema | Swagger.BaseSchema {
     if (type.dataType === 'void' || type.dataType === 'undefined') {
       return this.getSwaggerTypeForVoid(type.dataType);
     } else if (type.dataType === 'refEnum' || type.dataType === 'refObject' || type.dataType === 'refAlias') {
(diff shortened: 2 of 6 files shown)


simllll:
Offtopic: but locally the esm tests are failing when running npm run test

```
> tsoa-tests-esm@4.0.0-beta.0 pretest
> node --experimental-loader=ts-node/esm prepare.ts

(node:793294) ExperimentalWarning: --experimental-loader is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
TypeError: Unknown file extension ".ts" for /home/simon/WebstormProjects/tsoa/tests/esm/prepare.ts

```

is there anything special I need to consider?

simllll:
@WoH any feedback on that? :)

WoH:
I've looked at the PR, there are 2 things that I wanna get some feedback on: First I want to find out what all the unexpected ways are this can break stuff. Secondly, I'm not sure the location for the title determination is ideal, maybe we can do that earlier and attach it to the type?

simllll:
Thanks for coming back to me!

Regarding the first point: What do you think if we just hide this feature behind a flag, and only set the title if this flag is set? This would prevent all unknown side effects!

And about the second one: I was thinking and playing around about the place where to add this title prop, and it turned out this was the place with the least modifications and single places for the title definition. Adding it earlier in the type itself, made it a bit unclear (in my opinion) where the title is coming from and why, because an object can have multiple places where it can occur it also bloats the code quite a lot to have all the cases handled. Or did you mean something else?

simllll:
@WoH I added useTitleTagsForInlineObjects to the config and hide the feature behind this flag.

WoH:
Code looks good, but can you add a test that uses this setting?

simllll:
I've added two test cases, but I stumpled upon something else, it seems the schema definition is not 100% correct.
I've fixed it swagger.ts on line 220, but I think it could also be wrong at line 215.

packages/runtime/src/swagger/swagger.ts
```diff
@@ -217,7 +217,7 @@ export namespace Swagger {
 
   export interface Response3 {
     description: string;
-    content?: { [name: string]: Schema & Example };
+    content?: { [name: string]: { schema?: Schema } & Example };
```
simllll, line 220: can schema really be optional? It's as optional everywehre else, therefore I did it here too.. but not sure if needed

packages/runtime/src/swagger/swagger.ts
```diff
@@ -217,7 +217,7 @@ export namespace Swagger {
 
   export interface Response3 {
     description: string;
-    content?: { [name: string]: Schema & Example };
+    content?: { [name: string]: Schema3 & { schema?: Schema } & Example };
```
simllll, line 220: can schema really be optional? It's as optional everywehre else, therefore I did it here too.. but not sure if needed

WoH: Maybe it is simply related to partial responses being built?
I don't really know an example where there wouldn't be a schema

simllll: I've addressed this in a second PR https://github.com/lukeautry/tsoa/pull/1253 
schema is not optional there anymore, as there is not situation where this can be the case

simllll:
Any plan when this will land in the master / release? Really looking forward to this ...

PR #1258, opened 2022-06-04
Generates a json schema to improve developer experience

arthurfiorette:
This is a pretty simple pull request that auto generates a schema.json to include autocompletions to tsoa json configuration file.

I think you should also include the `$schema` property in every documentation example. (But that lives under another repo)

```json
{
  "$schema": "./node_modules/tsoa/schema.json",
  ...
}
```

Example:

image


WoH:
Love the idea, but shouldn't we treat the produced JSON as a build artifact and ignore it in git?

arthurfiorette:
I don't think so, this way people can link to specific versions using raw.githubusercontent

tsimbalar:
Maybe it should be hosted with the documentation ? 

https://tsoa-community.github.io/docs/

arthurfiorette:
That's also another good option. Turborepo does it.

https://turborepo.org/schema.json

But that's probably because their documentation lives under the same repo as the source.

WoH:
We can link into specific files on npm via socket.dev's file explorer.

https://socket.dev/npm/package/@tsoa/runtime/files/4.1.0/dist/config.d.ts

arthurfiorette:
Did you guys decie what option should be done?

tsimbalar:
I was going to suggest maybe adding it to https://www.schemastore.org , because most IDEs just know how to point to it ... but it turns out it's there already !!

see https://json.schemastore.org/tsoa.json or https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/tsoa.json 

Maybe it just needs to be updated there ?

Issue #1147, opened 2021-12-02
Koa generated routes overlaps on fixed & variable paths


ClemWiz:
## Sorting

- **I'm submitting a ...**

  - [x] bug report
  - [ ] feature request
  - [ ] support request

- I confirm that I
  - [x] used the [search](https://github.com/lukeautry/tsoa/search?type=Issues) to make sure that a similar issue hasn't already been submit

## Expected Behavior
Given this controller:
```ts
import { Controller, Get, Path, Route } from "tsoa";

@Route("/test")
export class TestController extends Controller {
    @Get("/all")
    async getAll(): Promise<number[]> {
        return [1,2,3];
    }

    @Get("/{id}")
    async getOne(@Path() id: number): Promise<number> {
        return id;
    }
}
```

I should receive
```json
[1, 2, 3]
```
when consuming `/test/all`;

## Current Behavior
But I receive:

```json
{
    "fields": {
        "id": {
            "message": "invalid float number",
            "value": "all"
        }
    }
}
```
As i set breakpoints in debug mod, I can see that the first route is called, then the second one and the validation fails on the second call. I feel like there is a next/middleware call that shouldn't happen, but I fail to capture the way `tsoa` is built and generate the routes' templates.

## Possible Solution
I know it's not a real solution, but if I comment the line
`return next();`
in the `returnHandler(...)` method of the generated `routes.ts`, I have the correct behavior.

## Steps to Reproduce
To reproduce the current behavior:
1. You can clone this [repo](https://github.com/ClemWiz/tsoa-paths)
2. Launch the server using `yarn start`
3. Use `curl --location --request GET 'localhost:8080/test/all'`

If you want the correct behavior (without using `tsoa`), you can:
1. Comment the l.12 `RegisterRoutes(router);` 
2. Un-comment the l.9-10.
3. Use `curl --location --request GET 'localhost:8080/test/all'`

## Context (Environment)

Version of the library: 3.14.1
Version of NodeJS: 14.16.1

- Confirm you were using yarn not npm: [x]

## Detailed Description

Has I said, I'm not familiar with the `tsoa` source code and I fail to understand a lot of mechanics used here. So I can't provide any good change that won't brake the whole purpose of the library. I can only guess that - at some point - the generated template tools functions keep calling `next` until they either fail to match the route path (so route is considered "consumed") or fail to validate the given arguments (so an exception is raised).

If my guesses are right, maybe we should not allow the first behavior and stop calling next when we matched at least one path.
[...]

PR #1265, opened 2022-06-15
Fix overlaping paths with koa

ClemWiz:
### All Submissions:

**Closing issues**
closes https://github.com/lukeautry/tsoa/issues/1147

There is an [existing PR](https://github.com/lukeautry/tsoa/pull/1262) already opened but I wanted to propose another approach.
I focus only on the template and don't change the logic in `routeGenerator.ts`. I think it allows more readability in the generator without adding somme koa/express/hapi specific code that could be handled solely in the templates.

It also allows to do the same on 2.x & 3.x with some tweaks (based on template's versions).

**Test plan**

I'm trying to implement integration tests, but I failed to write tests that reproduce the actual behavior.


Shiou-Ju:
Hi @ClemWiz! 

In my opinion, it would be a courtesy to just comment on my PR rather than opening a new one based on my solution.

PR #1279, opened 2022-06-30
Update file uploads for ESM

FrozenKiwi:
The multer used in fileUploads generated a 'require' statement, which is not legal in ESM packages.

In a local test, simply changing from require to default import fixed the issue

### All Submissions:

**Closing issues**

`closes 1278`

### DIFF
--- a/packages/cli/src/routeGeneration/templates/express.hbs
+++ b/packages/cli/src/routeGeneration/templates/express.hbs
@@ -22,7 +22,7 @@ import { IocContainer, IocContainerFactory } from '@tsoa/runtime';
 import type { RequestHandler } from 'express';
 import * as express from 'express';
 {{#if useFileUploads}}
-const multer = require('multer');
+import multer from 'multer';
 const upload = multer({{{json multerOpts}}});
 {{/if}}
 
--- a/packages/cli/src/routeGeneration/templates/express.hbs
+++ b/packages/cli/src/routeGeneration/templates/express.hbs
@@ -22,7 +22,7 @@ import { IocContainer, IocContainerFactory } from '@tsoa/runtime';
 import type { RequestHandler } from 'express';
 import * as express from 'express';
 {{#if useFileUploads}}
-import multer from 'multer';
+import * as multer from 'multer';
 const upload = multer({{{json multerOpts}}});
 {{/if}}
 
--- a/packages/cli/src/routeGeneration/templates/express.hbs
+++ b/packages/cli/src/routeGeneration/templates/express.hbs
@@ -19,10 +19,9 @@ const promiseAny = require('promise.any');
 import { iocContainer } from '{{iocModule}}';
 import { IocContainer, IocContainerFactory } from '@tsoa/runtime';
 {{/if}}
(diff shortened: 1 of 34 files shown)


FrozenKiwi:
Ok, this PR became much bigger than initial anticipated.

When loading into an ESM-native node project, the `import * as mutler from 'multer'` syntax doesn't work, because according to the spec this is supposed to be a namespace which doesn't support direct function calls.  See https://www.typescriptlang.org/tsconfig#esModuleInterop

The fix is to add `esModuleInterop`, which unfortunately then breaks all the tests as they rely on the older method of interop.

After spending an afternoon trying to get this working without forcing the `esModuleInterop` flag upstream, I can't see another way to get this working.

I'd understand if you drop this PR, however as ESM is the future it might pay to move this 1 step closer to compliance

packages/cli/src/routeGeneration/templates/express.hbs
```diff
@@ -19,10 +19,9 @@ const promiseAny = require('promise.any');
 import { iocContainer } from '{{iocModule}}';
 import { IocContainer, IocContainerFactory } from '@tsoa/runtime';
 {{/if}}
-import type { RequestHandler } from 'express';
```
FrozenKiwi, line 22: minor unrelated suggestion, only importing the type makes the intent of the file clearer

packages/cli/src/routeGeneration/templates/koa.hbs
```diff
@@ -24,7 +24,7 @@ import { iocContainer } from '{{iocModule}}';
 import { IocContainer, IocContainerFactory } from '@tsoa/runtime';
 {{/if}}
 import type { Middleware } from 'koa';
-import * as KoaRouter from '@koa/router';
```
FrozenKiwi, line 27: I don't know Koa so can't test this, but I think it's covered in the tests and I assume it's better to have a consistent import usage across the code base?

tests/tsconfig.json
```diff
@@ -9,6 +9,7 @@
     "downlevelIteration": true,
     "experimentalDecorators": true,
     "emitDecoratorMetadata": true,
+    "esModuleInterop": true,
```
FrozenKiwi, line 12: note, change required to compile tests with `import XXX from 'XXX'`

ClementValot:
Isn't this 

`{{#if esm}}
import multer from 'multer';
{{else}}
const multer = require('multer');
{{/if}}`

an easier fix?

FrozenKiwi:
Possibly?

The rest of the imports are `import`, not `require`, and there didn't seem to be a strong advantage to leaving this one tag in an obsolete format.

https://github.com/lukeautry/tsoa/blob/057c57b951d01f32c49a5ae801a29a46162b6696/packages/cli/src/routeGeneration/templates/express.hbs#L22

FrozenKiwi:
I guess it wouldn't have then been necessary to modify the `tsconfig`.  Honestly, it didn't occur to me, which is a little silly.  Up to you which way you want to fix this.

FrozenKiwi:
See new PR

Issue #1301, opened 2022-08-29
No routes generated when following the Getting started guide


amda-phd:
## Sorting

- **I'm submitting a ...**

  - [x] bug report
  - [ ] feature request
  - [x] support request

- I confirm that I
  - [x] used the [search](https://github.com/lukeautry/tsoa/search?type=Issues) to make sure that a similar issue hasn't already been submit

## Expected Behavior

The `routes.ts` file is produced and contains models and routes based on the provided controllers.

```json
\\ package.json
{
  "name": "tsoa-hello-world",
  "main": "build/src/server.js",
  "scripts": {
    "build": "tsoa spec-and-routes && tsc",
    "start": "node build/src/server.js"
  },
  "dependencies": {
    "body-parser": "^1.20.0",
    "express": "^4.18.1",
    "tsoa": "^4.1.1"
  },
  "devDependencies": {
    "@types/body-parser": "^1.19.2",
    "@types/express": "^4.17.13",
    "@types/node": "^18.7.13",
    "typescript": "^4.7.4"
  }
}
```

```json
// tsoa.json
{
  "entryFile": "src/server.ts",
  "noImplicitAdditionalProperties": "throw-on-extras",
  "controllerPathGlobs": ["src/**/*Controller.ts"],
  "spec": {
    "outputDirectory": "build",
    "specVersion": 3
  },
  "routes": {
    "basePath": "/api/v1",
    "routesDir": "src",
    "middleware": "express"
  }
}
```

```ts
\\ healthController.ts
import { Get, Route, Controller } from "tsoa";
import { Health } from "./health";
import { HealthService } from "./healthService";

@Route("health")
export class HealthController extends Controller {
  @Get()
  public async getHealth(): Promise<Health> {
    return new HealthService().get();
  }
}
```

A complete sample of a simplified implementation of the tsoa's Getting Started Guide can be found [here](https://github.com/amda-phd/tsoa-hello-world).

## Current Behavior

Running `yarn tsoa routes` produces a `routes.ts` file, but it doesn't contain any of the information from the controllers or any route.

```ts
// routes.ts
import { Controller, ValidationService, FieldErrors, ValidateError, TsoaRoute, HttpStatusCodeLiteral, TsoaResponse, fetchMiddlewares } from '@tsoa/runtime';
import type { RequestHandler } from 'express';
import * as express from 'express';

const models: TsoaRoute.Models = {
};
const validationService = new ValidationService(models);

[...]

PR #1303, opened 2022-08-29
fix: throw an error when there are no controllers found

simllll:
With the ts4.8 issue it turns out tsoa does not throw any kind of error if there are zero controllers found. Which makes it even a bit hard to find the issue in first place. To prevent this in the future I added a check that throws an Error if there are no controllers discovered.

### All Submissions:

**Closing issues**

potentially helps e.g. for #1301

In my opinion it doesn't make sense to use tsoa without any controllers, but maybe I'm missing some special use cases?

**Test plan**

the added test catches the expected error when there are no controllers


WoH:
https://github.com/TypeStrong/ts-node/issues/1839

simllll:
quite some dependency updates needed to get this running again.
* dropped node 12 from test matrix (as it does not work with node 12 anymore)
* upgraded all deps (except chalk because it's esm only in version 5) to latest

should we maybe move that to a seperat PR?

WoH:
Don't think this is only related to win (I hope)
https://github.com/lukeautry/tsoa/blob/master/packages/cli/src/metadataGeneration/metadataGenerator.ts#L41

simllll:
It seems though it's only related to windows, it finishes successfully on linux and mac on the github actions (locally also on linux, tests are running fine).

simllll:
Still fails on windows somehow with superagent.
```
tsoa-tests: err Error: read ECONNRESET
tsoa-tests:     at TCP.onStreamRead (node:internal/stream_base_commons:217:20) {
tsoa-tests:   errno: -4077,
tsoa-tests:   code: 'ECONNRESET',
tsoa-tests:   syscall: 'read',
tsoa-tests:   response: undefined
tsoa-tests: }
```

first was thinking this could be related https://github.com/visionmedia/supertest/issues/617 but there is actually no http2.. kinda hard to figure it out on a non-windows env...

PR #1306, opened 2022-09-03
Add root security option

tboothman:
This adds a new config field `spec.rootSecurity` to tsoa.json that lets you apply OpenAPI security to every controller/method.
The security is layered and you can override it in the controller or method.
e.g. `@NoSecurity()` on a controller disables the root level security, and you can turn it back on with `@Security()` on a method in that controller.

This does not add the root level security to the OpenAPI document, but it does add it to every method where it's used.

### All Submissions:

Some slightly weird config passing in cli.js because the config is in the swagger spec section but is used by both swagger and routes generation.

**Test plan**
Covered all the combinations in some new unit tests.
I tried out the cli for spec, route and both and it works as expected.

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -144,6 +144,22 @@ export const validateSpecConfig = async (config: Config): Promise<ExtendedSpecCo
     config.spec.contact.url = config.spec.contact.url || author?.url;
   }
 
+  if (config.spec.rootSecurity) {
+    if (!Array.isArray(config.spec.rootSecurity)) {
+      throw new Error("spec.rootSecurity must be an array");
+    }
+
+    if (config.spec.rootSecurity) {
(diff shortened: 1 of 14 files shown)


tboothman:
Any feedback on this? Would like to get it sorted before I forget what I did.
Happy to undo the swagger type changes that weren't necessary or whatever.
I'll make a PR for the documentation/website when it's done to explain how to use it. The current docs are lacking as they don't mention you can put `@Security` on a controller which was why I was excited to fix this issue because adding `@Security` to every method is a pain.

WoH:
The TS 4.8 test fails are not really concerning, we know how/why they are what they are.
However, I would prefer if we could have some integration tests where we can demonstrate an integration of rootSecurity at runtime

WoH (approved):

PR #1309, opened 2022-09-12
Add new Queries decorator to include all query parameters in one sing…

daniseijo:
This PR adds a new `@Queries` decorator that accepts an object and wraps all the query parameters into it. This is something very useful for example if the request has a lot of filter parameters that we want to handle. It has been requested several times for example in [this issue](https://github.com/lukeautry/tsoa/issues/353).

I have tried to keep this new decorator behavior as similar as the Body and the BodyProps decorators were implemented.

Use example:

```typescript
export interface QueryParams {
  numberParam: number;
  stringParam: string;
  booleanParam: boolean;
  optionalStringParam?: string;
}


@Get('AllQueriesInOneObject')
public async getAllQueriesInOneObject(@Queries() queryParams: QueryParams) {
  const model = new ModelService().getModel();
  model.optionalString = queryParams.optionalStringParam;
  model.numberValue = queryParams.numberParam;
  model.boolValue = queryParams.booleanParam;
  model.stringValue = queryParams.stringParam;

  return model;
}
```
### All Submissions:

- [X] This PR is associated with an existing issue? (https://github.com/lukeautry/tsoa/issues/353)

- [X] Has the issue had a maintainer respond to the issue and clarify that the feature is something that aligns with the [goals](https://github.com/lukeautry/tsoa#goal) and [philosophy](https://github.com/lukeautry/tsoa#philosophy) of the project?

I have not tested yet if it would be a problem to have both the Queries and the Query decorator in one single function. It would be nice to add some test cases testing this.

**Test plan**

For what I wanted, the code is working as expected. I have tried to keep this new decorator as similar as you have implemented the Body and the BodyProps decorators. I still want to add more tests to check the doc generation but wanted to check if this approach was the correct one before continuing.

### DIFF
--- a/packages/cli/src/swagger/specGenerator2.ts
+++ b/packages/cli/src/swagger/specGenerator2.ts
@@ -299,13 +299,13 @@ export class SpecGenerator2 extends SpecGenerator {
     return parameter;
   }
 
-  private buildQueriesParameter(source: Tsoa.Parameter): Swagger.Parameter3[] {
-    if (source.type.dataType === 'refAlias' && source.type.type.dataType === 'nestedObjectLiteral') {
-      const properties = source.type.type.properties;
+  private buildQueriesParameter(source: Tsoa.Parameter): Swagger.Parameter2[] {
+    if (source.type.dataType === 'refObject' || source.type.dataType === 'nestedObjectLiteral') {
+      const properties = source.type.properties;
 
       return properties.map(property => this.buildParameter(this.queriesPropertyToQueryParameter(property)));
     }
-    return [];
+    throw new Error(`Queries '${source.name}' parameter must be an object.`);
   }
 
   private buildParameter(source: Tsoa.Parameter): Swagger.Parameter2 {
--- a/packages/cli/src/swagger/specGenerator2.ts
+++ b/packages/cli/src/swagger/specGenerator2.ts
@@ -173,11 +173,21 @@ export class SpecGenerator2 extends SpecGenerator {
       pathMethod.security = method.security;
     }
 
+    const queriesParams: Tsoa.Parameter[] = method.parameters.filter(p => p.in === 'queries');
+
     pathMethod.parameters = method.parameters
       .filter(p => {
-        return !(p.in === 'request' || p.in === 'body-prop' || p.in === 'res');
+        return ['request', 'body-prop', 'res', 'queries'].indexOf(p.in) === -1;
       })
       .map(p => this.buildParameter(p));
+
+    if (queriesParams.length > 1) {
+      throw new Error('Only one queries parameter allowed per controller method.');
+    }
+
+    if (queriesParams.length === 1) {
+      pathMethod.parameters.push(...this.buildQueriesParameter(queriesParams[0]));
+    }
 
     const bodyPropParameter = this.buildBodyPropParameter(controllerName, method);
     if (bodyPropParameter) {
@@ -287,6 +297,15 @@ export class SpecGenerator2 extends SpecGenerator {
       parameter.schema.required = required;
     }
     return parameter;
+  }
+
+  private buildQueriesParameter(source: Tsoa.Parameter): Swagger.Parameter3[] {
+    if (source.type.dataType === 'refAlias' && source.type.type.dataType === 'nestedObjectLiteral') {
+      const properties = source.type.type.properties;
+
+      return properties.map(property => this.buildParameter(this.queriesPropertyToQueryParameter(property)));
+    }
+    return [];
   }
 
   private buildParameter(source: Tsoa.Parameter): Swagger.Parameter2 {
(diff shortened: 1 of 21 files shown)


WoH:
Looks great!
I'll write down some questions which I am asking myself, maybe some you already considered.

Validation: Seems to make sense, I'll have to check what happens if we combine with `@Query()` under different excess prop handling scenarios

Spec: No immediate questions, we can add that as another query parameter I presume.
Maybe need to check if it aligns with how express etc. would parse it.

In general I think the easiest way to deal with the combination of several Query and Queries mixes is to not allow it. Otherwise this may blow up in complexity.

There's an issue similar to this which proposed `@Query('*')` or something similar, but I like the explicit `@Queries` better.

daniseijo:
Hi! Thanks for the quick response, I would love a bit of help with the swagger part. I have created a method to spread the properties of the Queries object to show its internal query parameters on the swagger document. However, I'm not sure if this is the best approach.

Could you check it out and also help me with the next steps to get this up and ready?

PD.: I will try to add more negative test cases to check all the variations but I think I will not have time to advance until next week.

daniseijo:
> Do we have any updates here? I really like this solution 😂

Hi, I implemented a solution for the swagger part but I am not sure if there is a cleaner way. If it is ok, maybe @WoH could help me with the next steps and have it merge.

WoH:
Thanks for the ping, I'll try to find time over the weekend to leave suggestions if needed

WoH:
Can you rebase this PR? CI on master should work again

packages/cli/src/swagger/specGenerator3.ts
```diff
@@ -458,6 +467,15 @@ export class SpecGenerator3 extends SpecGenerator {
     return mediaType;
   }
 
+  private buildQueriesParameter(source: Tsoa.Parameter): Swagger.Parameter3[] {
+    if (source.type.dataType === 'refAlias' && source.type.type.dataType === 'nestedObjectLiteral') {
```
WoH, line 471: 'refObject' should probably work here. 
We should not only check 1 level deep on the alias, I've written similar code which checks deeep for isVoidType I think

If we can't handle, I'd rather throw than return an array I think?

packages/cli/src/swagger/specGenerator3.ts
```diff
@@ -516,6 +534,8 @@ export class SpecGenerator3 extends SpecGenerator {
         const exampleLabel = parameterExampleLabels?.[currentIndex];
         return { ...acc, [exampleLabel === undefined ? `Example ${exampleCounter++}` : exampleLabel]: { value: ex } };
       }, {});
+
+      console.log(parameter.examples);
```
WoH, line 538: remove

packages/cli/src/swagger/specGenerator2.ts
```diff
@@ -173,12 +173,22 @@ export class SpecGenerator2 extends SpecGenerator {
       pathMethod.security = method.security;
     }
 
+    const queriesParams: Tsoa.Parameter[] = method.parameters.filter(p => p.in === 'queries');
```
WoH, line 176: see spec3

WoH:
I think I've mentioned this, but we really don´t want both `@Query` and `@Queries` at the same time.
It'll be impossible to manage all that complexity when it comes to what belongs where for now.
For test cases: Please make sure `{ [name: string]: unknown/any }` (basically Records) can be used to wildcard anything.
Maybe `{ [name: string]: number }` and handle that which will be fun.

matthewyu-ptr:
This is a good feature, please don't close this PR

aakash-at:
Any update on this PR?

WoH:
Just to be clear, this was not a no from my side. Just throw an Error of both Decorators are parsed in the same method

daniseijo:
Sorry, my intention is to fix the changes but I left my work where I worked with tsoa. Since then I have had no time to continue with this. I will try to take some time to read and solve the issues on the PR as soon as possible. Sorry again for the delay.

packages/cli/src/swagger/specGenerator3.ts
daniseijo: Thinking about this, I think is better to fail if there is a nested object, to be coherent with the `@Query` decorator failing if we pass it a `refObject`. @WoH What do you think?

daniseijo:
Hi, I already made the changes to the code and added some more tests. @WoH, I wrote a response to one of your comments. Check it out and, if you think that behavior does not make sense, tell me and I will change it.

In any case, I think the failing tests in some environments are not my fault. Whenever you can, check the PR again and let me now if I am missing something.

WoH (approved):

u12206050:
Awesome! Can't wait for this to be published!

u12206050:
@WoH  Any chance you could push a new release with this then we can test it? :) Thanks for the work @daniseijo

filipRisteski:
I wanted to write something like this, but since it's here it would be very much appreciate if we can speed this up :) Thanks for the hard work @daniseijo

WoH:
it's available! `yarn add tsoa@next`

bsberry:
@daniseijo Thank you so much for implementing this. The lack of this almost stopped our adoption of tsoa cold.

grug:
Is there any idea of when 5.0.0 will come out of RC and released?

PR #1311, opened 2022-09-15
Fix 4.8 CI

WoH:


WoH:
Does not fix whatever is happening using koa + windows

PR #1331, opened 2022-10-26
Use a variable to determine if a type should be added or not when resolved.

### LINKED ISSUE #1330: Documenting Types which are not actually used by the API
## Sorting

- **I'm submitting a ...**

  - [X] bug report
  - [ ] feature request
  - [X] support request

- I confirm that I

## Expected Behavior

If I create a type and an interface which extends it, and only the interface is used as a return type, then only the interface should appear in swagger.json.

Consider the following type/interface definition and controller method:

```
export type defaultExcludeAttributesType =
  | 'createdAt'
  | 'updatedAt'
  | 'deletedAt';

export interface AddressAttributes {
  addressId: string;
  line1?: string;
  line2?: string;
  line3?: string;
  city?: string;
  stateProvince?: string;
  postalCode?: string;
  country?: string;
  createdAt?: Date;
  updatedAt?: Date;
  deletedAt?: Date;
}

 export interface AddressResponse extends Omit<AddressAttributes, defaultExcludeAttributesType> {}

@Route('/addressess')
export class AddressController extends Controller {
  @Get("{addressId}")
  public async getAddress(@Path("addressId") addressId: string): Promise<AddressResponse> {
    return new AddressResponse();
  }
}

```

I would expect ONLY AddressResponse to be in swagger.json since it is the only type/interface to be returned from a method.

## Current Behavior

Given the above, tsoa is generating types for both of those, such as the following:

```
"schemas": {
	"Pick_AddressAttributes.Exclude_keyofAddressAttributes.defaultExcludeAttributesType__": {
		"properties": {
			"addressId": {
				"type": "string"
			},
			"line1": {
				"type": "string"
			},
			"line2": {
				"type": "string"
			},
			"line3": {
				"type": "string"
			},
			"city": {
				"type": "string"
			},
			"stateProvince": {
				"type": "string"
			},
			"postalCode": {
				"type": "string"
			},
			"country": {
				"type": "string"
			}
		},
		"required": [
			"addressId"
		],
		"type": "object",
		"description": "From T, pick a set of properties whose keys are in the union K"
	},
	"AddressResponse": {
		"properties": {
			"addressId": {
				"type": "string"
			},
			"line1": {
				"type": "string"
			},
			"line2": {
				"type": "string"
			},
			"line3": {
				"type": "string"
			},
			"city": {
				"type": "string"
			},
			"stateProvince": {
				"type": "string"
			},
			"postalCode": {
				"type": "string"
			},
			"country": {
				"type": "string"
			}
		},
		"required": [
			"addressId"
		],
		"type": "object",
		"additionalProperties": false
	}
},

```

I believe I understand why, which is that typeResolver.resolve always resolves recursively for most typeReferences, but this ends up creating extra types that are not needed.  When a type is an interface (and not an alias), the type always has the correct properties, and does not use a ref.

## Possible Solution

Ideally there would be some process, either as part of resolving, or after all types are resolved, where if a type is not the subject of a ref, either from a controller method or some other object (base type, property, etc), then it will be skipped or removed.

Perhaps also add a decorator that would indicate that this type should always be serialized even if not used as a parameter, return, or ref.

## Steps to Reproduce

Create a project with the above types and controller.  Generate the swagger for it.  I can share a repo if needed.

## Context (Environment)

Version of the library: 4.1.3
Version of NodeJS: 14.17.1

- Confirm you were using yarn not npm: [X]

## Detailed Description

Do not document schema in swagger.json unless they are actually needed.  I believe this would mean if they are not a property or return type, and not referenced by any other schema, they can be skipped.

## Breaking change?

This will change the behavior, by documenting fewer schema, but they should be schema that are not used.  This should not break anything.

dpomerantz-jci:
### All Submissions:

**Closing issues**

closes #1330 

**Test plan**

Negative test was added, though it is relatively brittle.  The goal of this change is to ensure that heritage types for interfaces do not get added to the refType map simply because they are heritage types.  To test this, first a new model was created, HeritageTestModel2.  This was done so that any changes changes to or expectations upon HeritageTestModel will not affect this test.  HeritageTestModel extends HeritageBaseModel. In the fixture that includes HeritageTestModel, a test was added to ensure that HeritageTestModel2 is created, but HeritageBaseModel is not.  This can be broken if HeritageTestModel2's heritage is changed, and the the test will still pass.  It can also be broken be adding HeritageBaseModel to the model so that it is included in the refType map.  This would cause the test to fail even though it is doing what it is supposed to do.  It was verified that by adding a type which is in the refType map, the test fails.

Another test was added so that if HeritageTestModel2 is never found, it alerts the tester that this was a false positive.

### DIFF
--- a/packages/cli/src/metadataGeneration/typeResolver.ts
+++ b/packages/cli/src/metadataGeneration/typeResolver.ts
@@ -26,6 +26,7 @@ export class TypeResolver {
     private readonly parentNode?: ts.Node,
     private context: Context = {},
     private readonly referencer?: ts.TypeNode,
+    private readonly add: boolean = true,
   ) {}
 
   public static clearCache() {
@@ -419,7 +420,10 @@ export class TypeResolver {
 
     const referenceType = this.getReferenceType(typeReference);
 
-    this.current.AddReferenceType(referenceType);
+    if (this.add) {
+      this.current.AddReferenceType(referenceType);
+    }
+
     return referenceType;
   }
 
@@ -596,7 +600,7 @@ export class TypeResolver {
     };
   }
 
-  private getReferenceType(node: ts.TypeReferenceType): Tsoa.ReferenceType {
+  private getReferenceType(node: ts.TypeReferenceType, add = true): Tsoa.ReferenceType {
     let type: ts.EntityName;
     if (ts.isTypeReferenceNode(node)) {
       type = node.typeName;
@@ -647,7 +651,7 @@ export class TypeResolver {
 
       let referenceType: Tsoa.ReferenceType;
       if (ts.isTypeAliasDeclaration(declaration)) {
-        referenceType = this.getTypeAliasReference(declaration, name, node);
+        referenceType = this.getTypeAliasReference(declaration, name, node, add);
       } else if (ts.isEnumMember(declaration)) {
         referenceType = {
           dataType: 'refEnum',
@@ -670,7 +674,7 @@ export class TypeResolver {
     }
   }
 
-  private getTypeAliasReference(declaration: ts.TypeAliasDeclaration, name: string, referencer: ts.TypeReferenceType): Tsoa.ReferenceType {
+  private getTypeAliasReference(declaration: ts.TypeAliasDeclaration, name: string, referencer: ts.TypeReferenceType, add = true): Tsoa.ReferenceType {
     const example = this.getNodeExample(declaration);
 
     return {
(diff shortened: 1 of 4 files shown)


WoH (commented):
Notes:
- `add` is not a very descriptive name, `addToRefTypeMap` would be better.
- If we don't add models added by heritage clauses to ref types, we should make sure they don't end up in the spec. Aka add a small test case.
- I'm pretty sure we don't need them at runtime, since we combine the props into the underlying interface, so while we don't need an integration test, it wouldn't hurt either.

dpomerantz-jci:
> Notes:
> 
> * `add` is not a very descriptive name, `addToRefTypeMap` would be better.
> * If we don't add models added by heritage clauses to ref types, we should make sure they don't end up in the spec. Aka add a small test case.
> * I'm pretty sure we don't need them at runtime, since we combine the props into the underlying interface, so while we don't need an integration test, it wouldn't hurt either.

Parameter renamed as requested.

I've added a test, and hope this is sufficient.

dpomerantz-jci:
@WoH Tests were failing because lint changed things on commit.  I reverted the changes and added the disable comment.  Tests should now pass.  Please lmk if there are further changes required, or otherwise approve and merge.  Thanks.

WoH:
https://github.com/lukeautry/tsoa/releases/tag/v5.0.0-alpha.3

WoH:
Current main branch is the 5.x
Main reason was to drop support for older node versions and be safe for some potentially breaking fixes.

Issue #1340, opened 2022-11-10
Feature Request: Support for Serverless Backends


pclements12:
It would be nice if the RouteGenerator was a more generic interface that could support multiple backend designs besides a single Server template file. This would open up the TSOA library to be used by a lot more users and make the process for adding a new generator a little more clear.

## Sorting

- **I'm submitting a ...**

  - [ ] bug report
  - [x] feature request
  - [ ] support request

## Expected Behavior

It would be possible to supply a RouteGenerator module to the tsoa config that implements the RouteGenerator as an interface or an abstract base class. The RouteGenerator would only have a single method as an entry, but it would not be opinionated about the way that that backend would produce it's routes (or even other related resources such as CDK/Terraform/CloudFormation files to support that particular backend)

## Current Behavior

Currently the RouteGenerator pattern takes in a single template file and creates a single routes file.
For backends that require additional templating, for example a handler file per route, it's impossible to implement a template that satisfies this needs in the current architecture.

## Detailed Description

The determination of middleware type and template information could be moved into an implementing class of the generic RouteGenerator. When routes are generated, the RouteGenerator would be specified by the tsoa.json.

This could potentially still be backwards compatible by allowing a base route config that can be extended and provided as a constructor arg to the RouteGenerator.

A simple example of what this abstract base class might look like:

```ts

export abstract class AbstractRouteGenerator<Config extends ExtendedRoutesConfig> {
  constructor(protected readonly metadata: Tsoa.Metadata, protected readonly options: Config) {}

  /**
   * This is the entrypoint for a generator to create a custom set of routes
   */
  public abstract GenerateCustomRoutes(): Promise<void>;
```

In a serverless application hosted on AWS Lambda, this could be implemented to generate all of the handler files with the appropriate validations and meeting the models of the spec, and the CDK scripts for the API Gateway and Lambda stack that would be needed to support the HTTP/Route parts of the spec.

In this instance the route generator would need to produce a handler file for each route and possibly a "stack" file that creates and returns the CDK resources that would need to be deployed.

pclements12:
Also, I would be willing to contribute to this effort. I created a POC that would move the existing single file generator pattern to an implementation of the AbstractRouteGenerator and confirmed that all of the tests still pass with this approach.

PR #1342, opened 2022-11-11
feat(cli): Add support for generic custom route generators

pclements12:
### All Submissions:

**Closing issues**

This adds functionality similar to providing a custom middleware template, but extends that to allow the custom route generator to do whatever it wants. Custom middleware templates still work as-is, but the framework can be further extended via plugins with this generic approach.

**Test plan**

This PR re-organizes some of the existing code around route generation, but doesn't change how any of it functions. The existing unit and integration tests should all pass as-is without modification because of this.

Adds an example Custom route generator and executes it. Ensures that the output compiles to valid TS


pclements12:
Related to #1340

packages/cli/src/routeGeneration/routeGenerator.ts
```diff
@@ -1,86 +1,94 @@
...
 import { ExtendedRoutesConfig } from '../cli';
 import { Tsoa, TsoaRoute, assertNever } from '@tsoa/runtime';
-import { fsReadFile, fsWriteFile, fsExists } from '../utils/fs';
 import { isRefType } from '../utils/internalTypeGuards';
-import { normalisePath } from './../utils/pathUtils';
+import { convertBracesPathParams, normalisePath } from '../utils/pathUtils';
 
-export class RouteGenerator {
-  constructor(private readonly metadata: Tsoa.Metadata, private readonly options: ExtendedRoutesConfig) {}
+export abstract class AbstractRouteGenerator<Config extends ExtendedRoutesConfig> {
```
WoH, line 7: We should probably export this at the top level of both `@tsoa/cli` and `tsoa`

tests/fixtures/custom/custom-route-generator/serverlessRouteGenerator.ts
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  public async GenerateCustomRoutes() {
+    if (!fs.lstatSync(this.options.routesDir).isDirectory()) {
+      throw new Error(`routesDir should be a directory`);
+    }
+    console.log('Generating models');
+    await this.generateModels();
+    await this.generateRoutes();
+    await this.generateStack();
```
WoH, line 55: I am acking context here I think. What kind of stack?

tests/fixtures/custom/custom-route-generator/serverlessRouteGenerator.ts
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  async shouldWriteFile(fileName: string, content: string) {
```
WoH, line 58: I think extracting this also makes sense in the default (as a private method, not for overriding)

WoH:
I'm wondering if we may could to add your serverless target to the CI if it's possible?

tests/fixtures/custom/custom-route-generator/serverlessRouteGenerator.ts
pclements12: When using something like AWS Cloud Development Kit (CDK), you can programmatically generate your cloud resources for a serverless application. This would include things like an API Gateway and lambda function definitions for each of your tsoa controllers (or however you want to organize it). 

You would be able to use the tsoa metadata to setup each of these routes with the api gateway and tie them to the controller/handler code that was generated.

I didn't provide an example because I didn't want to add the extra dependencies that would be needed to generate it.

pclements12: I'm going to add an example to help here.

pclements12:
@WoH added an example aws cdk stack. I didn't want to before since it needs another test dependency to compile correctly, but I figured a working example was worthwhile. 

I didn't want to add this example in the main source since that would mean official "support" for this route generator in the same vein as express/hapi/koa, but I did want some real code that could be used as an example for anyone else who may want to build their own route generator for some other deployment style.

pclements12:
> I'm wondering if we could to add your serverless target to the CI if it's possible? Does it even make sense?

This might take some additional work. It's not exactly straightforward to run an AWS "serverless" style API locally.
It's possible, especially with the lambdas not having dependencies on other AWS resources, but would require bringing in the AWS SAM CLI and using the (beta) docker integration it has to run lambdas locally.

One of the benefits that I see of using TSOA actually, is that you could run your serverless app locally on express with just a tweak to the tsoa configuration, then point it to your custom configuration for deployment with CDK into AWS

tests/fixtures/custom/custom-route-generator/templates/api-stack.hbs
```diff
@@ -0,0 +1,61 @@
...
+        resource = this.resources[resourcePath];
+      } else {
+        // otherwise create a new resource for this path part
+        resource = parentResource.addResource(resourceName);
+      }
+      this.resources[resourcePath] = resource;
+      return resource;
+    }, api.root);
+  }
+}
```
WoH, line 61:
```suggestion
}

```

WoH (approved):

pclements12:
@WoH anything else you think this needs before merge?

Issue #1165, opened 2021-12-30
Stricter code checks


WoH:
Both eslint and typescript-eslint have a bunch of useful rules we had to turn off when switching from tslint although they very helpful. An incomplete list, mostly geared towards the tsoa, runtime and cli packages:

- [ ] Strict tsconfig in all packages
- [ ] `@typescript-eslint/explicit-module-boundary-types`
- [ ] `@typescript-eslint/member-delimiter-style`
- [ ] `@typescript-eslint/no-explicit-any`
- [ ] `@typescript-eslint/no-floating-promises`
- [ ]  `@typescript-eslint/no-non-null-assertion`
- [ ] `@typescript-eslint/no-parameter-properties`
- [ ] `@typescript-eslint/no-unsafe-assignment`
- [ ] `@typescript-eslint/no-unsafe-call`
- [ ]  `@typescript-eslint/no-unsafe-argument`
- [ ] `@typescript-eslint/no-unsafe-member-access`
- [ ] `@typescript-eslint/no-unsafe-return`
- [ ] `@typescript-eslint/no-unused-expressions`
- [ ] `'@typescript-eslint/no-unused-vars` (prefix would be fine)
- [ ] `arrow-parens`
- [ ] `max-classes-per-file`
- [ ] `no-array-constructor`
- [ ] `no-empty-function`
- [ ] `no-fallthrough`
- [ ] `no-invalid-this`
- [ ] `no-shadow`
- [ ] `require-await`
- [ ]  `valid-typeof`

In case anyone wants to add a PR, even just for a subset of issues one of the rules would report, please help out.

PR #1433, opened 2023-05-28
fix: enable compilerOptions.strict for all packages

AudunWA:
### All Submissions:

**Closing issues**
Fixes parts of #1165 ("Strict tsconfig in all packages" checkbox)

### Own comments
This adds `strict: true` to all `tsconfig.json` files, and removes redundant options that `strict` enables.
All resulting TS errors have been fixed, but some fixes might still hide already existing issues.
**Note:** I kept `noImplicitAny: false` for tests, but I can remove it if we want full strictness for test code as well.


packages/cli/src/metadataGeneration/extension.ts
```diff
@@ -39,18 +44,19 @@ export function getExtensionsFromJSDocComments(comments: string[]): Tsoa.Extensi
...
       });
     }
   });
 
   return extensions;
 }
 
-function validateExtensionKey(key: string) {
-  if (key.indexOf('x-') !== 0) {
+function validateExtensionKey(key: string): key is `x-${string}` {
```
WoH, line 57: I would prefer 
```ts
asserts key is `x-${string}`
```
This is more intuitive when throwing and avoids the throwing `if/then/else`.

Also gives rise to renaming this function to `assertValidExtensionKey`

packages/cli/src/swagger/specGenerator3.ts
```diff
@@ -605,14 +598,14 @@ export class SpecGenerator3 extends SpecGenerator {
...
   private groupEnums(types: Array<Swagger.Schema | Swagger.BaseSchema>) {
     const returnTypes: Array<Swagger.Schema | Swagger.BaseSchema> = [];
-    const enumValuesByType = {};
+    const enumValuesByType: Record<string, Record<string, boolean | string | number | null>> = {};
     for (const type of types) {
       if (type.enum && type.type) {
         for (const enumValue of type.enum) {
           if (!enumValuesByType[type.type]) {
-            enumValuesByType[type.type] = [];
+            enumValuesByType[type.type] = {};
```
WoH, line 606: Can we add a test here? It seems like this should be covered.

WoH (commented):
Minor suggestions, looks great overall

packages/cli/src/metadataGeneration/extension.ts
AudunWA: Cool, didn't know about that TS feature – it definitely looks better!

packages/cli/src/swagger/specGenerator3.ts
AudunWA: The result of the `groupEnums` should be the same, as JS allows setting string keys even on arrays. Thus, the array behaves as an object:
image

How would I test this change, any tips?

WoH: I see why this change makes sense, should've always been an `{}`.

WoH (approved):

Issue #1307, opened 2022-09-08
Missing ECMAScript modules (ESM) support in both controller search and routes generation


yashag-art:
## Sorting

- **I'm submitting a ...**

  - [x] bug report
  - [ ] feature request
  - [x] support request

- I confirm that I
  - [x] used the [search](https://github.com/lukeautry/tsoa/search?type=Issues) to make sure that a similar issue hasn't already been submit

## Expected Behavior

In `tsoa.json` or the programmatic API, when providing the following configuration
```ts
{
  controllerPathGlobs: ["src/**/*.controller.mts"]
}
```
it should search all files ending with `.controller.mts`.

Also when generating `routes` with `tsoa` an option to generates a `routes.mts` files with correct imports of controllers should be provided.

Full `tsoa.json` file:
```json
{
  "entryFile": "./src/index.mts",
  "noImplicitAdditionalProperties": "throw-on-extras",
  "controllerPathGlobs": ["src/**/*.controller.mts"],
  "spec": {
    "specVersion": 3,
    "outputDirectory": "./src/docs/api",
    "entryFile": "./src/index.mts"
  },
  "routes": {
    "routesDir": "src/routes",
    "middleware": "express",
    "esm": true
  }
}
```

## Current Behavior

The following error pops up every time I run `tsoa spec`:
```shell
GenerateMetadataError: [src/**/*.controller.mts] globs found 0 controllers.
```
If I omit the `controllerPathGlobs` field in `tsoa.json` it finds all the decorators in my controller files, but I have no control over where the search occurs (and I tried multiple variations of the glob). Through the programmatic API it doesn't work (I get this error) either way (with the option provided or without it).

And when trying to generate routes with `tsoa routes` the TypeScript file created is:
1. A `.ts` file and not `.mts` as all my other files (there is no option to influence this)
2. contains a broken controller import:
```ts
import { SystemController } from './../system/system.controller..js';
```
Which not only has a broken syntax, but understandably generates this TypeScript error:
`TS2307: Cannot find module './../system/system.controller..js' or its corresponding type declarations.` 

## Steps to Reproduce

1. Create a `tsoa.json` file with a `controllerPathGlobs` field searching for `.mts` files
2. Create a `.mts` file with `tsoa` decorators with the scope `controllerPathGlobs` searches for
3. Run `tsoa spec`
4. Now that you receive the error, omit `controllerPathGlobs` and run `tsoa spec` successfully
[...]

PR #1455, opened 2023-07-27
fix(ESM): support mts and cts files

Kampfmoehre:
### All Submissions:

- [x] This PR is associated with an existing issue? #1307 

**Closing issues**
#1307 

Haven't had the ability to also test `.cts` files but I think this should be straight forward.

**Test plan**

I added two unit tests for the route generator method `getRelativeImportPath` that replaces the extension in controller files.

### DIFF
--- a/packages/cli/src/metadataGeneration/metadataGenerator.ts
+++ b/packages/cli/src/metadataGeneration/metadataGenerator.ts
@@ -21,7 +21,7 @@ export class MetadataGenerator {
     controllers?: string[],
     private readonly rootSecurity: Tsoa.Security[] = [],
     public readonly defaultNumberType: NonNullable<Config['defaultNumberType']> = 'double',
-    esm: boolean = false,
+    esm = false,
   ) {
     TypeResolver.clearCache();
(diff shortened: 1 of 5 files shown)


packages/cli/src/metadataGeneration/metadataGenerator.ts
```diff
@@ -42,8 +43,8 @@ export class MetadataGenerator {
     };
   }
 
-  private setProgramToDynamicControllersFiles(controllers: string[]) {
-    const allGlobFiles = importClassesFromDirectories(controllers);
+  private setProgramToDynamicControllersFiles(controllers: string[], esm: boolean) {
+    const allGlobFiles = importClassesFromDirectories(controllers, esm ? ['.mts', '.ts', '.cts']: ['.ts']);
```
WoH, line 47: Does it hurt to add these in case we are CJS?

Kampfmoehre: I don't think it hurts. As far as I know you have 3 possible ways to work with ESM:
- keep `ts` files
- use `mts" files when you want to output specifically ESM files
- use `cts` files when you also want to output CommonJS

For our use case we only need `mts`/`mjs` files, but the logic is the same for the others as far as I know.

PR #1456, opened 2023-07-27
Correctly generate types for Record<string, T>

### LINKED ISSUE #1407: How to represent Records
I have a model that looks something like this 

```
type Tenancy = string

enum Role {
  User, 
  Admin
}

type User {
  id: string
  roles: Record<Tenancy, Role[]>
}
```

The generated swagger schema looks like this:

```
{
  "Record_TenancySlug.Role-Array_": {
    "properties": {},
    "type": "object",
    "description": "Construct a type with a set of properties K of type T"
  },
  "User": {
    "roles": {
      "$ref": "#/components/schemas/Record_TenancySlug.Role-Array_"
    }
  }
}
```

As you can see the typing is lost and all we can deduce is that the roles property is an object.

Is there someway to generate stronger typings?

Thanks

- **I'm submitting a ...**

  - [ ] bug report
  - [ ] feature request
  - [x] support request

- I confirm that I

ryanisaacg:
When encountering a Record whose key is string, tsoa would previously generate a blank object. This change allows the type resolver to recognize the record and generate its value type properly.

I would love guidance on where to add some tests for this PR! The test suite is a little opaque to me.

### All Submissions:

**Closing issues**

Closes #1407

**Test plan**

### DIFF
--- a/packages/cli/src/metadataGeneration/typeResolver.ts
+++ b/packages/cli/src/metadataGeneration/typeResolver.ts
@@ -422,7 +422,7 @@ export class TypeResolver {
 
       if (typeReference.typeName.text === 'Record' && typeReference.typeArguments?.length === 2) {
         const indexType = new TypeResolver(typeReference.typeArguments[0], this.current, this.parentNode, this.context).resolve();
-        if (indexType.dataType === 'string') {
+        if (indexType.dataType === 'string' || indexType.dataType === 'integer' || indexType.dataType === 'double') {
           const valueType = new TypeResolver(typeReference.typeArguments[1], this.current, this.parentNode, this.context).resolve();
 
           const nestedObjectMetaType: Tsoa.NestedObjectLiteralType = {
--- a/packages/cli/src/metadataGeneration/typeResolver.ts
+++ b/packages/cli/src/metadataGeneration/typeResolver.ts
@@ -420,6 +420,20 @@ export class TypeResolver {
         return stringMetaType;
       }
 
+      if (typeReference.typeName.text === 'Record' && typeReference.typeArguments?.length === 2) {
+        const indexType = new TypeResolver(typeReference.typeArguments[0], this.current, this.parentNode, this.context).resolve();
+        if (indexType.dataType === 'string') {
+          const valueType = new TypeResolver(typeReference.typeArguments[1], this.current, this.parentNode, this.context).resolve();
+
+          const nestedObjectMetaType: Tsoa.NestedObjectLiteralType = {
+            additionalProperties: valueType,
+            dataType: 'nestedObjectLiteral',
+            properties: [],
+          };
+          return nestedObjectMetaType;
+        }
+      }
+
       if (this.context[typeReference.typeName.text]) {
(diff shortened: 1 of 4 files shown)


WoH:
Best way to show this is to add a property to the Test Model here:
https://github.com/lukeautry/tsoa/blob/master/tests/fixtures/testModel.ts

And do assertions for the type here:
https://github.com/lukeautry/tsoa/blob/master/tests/unit/swagger/definitionsGeneration/definitions.spec.ts#L1492

and here:
https://github.com/lukeautry/tsoa/blob/master/tests/unit/swagger/schemaDetails3.spec.ts#L2547

packages/cli/src/metadataGeneration/typeResolver.ts
```diff
@@ -420,6 +420,20 @@ export class TypeResolver {
         return stringMetaType;
       }
 
+      if (typeReference.typeName.text === 'Record' && typeReference.typeArguments?.length === 2) {
+        const indexType = new TypeResolver(typeReference.typeArguments[0], this.current, this.parentNode, this.context).resolve();
+        if (indexType.dataType === 'string') {
```
WoH, line 425: Why only for string? We should be able to handle numbers, boolean, string enums, number enums etc.?

packages/cli/src/metadataGeneration/typeResolver.ts
```diff
@@ -420,6 +420,20 @@ export class TypeResolver {
...
+          const valueType = new TypeResolver(typeReference.typeArguments[1], this.current, this.parentNode, this.context).resolve();
```
WoH, line 426: In case we are passing from parent (GenericModel<T> has items: Record<string, T>) can you handle these without amending the context? (IIRC that should be created on processing the `GenericModel<T>`, but it would also be a good test probably.

ryanisaacg:
Thanks for the pointers! I've now added tests to cover my changes.

packages/cli/src/metadataGeneration/typeResolver.ts
ryanisaacg: thanks for the number callout! I've added support here for numbers and test cases to verify

- number enums and string enums are handled separately (I'm not sure where, but as you can see from the test suite they are handled correctly). They actually produce fairly different OpenAPI results, because of the way swagger handles dictionaries
- `boolean` is not a valid `Record` key type; the type for `Record`'s key is `string | number | symbol`. `symbol` isn't supported by tsoa it seems (which makes sense, because it doesn't belong in a public-facing API)

packages/cli/src/metadataGeneration/typeResolver.ts
ryanisaacg: How would I go about doing that? Should I be creating a new temporary context?

I'm not familiar with the codebase so I'm not sure what the context is here, or what purpose it serves, so I just followed the structure of other cases in this method.

WoH: This should be handled for you at this point already.

ryanisaacg: Sorry, I'm not sure what you mean.

WoH (approved):

PR #1468, opened 2023-08-17
fix: Don't URI encode schema keys when using $ref

### LINKED ISSUE #1461: Usage of `$` in type names leads to invalid OpenAPI specs
I created a sample repo to explain what's happening:

If a controller is referencing a type that looks like `HelloWorld`, then the resulting OpenAPI spec will create a a schema that is invalid (based on https://apitools.dev/swagger-parser/online/ and a few other tools I've been using, including https://github.com/stoplightio/prism)

```ts
export namespace $Enums {
  export const Test = {
    hello: "world",
    foo: "bar",
  };
  export type Test = (typeof $Enums.Test)[keyof typeof $Enums.Test];
}

export interface HelloWorld {
  hello: $Enums.Test;
}
```

https://github.com/alandotcom/tsoa-bug/blob/56a15eeb600fd6526658417d464719b564fa777b/src/types.ts

## Sorting

- **I'm submitting a ...**

  - [X] bug report
  - [ ] feature request
  - [ ] support request

- I confirm that I

## Expected Behavior

I believe that `%24` should be `$` on this line:

https://github.com/alandotcom/tsoa-bug/blob/56a15eeb600fd6526658417d464719b564fa777b/src/swagger.json#L9

## Current Behavior

```yaml

```

```ts

```

## Steps to Reproduce

Example repo here https://github.com/alandotcom/tsoa-bug/tree/main

## Context (Environment)

Version of the library: latest
Version of NodeJS: 18

- Confirm you were using yarn not npm: [X] (used pnpm, but unrelated to the problem)

## Detailed Description

Changing the generated swagger spec to be valid

## Breaking change?

I don't think this would be a breaking change

LucianBuzzo:
This fixes an isssue where using `$` in type names leads to invalid OpenAPI specs. This would happen because the `refName` property that is used as the object key for refs and to create `$ref` attributes is URI encoded on creation, when it only needs to be URI encoded when used in a `$ref` attribute.

Fixes #1461

**Test plan**

I had a hard time configuring a unit test to check this behaviour, if anyone can provide some guidance on how to do it please let me know, as the test cases in this repo are hard to follow.

### DIFF
--- a/packages/cli/src/metadataGeneration/metadataGenerator.ts
+++ b/packages/cli/src/metadataGeneration/metadataGenerator.ts
@@ -44,7 +44,7 @@ export class MetadataGenerator {
   }
 
   private setProgramToDynamicControllersFiles(controllers: string[], esm: boolean) {
-    const allGlobFiles = importClassesFromDirectories(controllers, esm ? ['.mts', '.ts', '.cts']: ['.ts']);
+    const allGlobFiles = importClassesFromDirectories(controllers, esm ? ['.mts', '.ts', '.cts'] : ['.ts']);
     if (allGlobFiles.length === 0) {
       throw new GenerateMetadataError(`[${controllers.join(', ')}] globs found 0 controllers.`);
     }
@@ -216,7 +216,7 @@ export class MetadataGenerator {
     if (!referenceType.refName) {
       return;
     }
-    this.referenceTypeMap[referenceType.refName] = referenceType;
+    this.referenceTypeMap[decodeURIComponent(referenceType.refName)] = referenceType;
   }
 
   public GetReferenceType(refName: string) {
--- a/packages/cli/src/metadataGeneration/typeResolver.ts
+++ b/packages/cli/src/metadataGeneration/typeResolver.ts
@@ -763,21 +763,19 @@ export class TypeResolver {
   }
 
   private getRefTypeName(name: string): string {
-    return encodeURIComponent(
-      name
-        .replace(/<|>/g, '_')
-        .replace(/\s+/g, '')
-        .replace(/,/g, '.')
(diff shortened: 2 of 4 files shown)


WoH:
Does this only affect the OpenAPI v2 generator? I think we should be consistent and also adapt the OpenAPI v3 Spec Generator.

WoH:
That said, I think https://github.com/lukeautry/tsoa/pull/1469 would also address this.

I'm not sure if you can URI Encode the ref value, but use the unencoded name as the components/definition key, is that specified somewhere?

LucianBuzzo:
@WoH I updated my PR to also fix the problem for the v3 generator. AFAICT, what happens is that when the `$ref` value is resolved, it decodes the URI, resulting in `%24` being decoded to `$` - however, because we use the encoded value as a key, it can't find it. The resolution lookup would be for `$variable`, but what's in the schema is `%24variable`.
By only URI encoding the `$ref` value, we avoid this particular issue, without resorting to special case escaping of the keys.
URI encoding only the `$ref` URI value is the canonically correct approach as far as I can see!

WoH:
LGTM

amrbahaa360:
Is the PR merged to version 5.1.1?
The issue still exists and has not been solved.

LucianBuzzo:
@amrbahaa360 It looks like v5.1.1 was released in February, which won't include this change. A new version needs to be released. @WoH would you be able to cut a new version that includes this fix? v5.1.2 perhaps?

WoH:
https://github.com/lukeautry/tsoa/releases/tag/v6.0.0-rc.4

LucianBuzzo:
Awesome thanks @WoH !

PR #1469, opened 2023-08-20
Escape backtick and template variable from type name

MacFJA:
### All Submissions:

---

This PR solve the issue with Swagger that can't handle `$` schema name.

My use case is the following:
```ts
type MyModel = {
  normalProp: string;
  [key: `variable${number}`]: string;
}
```

TSOA can't use the type `MyModal` which is understandable as one of the key is dynamic.
My approach what to create a TSOA specific type like this:
```ts
import { MyModal as SourceMyModel } from "./types";
type MyModel = Omit<SourceMyModel, `variable${number}`> & { variable1: string, variable2: string };
```
With this, TSOA is able to generate the OpenApi file,
but the generated schemas are:
```json
"Pick_SourceMyModel.Exclude_keyofSourceMyModel.%60variable%24_number_%60__": {
	"properties": {
		"normalProp": {
			"type": "string"
		}
	},
	"required": [
		"normalProp"
	],
	"type": "object",
	"description": "From T, pick a set of properties whose keys are in the union K"
},
"Omit_SourceMyModel.%60variable%24_number_%60_": {
	"$ref": "#/components/schemas/Pick_SourceMyModel.Exclude_keyofSourceMyModel.%60variable%24_number_%60__",
	"description": "Construct a type with the properties of T except for those in type K."
},
"MyModel": {
	"allOf": [
		{
			"$ref": "#/components/schemas/Omit_SourceMyModel.%60variable%24_number_%60_"
		},
		{
			"properties": {
				"variable2": {
					"type": "string"
				},
				"variable1": {
					"type": "string"
				}
			},
			"type": "object"
		}
	]
}
```

With my PR the output will be:
```json
"Pick_SourceMyModel.Exclude_keyofSourceMyModel.variable_number__": {
	"properties": {
		"normalProp": {
			"type": "string"
		}
	},
	"required": [
		"normalProp"
	],
	"type": "object",
	"description": "From T, pick a set of properties whose keys are in the union K"
},
"Omit_SourceMyModel.variable_number_": {
	"$ref": "#/components/schemas/Pick_SourceMyModel.Exclude_keyofSourceMyModel.variable_number__",
	"description": "Construct a type with the properties of T except for those in type K."
},
"MyModel": {
	"allOf": [
		{
			"$ref": "#/components/schemas/Omit_SourceMyModel.variable_number_"
		},
		{
			"properties": {
				"variable2": {
					"type": "string"
				},
				"variable1": {
					"type": "string"
				}
			},
			"type": "object"
		}
	]
}
```

### DIFF
--- a/packages/cli/src/metadataGeneration/typeResolver.ts
+++ b/packages/cli/src/metadataGeneration/typeResolver.ts
@@ -770,9 +770,11 @@ export class TypeResolver {
         .replace(/,/g, '.')
         .replace(/'([^']*)'/g, '$1')
         .replace(/"([^"]*)"/g, '$1')
+        .replace(/`([^`]*)`/g, '$1')
         .replace(/&/g, '-and-')
         .replace(/\|/g, '-or-')
         .replace(/\[\]/g, '-Array')
+        .replace(/\$\{([^}]+)\}/g, '_$1') // ${variable} -> _variable
         .replace(/{|}/g, '_') // SuccessResponse_{indexesCreated-number}_ -> SuccessResponse__indexesCreated-number__
         .replace(/([a-z]+):([a-z]+)/gi, '$1-$2') // SuccessResponse_indexesCreated:number_ -> SuccessResponse_indexesCreated-number_
         .replace(/;/g, '--')
--- a/tests/fixtures/controllers/typeInferenceController.ts
+++ b/tests/fixtures/controllers/typeInferenceController.ts
@@ -53,4 +53,11 @@ export class TypeInferenceController {
       demo21: '',
     };
   }
+
+  @Get('escape-string-literal-type')
+  public escapeStringLiteralType(): Omit<TruncationTestModel, `demo${number}`> {
+    return {
+      d: '',
+    };
+  }
 }
--- a/tests/unit/swagger/definitionsGeneration/metadata.spec.ts
+++ b/tests/unit/swagger/definitionsGeneration/metadata.spec.ts
@@ -1006,6 +1006,22 @@ describe('Metadata generation', () => {
         'demo21',
       ]);
     });
+
+    it('should generate escapeStringLiteralType method', () => {
+      const method = controller.methods.find(method => method.name === 'escapeStringLiteralType');
+      if (!method) {
+        throw new Error('Method escapeStringLiteralType not defined!');
+      }
+
+      expect(method.method).to.equal('get');
+      expect(method.path).to.equal('escape-string-literal-type');
+      const [response] = method.responses;
+      expect(response.schema?.dataType).to.eq('refAlias');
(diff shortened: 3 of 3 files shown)


WoH:
Looks reasonable, can you please add the example you provided as a test?

MacFJA:
@WoH Sorry for the delay.
I added the test, I hope I put it in the right place

packages/cli/src/metadataGeneration/typeResolver.ts
```diff
@@ -770,9 +770,11 @@ export class TypeResolver {
         .replace(/,/g, '.')
         .replace(/'([^']*)'/g, '$1')
         .replace(/"([^"]*)"/g, '$1')
+        .replace(/`([^`]*)`/g, '$1')
```
WoH, line 773: This concern is valid for both this and the other escape, but I think we can construct scenarios where this could lead to a clash (2 different references ending up with the same name), right?

MacFJA: Yes it could, and whenever char are replaced or removed.

l TS Type | Escaped value | Data group |
|---|---|---|
|`Omit<MyType, "_foo">`|`Omit_MyType._foo_`|The type `MyType` minus the key `_foo` (1)|
|`Omit<MyType, '_foo'>`|`Omit_MyType._foo_`|The type `MyType` minus the key `_foo` (1)|
|``Omit<MyType, `_foo`>``|`Omit_MyType._foo_`|The type `MyType` minus the key `_foo` (1)|
|``Omit<MyType, `${foo}`>``|`Omit_MyType._foo_`|The type `MyType` minus the key of type `foo` (2)|
|``Omit<MyType, `${fo}o`>``|`Omit_MyType._foo_`|The type `MyType` minus the key of (type `fo` + text `o`)  (3)|
|``Omit<MyType, `${"foo"}`>``|`Omit_MyType._foo_`|The type `MyType` minus the key `foo` (4)|
|_many more possibilities_|||

For the first 3, they are the same data, so it's not a big deal.

---

There is maybe another case (I don't tested it yet): classes, interfaces, types, enums are not "unique".
In TSOA, `refType` don't mention their source file, so:
```ts
// fileA.ts
import { MyType } from "./somewhere/in/the/project"
```
```ts
// fileB.ts
import { MyType } from "./in/another/place"
```
will be declared as the same schema in the OpenAPI file (to be confirmed), and so create a clash

WoH: Can we find a way to rewrite that so it does not clash with other, but different type names?

WoH:
> Is the PR merged to version 5.1.1? The issue still exists and has not been solved.

Would you like to improve on this PR and address the points I raised during review? Remember to add  tests.

Issue #1474, opened 2023-09-02
Bug: Type Inference for Nullable Union Types Inaccurate; Suggest Enabling Strict Mode


antoniel:
## Sorting

- **I'm submitting a ...**

  - [X] bug report
  - [ ] feature request
  - [ ] support request

- I confirm that I
  - [x] used the [search](https://github.com/lukeautry/tsoa/search?type=Issues) to make sure that a similar issue hasn't already been submit

## Expected Behavior

Given the follow typescript code, I expect that tsoa infer the the return type as ´string | null´.

```ts
import { Get, Route } from '@tsoa/runtime';

@Route('inference')
export class TypeInferenceWithNullableValueController {
  @Get('keys-inference-with-nullable')
  public multiKeysInterfaceInferenceWithNullable() {
    const maybeString: string | null = Math.random() > 0.5 ? 'foo' : null;
    return maybeString
  }
}
```

## Current Behavior

Tsoa should ideally infer its return type as string | null. However, it currently infers it solely as string, thereby omitting the null part of the union.. This behavior could lead to potential misunderstandings regarding the method's capabilities and can affect the OpenAPI documentation negatively.

## Possible Solution

One possible solution to this problem is to set the typescript program that is created by MetadataGenerator as strict by default, it can be done by compilerOptions. It will improve the inference of typescript in this case and in many others without any cons as I can see. 

## Steps to Reproduce

1. Create a controller that returns `string | null`
2. Run tsoa on this controller to generate openApi
3. Assets that the route was created with the response type as string, and not a union of string and null

## Context (Environment)

Version of the library: 5.1.1
Version of NodeJS: 18.12.0

- Confirm you were using yarn not npm: [X]

## Detailed Description

Upon a preliminary investigation, it appears that the TypeScript program created by the MetadataGenerator is not configured to operate in strict mode. Strict mode in TypeScript enhances type checking and adds certain compile-time checks that would likely resolve this issue.

## Breaking change?

It may change the output of existing definition, this change will increase the accuracy of the schemas generated by tsoa. This change will not reduce in any way the scope of existing OpenApi Schemas.

PR #1475, opened 2023-09-02
Enhance Nullable Union Type Inference through Strict Mode

antoniel:
#### Summary

This PR addresses issue #1474, aiming to improve the type inference mechanism for nullable union types in tsoa. By enabling TypeScript's strict mode, we significantly enhance the accuracy of type definitions and OpenAPI Schemas generated by the library.

#### Changes

- Modified TypeScript program's `compilerOptions` to include `strict: true`.
- Adjusted relevant unit tests to validate the improved type inference.

#### Impact

- This change may modify the output of existing API definitions. However, it enhances the schema's accuracy without limiting the current scope.
  
#### How to Test

1. Pull the changes and install updated dependencies.
2. Run the existing unit tests to validate that the change is working as expected.
3. Manually create a route using a nullable union type and confirm that the generated OpenAPI schema is accurate.

#### Context

- Library Version: 5.1.1
- Node Version: 18.12.0
- Package Manager: Yarn

---

### DIFF
--- a/packages/cli/src/metadataGeneration/metadataGenerator.ts
+++ b/packages/cli/src/metadataGeneration/metadataGenerator.ts
@@ -24,7 +24,7 @@ export class MetadataGenerator {
     esm = false,
   ) {
     TypeResolver.clearCache();
-    this.program = controllers ? this.setProgramToDynamicControllersFiles(controllers, esm) : createProgram([entryFile], compilerOptions || {});
+    this.program = controllers ? this.setProgramToDynamicControllersFiles(controllers, esm) : createProgram([entryFile], { strict: true, ...compilerOptions } || {});
     this.typeChecker = this.program.getTypeChecker();
   }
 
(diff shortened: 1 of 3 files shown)


WoH:
Does setting it in the tsconfig of your tsoa's config file not work?
See https://github.com/lukeautry/tsoa/blob/master/packages/runtime/src/config.ts#L41C14-L41C14

I am worried this might have unintended consequences when we run this on strict: false codebases and break typing completely. That said, maybe strict: true is by now common enough to test it.

antoniel:
Thank you for your suggestion! I did try setting it in the tsoa's config file, and it seems to work well. I agree that ensuring compatibility is essential. However, in scenarios where the OpenAPI file serves as a blueprint to generate validation schemas using tools like Zod, maintaining accurate type representation is critical. Having just string instead of string | null could lead to serious issues.

antoniel:
Maybe we can add explicitly in the documentation that setting the tsoa as strict in their config files is highly recommended, even if the tsconfig is not strict.

WoH:
That would be highly appreciated.

Issue #1490, opened 2023-10-04
Extended interface doesn't produce an `allOf` type


Mearman:
Hopefully self explanatory. The `ExtendedModelInterface` path doesn't produce a schema that references the subtypes
```typescript
export interface Model {
	a: string
}
export interface ExtendedModelInterface extends Model {
	b: string
}
export type ExtendedModelType = Model & ExtendedModelInterface;


@Route("model")
export class ModelController  {
	@Get("model")
	public async getBaseModel(): Promise<Model> {
		return {} as Model;
	}
	@Get("extended_model_interface")
	public async getExtendedModelInterface(): Promise<ExtendedModelInterface> {
		return {} as ExtendedModelInterface;
	}
	@Get("extemded_model_type")
	public async getExtendedModelType(): Promise<ExtendedModelType> {
		return {} as ExtendedModelType;
	}
}
```

```yml
components:
  schemas:
    Model:
      properties:
        a:
          type: string
      required:
        - a
      type: object
      additionalProperties: false
    ExtendedModelInterface:
      properties:
        a:
          type: string
        b:
          type: string
      required:
        - a
        - b
      type: object
      additionalProperties: false
    ExtendedModelType:
      allOf:
        - $ref: '#/components/schemas/Model'
        - $ref: '#/components/schemas/ExtendedModelInterface'
```

this is what I would expect
```yml
components:
  schemas:
    Model:
      properties:
        a:
          type: string
      required:
        - a
      type: object
      additionalProperties: false
    ExtendedModelInterface:
      allOf:
        - $ref: '#/components/schemas/Model'
        - properties:
            b:
              type: string
          required:
            - b
          type: object
          additionalProperties: false
    ExtendedModelType:
      allOf:
        - $ref: '#/components/schemas/Model'
        - $ref: '#/components/schemas/ExtendedModelInterface'

```

Mearman:
I had a look at https://github.com/lukeautry/tsoa/pull/400 and it looks like this is not a case that is tested

WoH:
OpenAPI does not support this so we came up with out own DIY inheritance.

WoH:
I would be fine with delaying that from before the metadata generation to the spec generators and have OpenAPI 3 use the "built in" way. Would you like to submit a PR for that?

Mearman:
Ah do you mean having both behaviours, based on whether targeting OpenAPI V2 or V3?

PR #1498, opened 2023-10-19
fix: Referencer, Name conflicts, Mapped unions, etc.

### LINKED ISSUE #477: Namespaced cross-references not recognized without parent namespace
When referencing an interface in the same namespace, tsoa does not recognize the referenced namespace without the parent namespace attached

## Sorting

* **I'm submitting a ...**
  - [ X] bug report

* I confirm that I
  - [ X] used the [search](https://github.com/lukeautry/tsoa/search?type=Issues) to make sure that a similar issue hasn't already been submit

## Expected Behavior

I should be able to import a Namespace that has references to other types in the same namespace.

```yaml

```

```ts

export Namespace Shape {
  export interface Dimension {
   
   }

   export interface Square {
        length: Dimension
    }
}

```

```
import 'Shape' from './models/types';

...Promise<Shape.Square>
```

## Current Behavior

Currently returns:

``` 
UnhandledPromiseRejectionWarning: Error: No matching model found for referenced type Dimension
```

In order to get it to work, we must do:
```ts

export Namespace Shape {
   export interface Square {
        length: Shape.Dimension
    }
}

```

```
import 'Shape' from './models/types';

...Promise<Shape.Square>
```

## Steps to Reproduce

1. Add namespace to model
2. Add a cross-reference to a type in the namespaced model
3. Run route generator

## Context (Environment)
Version of the library: 2.51
Version of NodeJS: v12.8.0
* Confirm you were using yarn not npm: [X ]

### LINKED ISSUE #623: Default tag only supports strings
## Sorting

* **I'm submitting a ...**
  - [ ] bug report
  - [x] feature request
  - [ ] support request

* I confirm that I

## Expected Behavior
default tag should parse input so that number examples can be provided
```ts
/**
 * @default 10
 */
export type PageLimit = number
```
```json
"PageLimit": {
  "default": 10
}
```

## Current Behavior
Currently all default tags are treated as strings
```ts
/**
 * @default 10
 */
export type PageLimit = number
```
```json
"PageLimit": {
  "default": "10"
}
```
## Possible Solution
A simple solution would be to use the same JSON parsing logic used when getting the node example. This would allow both numeric and object defaults while still keeping support for string defaults.

## Breaking change?
Because this would change parsing logic for existing usages for the default tag, it would be a breaking change. Current string defaults would now require quotes

### LINKED ISSUE #681: Deeply nested wide mapped types on enums do not resolve (sometimes)

### LINKED ISSUE #1191: Omitted partials lose their partial property status
## Sorting

- **I'm submitting a ...**

  - [x] bug report
  - [ ] feature request
  - [ ] support request

- I confirm that I

## Expected Behavior

When omitting a property from a type which extends a partial of a type, I would expect the partial properties to still be optional (see `Pick_IUpdatableTest.Exclude_keyofIUpdatableTest.id__`).

```yaml
components:
    examples: {}
    headers: {}
    parameters: {}
    requestBodies: {}
    responses: {}
    schemas:
        IUpdatableTest:
            properties:
                prop1:
                    type: string
                prop2:
                    type: boolean
                updatedOn:
                    type: string
                id:
                    type: string
            required:
                - updatedOn
                - id
            type: object
            additionalProperties: false
        Pick_IUpdatableTest.Exclude_keyofIUpdatableTest.id__:
            properties:
                prop1:
                    type: string
                prop2:
                    type: boolean
                updatedOn:
                    type: string
            required:
                - updatedOn
            type: object
            description: 'From T, pick a set of properties whose keys are in the union K'
        Omit_IUpdatableTest.id_:
            $ref: '#/components/schemas/Pick_IUpdatableTest.Exclude_keyofIUpdatableTest.id__'
            description: 'Construct a type with the properties of T except for those in type K.'
    securitySchemes: {}
info:
    title: TSOA
    version: 1.0.0
    description: 'Build swagger-compliant REST APIs using TypeScript and Node'
    license:
        name: MIT
    contact:
        name: unknown
openapi: 3.0.0
paths:
    '/{id}':
        patch:
            operationId: Update
            responses:
                '200':
                    description: Ok
                    content:
                        application/json:
                            schema:
                                $ref: '#/components/schemas/IUpdatableTest'
            security: []
            parameters:
                -
                    in: path
                    name: id
                    required: true
                    schema:
                        type: string
            requestBody:
                required: true
                content:
                    application/json:
                        schema:
                            $ref: '#/components/schemas/Omit_IUpdatableTest.id_'
servers:
    -
        url: /
```

## Current Behavior

When omitting a property from a type which extends a partial of a type, the properties that should be optional are now required (see `Pick_IUpdatableTest.Exclude_keyofIUpdatableTest.id__`).

```yaml
components:
    examples: {}
    headers: {}
    parameters: {}
    requestBodies: {}
    responses: {}
    schemas:
        IUpdatableTest:
            properties:
                prop1:
                    type: string
                prop2:
                    type: boolean
                updatedOn:
                    type: string
                id:
                    type: string
            required:
                - updatedOn
                - id
            type: object
            additionalProperties: false
        Pick_IUpdatableTest.Exclude_keyofIUpdatableTest.id__:
            properties:
                prop1:
                    type: string
                prop2:
                    type: boolean
                updatedOn:
                    type: string
            required:
                - prop1
                - prop2
                - updatedOn
            type: object
            description: 'From T, pick a set of properties whose keys are in the union K'
        Omit_IUpdatableTest.id_:
            $ref: '#/components/schemas/Pick_IUpdatableTest.Exclude_keyofIUpdatableTest.id__'
            description: 'Construct a type with the properties of T except for those in type K.'
    securitySchemes: {}
info:
    title: TSOA
    version: 1.0.0
    description: 'Build swagger-compliant REST APIs using TypeScript and Node'
    license:
        name: MIT
    contact:
        name: unknown
openapi: 3.0.0
paths:
    '/{id}':
        patch:
            operationId: Update
            responses:
                '200':
                    description: Ok
                    content:
                        application/json:
                            schema:
                                $ref: '#/components/schemas/IUpdatableTest'
            security: []
            parameters:
                -
                    in: path
                    name: id
                    required: true
                    schema:
                        type: string
            requestBody:
                required: true
                content:
                    application/json:
                        schema:
                            $ref: '#/components/schemas/Omit_IUpdatableTest.id_'
servers:
    -
        url: /

```

```ts
import { Route, Patch, Body, Path, Controller } from 'tsoa';

interface IUpdatedOn {
  updatedOn: string;
}

interface IIdentifiable<TId> {
  id: TId;
}

interface ICoreTest {
  prop1: string;
  prop2: boolean;
}

interface IUpdatableTest
  extends Partial<ICoreTest>,
    IUpdatedOn,
    IIdentifiable<string> {}

@Route('')
export class TestController extends Controller {
  @Patch('{id}')
  public async update(
    @Path() id: string,
    @Body() data: Omit<IUpdatableTest, 'id'>
  ): Promise<IUpdatableTest> {
  }
}

```

## Possible Solution

N/A

## Steps to Reproduce

See provided ts code, where `tsoa spec-and-routes` are then performed on.

1. 2. 3. 4.

## Context (Environment)

Version of the library: 3.14.1
Version of NodeJS: 14.17.4

## Detailed Description

N/A

## Breaking change?

By fixing this bug, I guess it could result in other peoples specs potentially changing without realising.

### LINKED ISSUE #1204: Record<string, unknown> does not allow additional properties
## Sorting

- **I'm submitting a ...**

  - [x] bug report
  - [ ] feature request
  - [ ] support request

- I confirm that I

## Expected Behavior

A body with a field `attributes: Record<string, unknown>` and should allow additional properties even with `"noImplicitAdditionalProperties": "throw-on-extras",` in tsoa.json.

## Current Behavior

A body with a field `attributes: Record<string, unknown>`  with `"noImplicitAdditionalProperties": "throw-on-extras",` in tsoa.json triggers a validation error:

```
{"requestBody.attributes":{"message":"\"test\" is an excess property and therefore is not allowed","value":{"test":"test"}}}
```

## Steps to Reproduce

1. Add an interface `interface GeometryPostBody { attributes?: Record<string | number, unknown> }`
2. Set the interface as the expected body of a POST request
3. Try to POST with the body `{"attributes": {"test": "test"}}`
4. Receive a validation error `{"requestBody.attributes":{"message":"\"test\" is an excess property and therefore is not allowed","value":{"test":"test"}}}`

## Context (Environment)

Version of the library: 3.14.1
Version of NodeJS: v16.13.1

### LINKED ISSUE #1266: Generic interfaces generate invalid schema name
Not sure If I'm doing something wrong here, as it seems like something that is supported by tsoa, but when generating the swagger files for generic interfaces, the names of the schemas will contain characters (mainly %) that SwaggerUI / OpenAPI can't work with. Sanitizing the schema name manually and then running the server fixes the issue

## Sorting

  - [x] bug report

- I confirm that I

## Expected Behavior

TSOA should generate a valid swagger.json file. Currently SwaggerUI can't resolve the reference, and pasting this in editor.swagger.io also informs the schema name is not valid:

```
Semantic error at components.schemas.FormatPayload_%5BExampleFormat%5D_
Component names can only contain the characters A-Z a-z 0-9 - . _
Jump to line 8
```

This is the schema where I manually sanitized the FormatPyaload_ExampleFormat_ name, which works:

```json
{
  "components": {
    "schemas": {
      "FormatPayload_ExampleFormat_": {
        "properties": {
          "example": {
            "properties": {
              "theValue": {
                "type": "string"
              }
            },
            "required": ["theValue"],
            "type": "object"
          }
        },
        "type": "object",
        "description": "Get the payload from a list of Format interfaces"
      }
    }
  }
}

```

I expect to write the code the same as it currently present, just that the generated swagger.json file is different.

## Current Behavior

```json
{
  "components": {
    "schemas": {
      "FormatPayload_%5BExampleFormat%5D_": {
        "properties": {
          "example": {
            "properties": {
              "theValue": {
                "type": "string"
              }
            },
            "required": ["theValue"],
            "type": "object"
          }
        },
        "type": "object",
        "description": "Get the payload from a list of Format interfaces"
      }
    }
  }
}

```

```ts
import { Body, Controller, Post, Route } from "tsoa";

export interface Format {
  key: string;
  value: unknown;
}

export interface ExampleFormat extends Format {
  key: "example";
  value: {
    theValue: string;
  };
}

interface CreateUserRequest {
  formats: FormatPayload<[ExampleFormat]>;
}

/**
 * Get the payload from a list of Format interfaces
 *
 * @example
 * ```
 *
 * type CreateOfferFormat = FormatPayload<[ExampleFormat]>
 *
 * // equal to
 * type CreateOfferFormat = {
 *  example: {
 *   theValue: string;
 *  }
 * }
 * ```
 */
export declare type FormatPayload<Fs extends Format[]> = {
  [Format in Fs[number] as Format["key"]]?: Format["value"];
};

@Route("example")
export class ExampleController extends Controller {
  @Post()
  public async createUser(
    @Body()
    requestBody: CreateUserRequest
  ) {
    console.log(requestBody);

    return 10;
  }
}
```

## Possible Solution

A possible solutions is to take into account characters allowed by OpenAPI. From the error message at editor.swagger.io this includes:

```
Component names can only contain the characters A-Z a-z 0-9 - . _
```

## Steps to Reproduce

1. 2. 3. 4.

Repo: https://github.com/TimoGlastra/tsoa-error

To reproduce:

1. clone the repo
2. `yarn install`
3. `yarn build`
4. See the generated swagger.json in the build folder with incorrect characters
5. `yarn start`
6. open http://localhost:3000, see the error displayed in swagger ui
7. Copy the swagger.json file and paste it at editor.swagger.io. See the error displayed in the error about incorrect naming.

## Context (Environment)

Version of the library: 4.1.0
Version of NodeJS: 16.13.0

## Detailed Description

Change I want to propose is to generate valid OpenAPI schema names for generic interface that doesn't use percent encoded names.

## Breaking change?

This will be a breaking change, and will most likely affect open api generator as the interface names will be different. However I highly doubt any OpenAPI generator can work with percent encoded schema names.

### LINKED ISSUE #1447: Recursive Partial validation error
tsoa throws a validation error when using a recursive partial type

## Sorting

- **I'm submitting a ...**

  - [x] bug report
  - [x] feature request
  - [ ] support request

- I confirm that I

## Expected Behavior

```ts
// settings.model.ts
type RecursivePartial<T> = {
  [P in keyof T]?: RecursivePartial<T[P]>;
};

// Creates an object where all properties, including subproperties are optional.
type SettingsUpdateParams = RecursivePartial<Settings>;

// settings.controller.ts
@SuccessResponse(200, "OK")
  @Post()
  @Security("access_token")
  public async set(@Request() request: ExpressRequest, @Body() params: SettingsUpdateParams): Promise<void> {
    this.settingsService.set(request.user.uid, params);
  }
```

This should allow any of the settings sub-properties to be valid. However, they are not, as a validation error is thrown.

## Current Behavior

```ts
type SettingsUpdateParams = {
  appearance?: Partial<AppearanceSettings>;
  beta?: Partial<BetaSettings>;
  //...
};
```

This was my workaround, manually making each sub property `Partial<>`

## Context (Environment)

Version of the library: 5.1.1
Version of NodeJS: 16.15.0

## Detailed Description

It would be nice if recursive types were supported by tsoa. Quality of life thing.

## Breaking change?

No.

### LINKED ISSUE #1472: Query parameter default value not generated to spec file in case of custom types
## Sorting

- **I'm submitting a ...**

  - [x] bug report
  - [ ] feature request
  - [ ] support request

- I confirm that I

## Expected Behavior

In case of custom type `TimeFormat` the following:
```ts
type TimeFormat = 'iso8601' | 'epoch_ms';

    @Query()
    timeFormat: TimeFormat = 'iso8601'
```
should generate open api spec parameter like:
```json
	{
		"in": "query",
		"name": "timeFormat",
		"required": false,
		"schema": {
			"default": "iso8601",
			"type": "string",
			"enum": [
				"iso8601",
				"epoch_ms"
			]
		}
	}
```

## Current Behavior

Instead it generates:
```json
	{
		"in": "query",
		"name": "timeFormat",
		"required": false,
		"schema": {
			"$ref": "#/components/schemas/TimeFormat"
		}
	}
```

Currently, if type is written open directly into query parameter definition (instead of using custom type):
```ts
    @Query()
    timeFormat: 'iso8601' | 'epoch_ms' = 'iso8601'
```
it works, default value is generated.

## Context (Environment)

Version of the library: 5.1.1 and 6.0.0-rc.3
Version of NodeJS: v20.2.0

nemnandor:
### All Submissions:

**Closing issues**

closes #477
closes #623
closes #681
closes #1191
closes #1204
closes #1266
closes #1447 
closes #1472

Fixes:

I have solved locally some problems which are needed for my company. After that I read the TSOA commit guides, so some fixes maybe have no related issues.

1. Namespaces: Every type should be stored with their full-path namespaced types. Even if the source code not uses this name. This reduces conflicts between types.
2. Name conflicts: If every reference type name parts are uniq, then the full type name can be handled as uniq. In this type name, we need to sign the jsdoc params if exists, because its they can make differences in the result. Somehow the reference type names are transformed with an irreversible getRefTypeName(), here we need to check if the result name comes from the same original type names, or from more different names.
3. Handling interfaces - namespaces - enums which have more than one declaration
4. Keep jsdoc comments + defaults after mapping - as intellisense shows
5. Do not use original definition in mapped types, because type can change.
6. Accept unions, intersections, nulls, enums in mapped type.
7. Fix referencer issues. TypeToTypeNode & TypeNodeToType functions are not exactly the inverse function of each other -> use the original types, and remove conversions.
8. Do not name more complicated reference type internal parts as "any". - It is needed to recognize name conflicts.
9. Better keyof handling
10. Do not mark required if a field has a default value. (Not too relevant, but more beautiful)
11. Default values from JSdoc uses the expected type. `@default true` is not a string, but a boolean.

- My code sometimes returns with an equivalent type definition, not the original (If #1490 is a bug, then I made bugs). Check original test changes.
- I have not implemented Partial<string> or Partial<number>-like mapped types. I can implement it, but not really know if it is interesting or not/where is the end of this strange types.
- I solved name conflict handling, but for it I generate the type name every time (previously tsoa made it in the else branch, but normally we used the typescript name). For some name, some type formations are not implemented, like:
```type A = Partial<{ [a in keyof B]: string }>```
But there is a workaround:
```type AInternal = { [a in keyof B]: string }; type A = Partial<AInternal>```
- A mapped type can change the type of a member, but can not change the jsdoc of it (sometimes deletes it, but sometimes not). So we can create a number member (with default value 2, and minimum 0), and after that, we can construct a mapped type which can change the type to string. Everything is fine with intellisense, but we can define a type which should have a value, because if not, the default value will have another type, so typecheck fails. Now TSOA handles type restrictions which are related only to one type, so in this case there will be not relevant type checks, not really interesting.
- If a type (interface) has multiple declaration, and both declarations contains the same member with different jsdoc, the program chooses one of them.
- Union types in typescript are a little bit strange, A | B ~ { some properties of A & some properties of B}. TSOA validation not handles them correctly (TSOA validation implementation: A | B = Only A | Only B). In a type with multiple different additionalProperties type I use type unions to "approxximate" the expected operation. It is ok until someone fixes TSOA validation. But I'm not sure, that most programmers love or hate this TSOA validation bug, I enjoy it, it is more strict than ts.

I'm sure, that there are a lot of breaking changes.

**Test plan**

Added test type literals to the TestModel interface. Every literal checks one solved problem. The members represents one test case.

### DIFF
--- a/packages/cli/src/metadataGeneration/exceptions.ts
+++ b/packages/cli/src/metadataGeneration/exceptions.ts
@@ -11,7 +11,7 @@ export class GenerateMetadataError extends Error {
 }
 
 export class GenerateMetaDataWarning {
-  constructor(private message: string, private node: Node | TypeNode, private onlyCurrent = false) {}
+  constructor(private message: string, private node: Node | TypeNode, private onlyCurrent = false) { }
 
   toString() {
     return `Warning: ${this.message}\n${prettyLocationOfNode(this.node)}\n${prettyTroubleCause(this.node, this.onlyCurrent)}`;
@@ -20,19 +20,23 @@ export class GenerateMetaDataWarning {
 
 export function prettyLocationOfNode(node: Node | TypeNode) {
   const sourceFile = node.getSourceFile();
-  const token = node.getFirstToken() || node.parent.getFirstToken();
-  const start = token ? `:${sourceFile.getLineAndCharacterOfPosition(token.getStart()).line + 1}` : '';
-  const end = token ? `:${sourceFile.getLineAndCharacterOfPosition(token.getEnd()).line + 1}` : '';
-  const normalizedPath = normalize(`${sourceFile.fileName}${start}${end}`);
-  return `At: ${normalizedPath}.`;
+  if (sourceFile) {
+    const token = node.getFirstToken() || node.parent.getFirstToken();
+    const start = token ? `:${sourceFile.getLineAndCharacterOfPosition(token.getStart()).line + 1}` : '';
+    const end = token ? `:${sourceFile.getLineAndCharacterOfPosition(token.getEnd()).line + 1}` : '';
+    const normalizedPath = normalize(`${sourceFile.fileName}${start}${end}`);
+    return `At: ${normalizedPath}.`;
+  } else {
+    return `At unknown position...`;
+  }
 }
 
 export function prettyTroubleCause(node: Node | TypeNode, onlyCurrent = false) {
   let name: string;
   if (onlyCurrent || !node.parent) {
-    name = node.pos !== -1 ? node.getText() : (node as any).name.text;
+    name = node.pos !== -1 ? node.getText() : ((node as any).name?.text || '<unknown name>');
   } else {
-    name = node.parent.pos !== -1 ? node.parent.getText() : (node as any).parent.name.text;
+    name = node.parent.pos !== -1 ? node.parent.getText() : ((node as any).parent.name?.text || '<unknown name>');
   }
   return `This was caused by '${name}'`;
 }
--- a/packages/cli/src/metadataGeneration/metadataGenerator.ts
+++ b/packages/cli/src/metadataGeneration/metadataGenerator.ts
@@ -12,7 +12,7 @@ export class MetadataGenerator {
   public readonly typeChecker: TypeChecker;
   private readonly program: Program;
   private referenceTypeMap: Tsoa.ReferenceTypeMap = {};
-  private modelDefinitionPosMap: { [name: string]: { fileName: string, pos: number }[] } = {};
+  private modelDefinitionPosMap: { [name: string]: Array<{ fileName: string; pos: number }> } = {};
   private expressionOrigNameMap: Record<string, string> = {};
 
   constructor(
@@ -223,12 +223,12 @@ export class MetadataGenerator {
     return this.referenceTypeMap[refName];
   }
 
-  public CheckModelUnicity(refName: string, positions: { fileName: string, pos: number }[]) {
+  public CheckModelUnicity(refName: string, positions: Array<{ fileName: string; pos: number }>) {
     if (!this.modelDefinitionPosMap[refName]) {
       this.modelDefinitionPosMap[refName] = positions;
     } else {
-      let origPositions = this.modelDefinitionPosMap[refName];
-      if (!(origPositions.length == positions.length && positions.every(pos => origPositions.find(origPos => pos.pos == origPos.pos && pos.fileName == origPos.fileName)))) {
+      const origPositions = this.modelDefinitionPosMap[refName];
+      if (!(origPositions.length === positions.length && positions.every(pos => origPositions.find(origPos => pos.pos === origPos.pos && pos.fileName === origPos.fileName)))) {
         throw new Error(`Found 2 different model definitions for model ${refName}: orig: ${JSON.stringify(origPositions)}, act: ${JSON.stringify(positions)}`);
       }
     }
@@ -238,7 +238,7 @@ export class MetadataGenerator {
     if (!this.expressionOrigNameMap[formattedRefName]) {
       this.expressionOrigNameMap[formattedRefName] = refName;
     } else {
-      if (this.expressionOrigNameMap[formattedRefName] != refName) {
+      if (this.expressionOrigNameMap[formattedRefName] !== refName) {
         throw new Error(`Found 2 different type expressions for formatted name "${formattedRefName}": orig: "${this.expressionOrigNameMap[formattedRefName]}", act: "${refName}"`);
       }
     }
--- a/packages/cli/src/metadataGeneration/metadataGenerator.ts
+++ b/packages/cli/src/metadataGeneration/metadataGenerator.ts
@@ -1,18 +1,19 @@
+import { Config, Tsoa } from '@tsoa/runtime';
 import { minimatch } from 'minimatch';
+import { createProgram, forEachChild, isClassDeclaration, type ClassDeclaration, type CompilerOptions, type Program, type TypeChecker } from 'typescript';
+import { getDecorators } from '../utils/decoratorUtils';
 import { importClassesFromDirectories } from '../utils/importClassesFromDirectories';
 import { ControllerGenerator } from './controllerGenerator';
 import { GenerateMetadataError } from './exceptions';
-import { Config, Tsoa } from '@tsoa/runtime';
 import { TypeResolver } from './typeResolver';
-import { getDecorators } from '../utils/decoratorUtils';
-import { type TypeChecker, type Program, type ClassDeclaration, type CompilerOptions, createProgram, forEachChild, isClassDeclaration } from 'typescript';
 
 export class MetadataGenerator {
   public readonly controllerNodes = new Array<ClassDeclaration>();
   public readonly typeChecker: TypeChecker;
   private readonly program: Program;
   private referenceTypeMap: Tsoa.ReferenceTypeMap = {};
-  private circularDependencyResolvers = new Array<(referenceTypes: Tsoa.ReferenceTypeMap) => void>();
+  private modelDefinitionPosMap: { [name: string]: { fileName: string, pos: number }[] } = {};
+  private expressionOrigNameMap: Record<string, string> = {};
 
   constructor(
     entryFile: string,
@@ -35,7 +36,6 @@ export class MetadataGenerator {
 
     this.checkForMethodSignatureDuplicates(controllers);
     this.checkForPathParamSignatureDuplicates(controllers);
-    this.circularDependencyResolvers.forEach(c => c(this.referenceTypeMap));
 
     return {
       controllers,
@@ -214,17 +214,34 @@ export class MetadataGenerator {
 
   public AddReferenceType(referenceType: Tsoa.ReferenceType) {
     if (!referenceType.refName) {
-      return;
+      throw new Error('no reference type name found');
     }
-    this.referenceTypeMap[decodeURIComponent(referenceType.refName)] = referenceType;
+    this.referenceTypeMap[referenceType.refName] = referenceType;
   }
 
   public GetReferenceType(refName: string) {
     return this.referenceTypeMap[refName];
   }
 
-  public OnFinish(callback: (referenceTypes: Tsoa.ReferenceTypeMap) => void) {
-    this.circularDependencyResolvers.push(callback);
+  public CheckModelUnicity(refName: string, positions: { fileName: string, pos: number }[]) {
+    if (!this.modelDefinitionPosMap[refName]) {
+      this.modelDefinitionPosMap[refName] = positions;
+    } else {
+      let origPositions = this.modelDefinitionPosMap[refName];
+      if (!(origPositions.length == positions.length && positions.every(pos => origPositions.find(origPos => pos.pos == origPos.pos && pos.fileName == origPos.fileName)))) {
+        throw new Error(`Found 2 different model definitions for model ${refName}: orig: ${JSON.stringify(origPositions)}, act: ${JSON.stringify(positions)}`);
+      }
+    }
+  }
+
+  public CheckExpressionUnicity(formattedRefName: string, refName: string) {
+    if (!this.expressionOrigNameMap[formattedRefName]) {
+      this.expressionOrigNameMap[formattedRefName] = refName;
+    } else {
+      if (this.expressionOrigNameMap[formattedRefName] != refName) {
+        throw new Error(`Found 2 different type expressions for formatted name "${formattedRefName}": orig: "${this.expressionOrigNameMap[formattedRefName]}", act: "${refName}"`);
+      }
+    }
   }
 
   private buildControllers() {
--- a/packages/cli/src/metadataGeneration/typeResolver.ts
+++ b/packages/cli/src/metadataGeneration/typeResolver.ts
@@ -193,7 +193,7 @@ export class TypeResolver {
             .filter(property => isIgnored(property) === false)
             // Transform to property
             .map(property => {
-              const propertyType = this.current.typeChecker.getTypeOfSymbol(property);
+              const propertyType = this.current.typeChecker.getTypeOfSymbolAtLocation(property, this.typeNode);
 
               const typeNode = this.current.typeChecker.typeToTypeNode(propertyType, undefined, ts.NodeBuilderFlags.NoTruncation)!;
               const parent = getOneOrigDeclaration(property); //If there are more declarations, we need to get one of them, from where we want to recognize jsDoc
--- a/packages/cli/src/metadataGeneration/typeResolver.ts
+++ b/packages/cli/src/metadataGeneration/typeResolver.ts
@@ -121,7 +121,7 @@ export class TypeResolver {
           description: this.getNodeDescription(propertySignature),
           format: this.getNodeFormat(propertySignature),
           name: (propertySignature.name as ts.Identifier).text,
-          required: !propertySignature.questionToken && def === undefined,
+          required: !propertySignature.questionToken,
           type,
           validators: getPropertyValidators(propertySignature) || {},
           deprecated: isExistJSDocTag(propertySignature, tag => tag.tagName.text === 'deprecated'),
@@ -219,7 +219,7 @@ export class TypeResolver {
               // Push property
               return {
                 name: property.getName(),
-                required: required && def === undefined,
+                required,
                 deprecated: parent ? isExistJSDocTag(parent, tag => tag.tagName.text === 'deprecated') || isDecorator(parent, identifier => identifier.text === 'Deprecated') : false,
                 type,
                 default: def,
@@ -1272,7 +1272,7 @@ export class TypeResolver {
       example: this.getNodeExample(propertySignature),
       format: this.getNodeFormat(propertySignature),
       name: identifier.text,
-      required: required && def === undefined,
+      required,
       type: new TypeResolver(propertySignature.type, this.current, propertySignature.type.parent, this.context).resolve(),
       validators: getPropertyValidators(propertySignature) || {},
       deprecated: isExistJSDocTag(propertySignature, tag => tag.tagName.text === 'deprecated'),
@@ -1311,7 +1311,7 @@ export class TypeResolver {
       example: this.getNodeExample(propertyDeclaration),
       format: this.getNodeFormat(propertyDeclaration),
       name: identifier.text,
-      required: required && def === undefined,
+      required,
       type,
       validators: getPropertyValidators(propertyDeclaration) || {},
       // class properties and constructor parameters may be deprecated either via jsdoc annotation or decorator
--- a/packages/cli/src/metadataGeneration/typeResolver.ts
+++ b/packages/cli/src/metadataGeneration/typeResolver.ts
@@ -10,16 +10,16 @@ import { getInitializerValue } from './initializer-value';
 import { MetadataGenerator } from './metadataGenerator';
 
 const localReferenceTypeCache: { [typeName: string]: Tsoa.ReferenceType } = {};
-const inProgressTypes: { [typeName: string]: ((realType: Tsoa.ReferenceType) => void)[] } = {};
+const inProgressTypes: { [typeName: string]: Array<(realType: Tsoa.ReferenceType) => void> } = {};
 
 type OverrideToken = ts.Token<ts.SyntaxKind.QuestionToken> | ts.Token<ts.SyntaxKind.PlusToken> | ts.Token<ts.SyntaxKind.MinusToken> | undefined;
 type UsableDeclaration = ts.InterfaceDeclaration | ts.ClassDeclaration | ts.PropertySignature | ts.TypeAliasDeclaration | ts.EnumMember;
 type UsableDeclarationWithoutPropertySignature = Exclude<UsableDeclaration, ts.PropertySignature>;
 interface Context {
   [name: string]: {
-    type: ts.TypeNode,
-    name: string
-  }
+    type: ts.TypeNode;
+    name: string;
+  };
 }
 
 export class TypeResolver {
@@ -29,7 +29,7 @@ export class TypeResolver {
     private readonly parentNode?: ts.Node,
     private context: Context = {},
     private readonly referencer?: ts.Type,
-  ) { }
+  ) {}
 
   public static clearCache() {
     Object.keys(localReferenceTypeCache).forEach(key => {
@@ -114,7 +114,7 @@ export class TypeResolver {
       const properties = this.typeNode.members.filter(ts.isPropertySignature).reduce<Tsoa.Property[]>((res, propertySignature: ts.PropertySignature) => {
         const type = new TypeResolver(propertySignature.type as ts.TypeNode, this.current, propertySignature, this.context).resolve();
 
-        let def = TypeResolver.getDefault(propertySignature);
+        const def = TypeResolver.getDefault(propertySignature);
         const property: Tsoa.Property = {
           example: this.getNodeExample(propertySignature),
           default: def,
@@ -157,40 +157,37 @@ export class TypeResolver {
     }
 
     if (ts.isMappedTypeNode(this.typeNode)) {
-      let mappedTypeNode = this.typeNode;
+      const mappedTypeNode = this.typeNode;
       const getOneOrigDeclaration = (prop: ts.Symbol): ts.Declaration | undefined => {
         if (prop.declarations) {
           return prop.declarations[0];
         }
-        let syntheticOrigin: ts.Symbol = (prop as any).links?.syntheticOrigin;
-        if (syntheticOrigin && syntheticOrigin.name == prop.name) { //Otherwise losts jsDoc like in intellisense
+        const syntheticOrigin: ts.Symbol = (prop as any).links?.syntheticOrigin;
+        if (syntheticOrigin && syntheticOrigin.name === prop.name) {
+          //Otherwise losts jsDoc like in intellisense
           return syntheticOrigin.declarations?.[0];
         }
         return undefined;
-      }
+      };
       const isIgnored = (prop: ts.Symbol) => {
         const declaration = getOneOrigDeclaration(prop);
-        return declaration !== undefined &&
-          (
-            getJSDocTagNames(declaration).some(tag => tag == 'ignore') ||
-            (
-              !ts.isPropertyDeclaration(declaration) &&
-              !ts.isPropertySignature(declaration) &&
-              !ts.isParameter(declaration)
-            )
-          );
+        return (
+          declaration !== undefined &&
+          (getJSDocTagNames(declaration).some(tag => tag === 'ignore') || (!ts.isPropertyDeclaration(declaration) && !ts.isPropertySignature(declaration) && !ts.isParameter(declaration)))
+        );
       };
 
-      let calcMappedType = (type: ts.Type): Tsoa.Type => {
-        if (type.flags & ts.TypeFlags.Union) { //Intersections are not interesting somehow...
-          let types = (type as ts.UnionType).types;
-          let resolvedTypes = types.map(calcMappedType);
+      const calcMappedType = (type: ts.Type): Tsoa.Type => {
+        if (type.flags & ts.TypeFlags.Union) {
+          //Intersections are not interesting somehow...
+          const types = (type as ts.UnionType).types;
+          const resolvedTypes = types.map(calcMappedType);
           return {
             dataType: 'union',
-            types: resolvedTypes
+            types: resolvedTypes,
           };
         } else if (type.flags & ts.TypeFlags.Object) {
-          let typeProperties: ts.Symbol[] = type.getProperties();
+          const typeProperties: ts.Symbol[] = type.getProperties();
           const properties: Tsoa.Property[] = typeProperties
             // Ignore methods, getter, setter and @ignored props
             .filter(property => isIgnored(property) === false)
@@ -199,26 +196,22 @@ export class TypeResolver {
               const propertyType = this.current.typeChecker.getTypeOfSymbol(property);
 
               const typeNode = this.current.typeChecker.typeToTypeNode(propertyType, undefined, ts.NodeBuilderFlags.NoTruncation)!;
-              let parent = getOneOrigDeclaration(property); //If there are more declarations, we need to get one of them, from where we want to recognize jsDoc
+              const parent = getOneOrigDeclaration(property); //If there are more declarations, we need to get one of them, from where we want to recognize jsDoc
               const type = new TypeResolver(typeNode, this.current, parent, this.context, propertyType).resolve();
 
-              let required = !(property.flags & ts.SymbolFlags.Optional);
+              const required = !(property.flags & ts.SymbolFlags.Optional);
 
               const comments = property.getDocumentationComment(this.current.typeChecker);
-              let description = comments.length ? ts.displayPartsToString(comments) : undefined;
+              const description = comments.length ? ts.displayPartsToString(comments) : undefined;
 
-              let initializer = (parent as any)?.initializer;
-              let def = initializer ? getInitializerValue(initializer, this.current.typeChecker) :
-                parent ? TypeResolver.getDefault(parent) : undefined;
+              const initializer = (parent as any)?.initializer;
+              const def = initializer ? getInitializerValue(initializer, this.current.typeChecker) : parent ? TypeResolver.getDefault(parent) : undefined;
 
               // Push property
               return {
                 name: property.getName(),
                 required: required && def === undefined,
-                deprecated: parent ? (
-                  isExistJSDocTag(parent, tag => tag.tagName.text === 'deprecated') ||
-                  isDecorator(parent, identifier => identifier.text === 'Deprecated')
-                ) : false,
+                deprecated: parent ? isExistJSDocTag(parent, tag => tag.tagName.text === 'deprecated') || isDecorator(parent, identifier => identifier.text === 'Deprecated') : false,
                 type,
                 default: def,
                 // validators are disjunct via types, so it is now OK.
@@ -228,7 +221,7 @@ export class TypeResolver {
                 description,
                 format: parent ? this.getNodeFormat(parent) : undefined,
                 example: parent ? this.getNodeExample(parent) : undefined,
-                extensions: parent ? this.getNodeExtension(parent) : undefined
+                extensions: parent ? this.getNodeExtension(parent) : undefined,
               };
             });
 
@@ -236,14 +229,14 @@ export class TypeResolver {
             dataType: 'nestedObjectLiteral',
             properties,
           };
-          let indexInfos = this.current.typeChecker.getIndexInfosOfType(type);
-          let indexTypes = indexInfos.map(indexInfo => {
+          const indexInfos = this.current.typeChecker.getIndexInfosOfType(type);
+          const indexTypes = indexInfos.map(indexInfo => {
             const typeNode = this.current.typeChecker.typeToTypeNode(indexInfo.type, undefined, ts.NodeBuilderFlags.NoTruncation)!;
             const type = new TypeResolver(typeNode, this.current, mappedTypeNode, this.context, indexInfo.type).resolve();
             return type;
           });
           if (indexTypes.length) {
-            if (indexTypes.length == 1) {
+            if (indexTypes.length === 1) {
               objectLiteral.additionalProperties = indexTypes[0];
             } else {
               // { [k: string]: string; } & { [k: number]: number; }
@@ -255,7 +248,7 @@ export class TypeResolver {
               //Every additional property key assumed as string
               objectLiteral.additionalProperties = {
                 dataType: 'union',
-                types: indexTypes
+                types: indexTypes,
               };
             }
           }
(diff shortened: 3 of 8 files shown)

### CODE REVIEW

File: tests/unit/swagger/definitionsGeneration/definitions.spec.ts [hunk 1]
```diff
@@ -1119,7 +1137,7 @@ describe('Definition generation', () => {
               {
                 properties: {
                   list: {
-                    items: { $ref: '#/definitions/ThingContainerWithTitle_string_' },
+                    items: { type: 'number', format: 'double' },
```
nemnandor, line 1140: Equivalent change, but maybe the previous solution was better readable

tests/unit/swagger/definitionsGeneration/definitions.spec.ts
```diff
@@ -1169,13 +1188,13 @@ describe('Definition generation', () => {
                     format: undefined,
                   },
                   indexedResponseObject: {
-                    $ref: '#/definitions/Record_id.any_',
+                    $ref: '#/definitions/Record_id._myProp1-string__',
```
nemnandor, line 1191: Previous name generation worked well only if we were not in a nested mapped type. Fixed to check&handle name conflicts

tests/unit/swagger/definitionsGeneration/definitions.spec.ts
```diff
@@ -1169,13 +1188,13 @@ describe('Definition generation', () => {
...
                     description: undefined,
                     example: undefined,
                     format: undefined,
                   },
                   indexedType: { type: 'string', default: undefined, description: undefined, format: undefined, example: undefined },
-                  indexedTypeToAlias: { $ref: '#/definitions/IndexedInterfaceAlias', description: undefined, format: undefined, example: undefined },
+                  indexedTypeToAlias: { $ref: '#/definitions/IndexedInterface', description: undefined, format: undefined, example: undefined },
```
nemnandor, line 1197: Equivalent change, but maybe previous was better readable in some cases

tests/unit/swagger/definitionsGeneration/definitions.spec.ts
```diff
@@ -131,7 +131,7 @@ describe('Definition generation', () => {
           'TestSubModelContainerNamespace.InnerNamespace.TestSubModelContainer2',
           'TestSubModel2',
           'TestSubModelNamespace.TestSubModelNS',
-          'TsoaTest.TestModel73',
+          'tsoaTest.TsoaTest.TestModel73',
```
nemnandor, line 134: For safe name conflict handling, every type needs to use its fully namespaced name. Not accepted to think, that TsoaTest type is equal to another interface in the tsoaTest namespace

tests/unit/swagger/definitionsGeneration/definitions.spec.ts
```diff
@@ -564,6 +564,7 @@ describe('Definition generation', () => {
                   default: undefined,
                 },
               },
+              required: ['record-foo', 'record-bar'],
```
nemnandor, line 567: Wrong previous tests

tests/unit/swagger/definitionsGeneration/definitions.spec.ts
```diff
@@ -834,7 +852,7 @@ describe('Definition generation', () => {
               example: 42,
               minimum: 42,
               maximum: 42,
-              default: '42',
+              default: 42,
```
nemnandor, line 855: Now default types are ok, so we can use default values with @default jsdoc comment

packages/cli/src/metadataGeneration/exceptions.ts
```diff
@@ -11,7 +11,7 @@ export class GenerateMetadataError extends Error {
 }
```
nemnandor, line 12: Small thing, do not crash while trying to make exceptions to throw (or when make warning messages).

packages/cli/src/metadataGeneration/metadataGenerator.ts
```diff
@@ -35,7 +36,6 @@ export class MetadataGenerator {
 
     this.checkForMethodSignatureDuplicates(controllers);
     this.checkForPathParamSignatureDuplicates(controllers);
-    this.circularDependencyResolvers.forEach(c => c(this.referenceTypeMap));
```
nemnandor, line 38: Handled in another place

packages/cli/src/metadataGeneration/metadataGenerator.ts
```diff
@@ -214,17 +214,34 @@ export class MetadataGenerator {
 
   public AddReferenceType(referenceType: Tsoa.ReferenceType) {
     if (!referenceType.refName) {
-      return;
+      throw new Error('no reference type name found');
     }
-    this.referenceTypeMap[decodeURIComponent(referenceType.refName)] = referenceType;
+    this.referenceTypeMap[referenceType.refName] = referenceType;
```
nemnandor, line 219: decodeURIComponent() not interesting here, because typeResolver guarantees, that only OpenApi-accepted names can be generated.

packages/cli/src/metadataGeneration/metadataGenerator.ts
```diff
@@ -214,17 +214,34 @@ export class MetadataGenerator {
...
   }
 
   public GetReferenceType(refName: string) {
     return this.referenceTypeMap[refName];
   }
 
-  public OnFinish(callback: (referenceTypes: Tsoa.ReferenceTypeMap) => void) {
-    this.circularDependencyResolvers.push(callback);
+  public CheckModelUnicity(refName: string, positions: Array<{ fileName: string; pos: number }>) {
```
nemnandor, line 226: Type unicity check:
* We want to check, that type A<B<...>,{ c: C}> is uniq or not. For this, we need to check if every A, B & C type identifier definitions are uniq or not. Uniq = A type can have more than one definition, but their definition positions array must be uniq. Position = fileName + pos. Not stored, so not interesting, which local machine the check is made.
* A<B> and A_B_ types are both renamed to A_B_ to use it in OpenApi. This is a second type of name conflicts.

tests/fixtures/testModel.ts
```diff
@@ -209,6 +209,324 @@ export interface TestModel extends Model {
...
+     * //Alma
+     * `\\`: '\n'
+     *
+     * }
+     *
+     */
+    jsonCharacters: undefined; //type is not really interesting
+  };
+
+  jsDocTypeNames?: {
```
nemnandor, line 263: This types are different because of their jsdoc comments. Should generate different TSOA types for it (with, of course, different names).
Test checks situations which were handled previously as Partial_any_ to cover a more special situation

tests/fixtures/testModel.ts
```diff
@@ -209,6 +209,324 @@ export interface TestModel extends Model {
...
+      /** @extension {"x-key-1": "value-1"} */
+      [a: string]: string;
+    }>;
+    indexedIgnored: Partial<{
+      /** @ignore */
+      [a: string]: string;
+    }>;
+  };
+
+  jsdocMap?: {
```
nemnandor, line 339: Shows how to change jsdocs while changing the interface with mapped types

tests/fixtures/testModel.ts
```diff
@@ -209,6 +209,324 @@ export interface TestModel extends Model {
...
+    doubleReplacedTypes: ReplaceStringAndNumberTypes<ReplaceStringAndNumberTypes<JsDocced>>;
+    postfixed: Postfixed<JsDocced, '_PostFix'>;
+    values: Values<JsDocced>;
+    typesValues: InternalTypes<Values<JsDocced>>;
+    onlyOneValue: JsDocced['numberValue'];
+    synonym: JsDoccedSynonym;
+    synonym2: JsDoccedSynonym2;
+  };
+
+  duplicatedDefinitions?: {
```
nemnandor, line 352: identifiers with more than one definition

packages/cli/src/metadataGeneration/typeResolver.ts
```diff
@@ -1,22 +1,25 @@
...
+const inProgressTypes: { [typeName: string]: Array<(realType: Tsoa.ReferenceType) => void> } = {};
 
 type OverrideToken = ts.Token<ts.SyntaxKind.QuestionToken> | ts.Token<ts.SyntaxKind.PlusToken> | ts.Token<ts.SyntaxKind.MinusToken> | undefined;
 type UsableDeclaration = ts.InterfaceDeclaration | ts.ClassDeclaration | ts.PropertySignature | ts.TypeAliasDeclaration | ts.EnumMember;
 type UsableDeclarationWithoutPropertySignature = Exclude<UsableDeclaration, ts.PropertySignature>;
 interface Context {
-  [name: string]: ts.TypeReferenceNode | ts.TypeNode;
+  [name: string]: {
+    type: ts.TypeNode;
+    name: string;
```
nemnandor, line 21: name is only for cache

packages/cli/src/metadataGeneration/typeResolver.ts
```diff
@@ -111,13 +113,15 @@ export class TypeResolver {
...
+        const def = TypeResolver.getDefault(propertySignature);
         const property: Tsoa.Property = {
           example: this.getNodeExample(propertySignature),
-          default: getJSDocComment(propertySignature, 'default'),
+          default: def,
           description: this.getNodeDescription(propertySignature),
           format: this.getNodeFormat(propertySignature),
           name: (propertySignature.name as ts.Identifier).text,
-          required: !propertySignature.questionToken,
+          required: !propertySignature.questionToken && def === undefined,
```
nemnandor, line 124: If a type has default value, then it can not be required.
If a string has a number default value (for example, because of a type mapping), then no problem, the not defined value will be replaced with the number, and after that the typeCheck says, it is not the expected type.

packages/cli/src/metadataGeneration/typeResolver.ts
```diff
@@ -152,143 +156,224 @@ export class TypeResolver {
...
+          const typeProperties: ts.Symbol[] = type.getProperties();
+          const properties: Tsoa.Property[] = typeProperties
+            // Ignore methods, getter, setter and @ignored props
+            .filter(property => isIgnored(property) === false)
+            // Transform to property
+            .map(property => {
+              const propertyType = this.current.typeChecker.getTypeOfSymbol(property);
+
+              const typeNode = this.current.typeChecker.typeToTypeNode(propertyType, undefined, ts.NodeBuilderFlags.NoTruncation)!;
+              const parent = getOneOrigDeclaration(property); //If there are more declarations, we need to get one of them, from where we want to recognize jsDoc
```
nemnandor, line 199: If there are more than one declaration for a type member, then choose one of them for handling jsdoc.

packages/cli/src/metadataGeneration/typeResolver.ts
```diff
@@ -632,87 +660,376 @@ export class TypeResolver {
...
+        pos: declaration.pos,
+      }));
+      this.current.CheckModelUnicity(name, declarationPositions);
     }
+    return name;
+  }
 
-    const name = this.contextualizedName(resolvableName);
+  private calcMemberJsDocProperties(arg: ts.PropertySignature): string {
+    const def = TypeResolver.getDefault(arg);
```
nemnandor, line 726: If two types differ only in jsdoc, then their name needs to differ, too. Because of it, I need to inject the jsdoc parameters into the type name.

packages/cli/src/metadataGeneration/typeResolver.ts
```diff
@@ -1213,6 +1504,61 @@ export class TypeResolver {
...
     return getDecorators(node, identifier => identifier.text === id);
   }
+
+  private static getDefault(node: ts.Node) {
+    const defaultStr = getJSDocComment(node, 'default');
+    if (typeof defaultStr == 'string' && defaultStr !== 'undefined') {
+      let textStartCharacter: `"` | "'" | '`' | undefined = undefined;
+      const inString = () => textStartCharacter !== undefined;
+
+      let formattedStr = '';
```
nemnandor, line 1514: #623

packages/cli/src/metadataGeneration/typeResolver.ts
```diff
@@ -152,143 +156,233 @@ export class TypeResolver {
...
-        properties,
+        } else if (type.flags & ts.TypeFlags.Undefined) {
+          return {
+            dataType: 'undefined',
+          };
+        } else if (type.flags & ts.TypeFlags.Null) {
+          return {
+            dataType: 'enum',
+            enums: [null],
+          };
```
nemnandor, line 197: undefined & null types are interesting if strictNullChecks flag is true in ts compilerOptions

WoH:
Can you take a look at CI? Let's start there.

packages/cli/src/metadataGeneration/typeResolver.ts
WoH: How does this interact with: https://github.com/lukeautry/tsoa/pull/1454

My approach was to leave that up to  the maping of internal types -> OpenAPI 2/3, while keeping the metadata rather unrefined and applying "intentions" later.

packages/cli/src/metadataGeneration/typeResolver.ts
WoH: In all other cases we throw if there are multiple. This seems like an intentional deviation?

packages/cli/src/metadataGeneration/typeResolver.ts
```diff
@@ -152,143 +156,233 @@ export class TypeResolver {
...
+    }
 
-      if (type.aliasSymbol) {
-        let declaration = type.aliasSymbol.declarations?.[0] as ts.TypeAliasDeclaration | ts.EnumDeclaration | ts.DeclarationStatement;
-        if (declaration.name) {
-          declaration = this.getModelTypeDeclaration(declaration.name as ts.EntityName) as ts.TypeAliasDeclaration | ts.EnumDeclaration | ts.DeclarationStatement;
+    // keyof
+    if (ts.isTypeOperatorNode(this.typeNode) && this.typeNode.operator === ts.SyntaxKind.KeyOfKeyword) {
+      const type = this.current.typeChecker.getTypeFromTypeNode(this.typeNode);
+      if (type.getFlags() & ts.TypeFlags.Index) {
```
WoH, line 285: Is there a helper that makes this typesafe without casting?

packages/cli/src/metadataGeneration/typeResolver.ts
```diff
@@ -152,143 +156,233 @@ export class TypeResolver {
...
+        // in case of generic: keyof T. Not handles all possible cases
+        const symbol = (type as ts.IndexType).type.getSymbol();
+        if (symbol && symbol.getFlags() & ts.TypeFlags.TypeParameter) {
+          const typeName = symbol.getEscapedName();
```
WoH, line 289: add: assert(typeof typename === 'string')

packages/cli/src/metadataGeneration/typeResolver.ts
nemnandor: Ok, I removed the automatic non-required if there is a default value.

packages/cli/src/metadataGeneration/typeResolver.ts
nemnandor: Currently, tsoa gets the [0]-th element. I think my code is not worse than the original.
Of course, I can make, that the parent member of TypeResolver is not a single value, but an array, and everywhere we want to use it, we check if every elements gets the same result or not. If you think it makes a better world.

packages/cli/src/metadataGeneration/typeResolver.ts
nemnandor: Good notice, I have fixed it.

packages/cli/src/metadataGeneration/typeResolver.ts
nemnandor: fixed

WoH (approved):

WoH:
LGTM, thanks for the PR and sorry for the time it took to review.

PR #1533, opened 2024-01-08
add ability to include custom swagger extensions in config

### LINKED ISSUE #1532: Add ability for custom annotations to write Extensions
## Sorting

- **I'm submitting a ...**

  - [ ] bug report
  - [x] feature request
  - [ ] support request

- I confirm that I

## Expected Behavior

```yaml
        "x-rate-limit-docs": [
          {
            "rate": "15",
            "period": "second"
          }
        ]
```

I expected that I could have my custom annotation write extensions to swagger.json by calling `Extension` inside it, but that turned out not to be the case. I dug in and came up with this proposal:

I want to be able to configure extensions for custom middleware, e.g. 
```ts
const specAndRouteArgs: SwaggerArgs = {
  json: true,
  configuration: {
    spec: specConfig,
    routes: routesConfig,
    entryFile: "src/index.ts",
    customSwaggerExtensions: [
      {
        decoratorName: "RateLimit",
        name: "x-rate-limit-docs",
        value: (args) => {
          return [{ rate: args[0].getText(), period: args[1].getText() }];
        }
      },
    ],
  },
};
```
This would also create a work-around for https://github.com/lukeautry/tsoa/issues/771 https://github.com/lukeautry/tsoa/issues/1213 https://github.com/lukeautry/tsoa/issues/1520 etc. because you could write a function to dynamically resolve extensions

## Current Behavior

Currently there is no way to add extensions in swagger to all endpoints that have a certain decorator (in my case a custom `@RateLimit` decorator that wraps `Middlewares`)

```yaml

```

```ts

```

## Possible Solution

I think the best way to solve this problem is to add an optional customSwaggerExtension object to the config 

## Steps to Reproduce

N/A

## Context (Environment)

Version of the library: 
Version of NodeJS: 

## Detailed Description

For my service, I wrote a custom annotation called `@RateLimit` that uses `Middlewares` under the hood to rate limit API calls based on the passed in rate and key. I want these rate limits to also be reflected in the documentation, which requires adding an extension to the swagger.json exactly like the `@Extension` annotation does today. I don't want to have to annotate each of my API calls twice, once for the actual rate limit and once to document it.

In my proposal, users can optionally add a `customSwaggerExtensions` field to the tsoa configuration which will create an extension to go with a given annotation. The typescript is:
```ts
customSwaggerExtensions?: Array<
    {
      decoratorName: string,
      name: `x-${string}`,
      value: (args: ts.NodeArray<ts.Expression>) => Array<ExtensionType | ExtensionType[]>,
    }>
```

I am submitting a proof-of-concept PR in conjunction with the issue--would love any and all feedback, and am happy to work to make it meet the contribution standards once I know that you all are aligned with the philosophy

## Breaking change?
No

gwprice115:
### All Submissions:

More than happy to write tests and stuff once I get confirmation this is the right approach

**Closing issues**
closes #1532

**Test plan**

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -328,7 +328,7 @@ async function SpecGenerator(args: SwaggerArgs) {
     const compilerOptions = validateCompilerOptions(config.compilerOptions);
     const swaggerConfig = await validateSpecConfig(config);
 
-    await generateSpec(swaggerConfig, compilerOptions, config.ignore);
+    await generateSpec(swaggerConfig, compilerOptions, config.customSwaggerExtensions, config.ignore);
   } catch (err) {
     // eslint-disable-next-line no-console
     console.error('Generate swagger error.\n', err);
@@ -346,7 +346,7 @@ async function routeGenerator(args: ConfigArgs) {
     const compilerOptions = validateCompilerOptions(config.compilerOptions);
     const routesConfig = await validateRoutesConfig(config);
 
-    await generateRoutes(routesConfig, compilerOptions, config.ignore);
+    await generateRoutes(routesConfig, compilerOptions, config.customSwaggerExtensions, config.ignore);
   } catch (err) {
     // eslint-disable-next-line no-console
     console.error('Generate routes error.\n', err);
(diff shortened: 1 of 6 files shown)


TristanSpeakEasy:
Would love to see this go in will make our lives much easier

WoH:
This would probably require us to also be able to use a .tsoa.js as config file instead of  the tsoa.json, if you don't call generateSwagger directly, but use the cli. (And also live in the spec config)
Also, it should be bw-compat if not provided.

gwprice115:
> This would probably require us to also be able to use a .tsoa.js as config file instead of the tsoa.json, if you don't call generateSwagger directly, but use the cli. (And also live in the spec config) Also, it should be bw-compat if not provided.

gotcha makes sense--would you accept it if it was a static reference similar to how auth is specified?

WoH:
Yes

Issue #1550, opened 2024-01-29
 Nullable $refs should include "type": "object"


gwprice115:
## Sorting

- **I'm submitting a ...**

  - [X] bug report
  - [ ] feature request
  - [ ] support request

- I confirm that I
  - [X] used the [search](https://github.com/lukeautry/tsoa/search?type=Issues) to make sure that a similar issue hasn't already been submit

## Expected Behavior

Nullable $refs should include "type": "object"

https://github.com/openapistack/openapi-backend/issues/228
https://github.com/ajv-validator/ajv/issues/1882#issuecomment-1030159275

```yaml

```

```ts

```

## Current Behavior

For better compatibility with consumers, references should include "type": "object" when nullable is included (see referenced issues) 

```yaml

```

```ts

```

## Steps to Reproduce

1. 2. 3. 4.

## Context (Environment)

Version of the library: 
Version of NodeJS: 

- Confirm you were using yarn not npm: [X]

PR #1551, opened 2024-01-29
put type field for nullable

gwprice115:
### All Submissions:

**Closing issues**
closes https://github.com/lukeautry/tsoa/issues/1550

**Test plan**


gwprice115:
would love some confirmation that this will be accepted before i write the test

gwprice115:
i think this is correct but i've hacked around it locally so won't push for it

WoH:
I think the result is not correct here, esp/ since we should only put a type on the target of the ref nd only it it's nullable, right?

gwprice115:
I think you're right it's only needed if it's nullable, but putting it on the target of the ref won't solve the issues i linked in the Issue

PR #1564, opened 2024-02-16
fix: array validation should throw an error if the provided value is not an array

### LINKED ISSUE #1563: Array validation does not check if the provided value is an array
## Sorting

- **I'm submitting a ...**

  - [x] bug report
  - [ ] feature request
  - [ ] support request

- I confirm that I

## Expected Behavior

If a request body contains a parameter typed as an array, it should be validated that the provided value is actually an array.

## Current Behavior

If a request body contains a parameter typed as an array:
```ts
interface SomeRequestBody {
  some_parameter: string[];
}
```
You can actually provide a Request having the parameter set to the arrays generic type without the validation catching it:
```ts
// This won't return a validation error response but it should:
fetch('/some/endpoint', {
  method: 'POST',
  headers: {
    "Content-Type": "application/json",
  },
  body: JSON.stringify({
    // notice how we provide a string instead of a string array here
    some_parameter: "this should be invalid as it is a string and not an array"
  })
})
```

## Possible Solution

I'll provide a fix for this in the code :).

## Steps to Reproduce

See above

## Context (Environment)

Version of the library: v6.0.1
Version of NodeJS: v18.18.2

## Detailed Description

I'll use the Array.isArray() function to validate that the input to `validateArray()` is in fact an array. If not, a validation error will be produced by the function.

## Breaking change?

Nope :tada:

friedow:
### All Submissions:

- [x] Have you followed the guidelines in our [Contributing](https://github.com/lukeautry/tsoa/tree/master/docs/CONTRIBUTING.md) document? AFAIK yes :)
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/lukeautry/tsoa/pulls) for the same update/change? Yes
- [x] Have you written unit tests? Yup
- [x] Have you written unit tests that cover the negative cases (i.e.: if bad data is submitted, does the library respond properly)? Yup
- [x] This PR is associated with an existing issue? This closes #1563.

### DIFF
--- a/packages/runtime/src/routeGeneration/templateHelpers.ts
+++ b/packages/runtime/src/routeGeneration/templateHelpers.ts
@@ -417,7 +417,7 @@ export class ValidationService {
   }
 
   public validateArray(name: string, value: any[], fieldErrors: FieldErrors, swaggerConfig: AdditionalProps, schema?: TsoaRoute.PropertySchema, validators?: ArrayValidator, parent = '') {
-    if (!schema || value === undefined) {
+    if (!schema || !Array.isArray(value)) {
       const message = validators && validators.isArray && validators.isArray.errorMsg ? validators.isArray.errorMsg : `invalid array`;
       fieldErrors[parent + name] = {
         message,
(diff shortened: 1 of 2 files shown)


friedow:
Closing this in favor of #1578.

PR #1578, opened 2024-03-11
feat: make coercion of body values configurable

### LINKED ISSUE #1444: Null value for boolean attribute is converted to false. How to alter this behavior?
This is related to validation. Null values sent for attributes defined as boolean are converted to false.

## Sorting

- **I'm submitting a ...**

  - [ ] bug report
  - [x] feature request
  - [ ] support request

- I confirm that I
  - [x ] used the [search](https://github.com/lukeautry/tsoa/search?type=Issues) to make sure that a similar issue hasn't already been submit

## Current & Expected Behavior

When an attribute is defined as a boolean
For eg:
```ts
interface UserCreateRequest {
    preferredDelivery: boolean
}
```
and in payload/request body, if null is sent
For eg:
```ts
{
    "preferredDelivery": null
}
```

After the request passes through validation, the `preferredDelivery`'s value is observed to be false.
The expected behavior is that the null value should have been preserved.

## Possible Solution

I understand in some cases this is okay and desired, but is some cases this behavior is needed. And I couldn't find a way to alter this behavior.
The validator library (https://github.com/validatorjs/validator.js) seem to have `{ loose: false }` option but I could not find a way to use it here.
Any help or guidance on this?

### LINKED ISSUE #1563: Array validation does not check if the provided value is an array

friedow:
This will add a config flag to enable / disable coercion on request bodies. The default for body coercion is enabled for this not to become a breaking change. 

### All Submissions:

**Closing issues**

This fixes #1563 and it resolves #1444.

- [x] Has the issue had a maintainer respond to the issue and clarify that the feature is something that aligns with the [goals](https://github.com/lukeautry/tsoa#goal) and [philosophy](https://github.com/lukeautry/tsoa#philosophy) of the project? @WoH ;)


WoH:
Can you rebase this?

packages/runtime/src/routeGeneration/templateHelpers.ts
```diff
@@ -154,8 +164,8 @@ export class ValidationService {
     return value;
   }
 
-  public validateInt(name: string, value: any, fieldErrors: FieldErrors, validators?: IntegerValidator, parent = '') {
-    if (!validator.isInt(String(value))) {
+  public validateInt(name: string, value: any, fieldErrors: FieldErrors, isBodyParam: boolean, swaggerConfig: AdditionalProps, validators?: IntegerValidator, parent = '') {
+    if ((isBodyParam && swaggerConfig.bodyCoercion === false && typeof value !== 'number') || !validator.isInt(String(value))) {
```
WoH, line 168: Can you extract this to a fn?

WoH: At least the first 2 conditions into fn shouldCoerce, possibly with a callback for the third condition(s)

friedow: I've extracted a function, what do you think?

friedow: I've extracted this to a function, what do you think?

packages/runtime/src/routeGeneration/templateHelpers.ts
```diff
@@ -6,14 +6,23 @@ import { Tsoa } from '../metadataGeneration/tsoa';
...
+  fieldErrors: FieldErrors,
+  isBodyParam: boolean,
+  parent = '',
+  swaggerConfig: AdditionalProps,
+) {
+  return new ValidationService(generatedModels).ValidateParam(property, value, name, fieldErrors, isBodyParam, parent, swaggerConfig);
 }
 
 export class ValidationService {
   constructor(private readonly models: TsoaRoute.Models) {}
```
WoH, line 23: I am not sure, do we ever need to swap the minimalSwaggerConfig: AdditionalProps AND the isBodyParam? maybe we can init the ValidationService with them and avoid carrying the 2 new params around all the time.

friedow: We can probably move the `minimalSwaggerConfig` to be a property on the ValidationService because it won't change at runtime. The `isBodyParam` parameter though is tied closely to the value being validated. Thats why I think we need to keep this one as a parameter which is carried through. WDYT?

WoH: yeah, let's move minimalSwaggerConfig to the constructor (deep optional).
I just checked and yes, since we reuse the validation service, keep the isBody as arg to validate.

packages/runtime/src/routeGeneration/templateHelpers.ts
```diff
@@ -6,14 +6,23 @@ import { Tsoa } from '../metadataGeneration/tsoa';
...
+  property: TsoaRoute.PropertySchema,
+  value: any,
+  generatedModels: TsoaRoute.Models,
+  name = '',
+  fieldErrors: FieldErrors,
+  isBodyParam: boolean,
+  parent = '',
+  swaggerConfig: AdditionalProps,
+) {
+  return new ValidationService(generatedModels).ValidateParam(property, value, name, fieldErrors, isBodyParam, parent, swaggerConfig);
```
WoH, line 19:
```suggestion
  return new ValidationService(generatedModels, swaggerConfig).ValidateParam(property, value, name, fieldErrors, isBodyParam, parent);
```

packages/runtime/src/routeGeneration/templateHelpers.ts
friedow: I was in the midst of doing this change when I found these pieces of code:
- https://github.com/friedow/tsoa/blob/feat/configure-implicit-body-coercion/packages/runtime/src/routeGeneration/templateHelpers.ts#L603-L606
- https://github.com/friedow/tsoa/blob/feat/configure-implicit-body-coercion/packages/runtime/src/routeGeneration/templateHelpers.ts#L572-L575

This changes the swaggerConfig of validation at runtime. If I move this config to be a property on the `ValidationService` this would unfortunately become a breaking change :/.

packages/runtime/src/routeGeneration/templateHelpers.ts
friedow: See the comment above.

packages/runtime/src/routeGeneration/templateHelpers.ts
WoH: Theoretically, you could just run these valudations not on this, but on a new subValidationService.

friedow: I like this. Will refactor it like this :).

friedow: Done ;)

friedow:
I've rebased the branch. Any more comments @WoH or is this ready to be merged?

WoH:
Since this is a larger one, I haven't had time to go though it in detail yet, but the structure LGTM.
If there are no obvious things missing, I will merge this for a new minor probably this week.
@jackey8616 you have any notes?

jackey8616:
LGTM.

friedow:
:partying_face:

PR #1597, opened 2024-03-20
Fix/mapped generic type with date

### LINKED ISSUE #1596: Handling parameters that include mapped types with `Date` as a generic argument
An error is raised by invoking `npx tsoa spec-and-routes` when one of the endpoints within routers has a `@Body()` parameter that includes a mapped type field with `Date` as a generic argument. The error is introduced with the `tsoa 6.0.0` version. It was not present up until and including the `tsoa 5.1.1` version.

## Sorting

- **I'm submitting a ...**

  - [x] bug report
  - [ ] feature request
  - [ ] support request

- I confirm that I

## Expected Behavior

The following TypeScript code should be processed by `tsoa spec-and-routes` without any issue.

```ts
import { Body, Controller, Post, Route, Tags } from "tsoa";

interface Mapped<T> {
  array: T[];
}

interface DummyBody {
  field: DummyBodyField;
}

interface DummyBodyField {
  date: Mapped<Date>; // !!! date causes problem !!!
  string: Mapped<string>; // string is ok
  number: Mapped<number>; // number is ok
  boolean: Mapped<boolean>; // boolean is ok
}

@Route("dummy")
export class DummyController extends Controller {
  @Post()
  async dummy(@Body() dummyBody: DummyBody): Promise<void> {
    console.log(dummyBody);
  }
}
```

## Current Behavior

For a TypeScript code in the _Expected Behavior_ section, primitive types `string`, `number`, and `boolean` do not cause the error. I also think that custom simple JavaScript classes do not cause the error. Standard JavaScript `Date` type, on the other hand, seems to be causing an error.

The following error is returned by `tsoa spec-and-routes`.

```
There was a problem resolving type of 'DummyBodyField'.
There was a problem resolving type of 'DummyBody'.
Generate routes error.
 GenerateMetadataError: Cannot read properties of undefined (reading 'kind') 
 in 'DummyController.dummy'
    at /workspaces/workspace/node_modules/@tsoa/cli/dist/metadataGeneration/methodGenerator.js:104:23
    at Array.map (<anonymous>)
    at MethodGenerator.buildParameters (/workspaces/workspace/node_modules/@tsoa/cli/dist/metadataGeneration/methodGenerator.js:96:14)
    at MethodGenerator.Generate (/workspaces/workspace/node_modules/@tsoa/cli/dist/metadataGeneration/methodGenerator.js:66:33)
    at /workspaces/workspace/node_modules/@tsoa/cli/dist/metadataGeneration/controllerGenerator.js:46:41
    at Array.map (<anonymous>)
    at ControllerGenerator.buildMethods (/workspaces/workspace/node_modules/@tsoa/cli/dist/metadataGeneration/controllerGenerator.js:46:14)
    at ControllerGenerator.Generate (/workspaces/workspace/node_modules/@tsoa/cli/dist/metadataGeneration/controllerGenerator.js:35:27)
    at /workspaces/workspace/node_modules/@tsoa/cli/dist/metadataGeneration/metadataGenerator.js:210:41
    at Array.map (<anonymous>)
``` 

## Steps to Reproduce

The bug should be easily reproducible using the code supplied in the _Expected Behavior_ section.

There is also a devbox on [codesandbox.io](https://codesandbox.io/p/devbox/tsoa-issue-n7ty4s).

## Context (Environment)

In addition to the devbox where `yarn` was used, the error persist on the local machine where `npm` was used.

Version of the library: `6.1.5`
Version of NodeJS: `v20.5.0`

jackey8616:
### All Submissions:

**Closing issues**
Closes #1596 

I'm not sure this fix will effect other not covered cases.
Need folks to help me figure it out are there any possible missed cases.


jackey8616:
@WoH need a rerun. thx

packages/cli/src/metadataGeneration/typeResolver.ts
```diff
@@ -695,6 +695,11 @@ export class TypeResolver {
     } else {
       const declarations = this.getModelTypeDeclarations(type);
 
+      if (declarations.length === 0) {
+        // Escape some types like Date.
+        return name;
+      }
```
WoH, line 701: If we dont find them, we should just not support it. This may work by a lucky accident, but is probably a huge footgun.

jackey8616: Maybe we can just escape for `Date` ?

WoH:
Is this needed if the declaration is included? 
Usually we handle these high up in the type resolver (see: 'Promise').

I am generally very sceptical about these since they basically opt out of everything predictable.

jackey8616:
> Is this needed if the declaration is included? Usually we handle these high up in the type resolver (see: 'Promise').
> 
> I am generally very sceptical about these since they basically opt out of everything predictable.

I'm not sure the meaning of `declaration is included`, could you give me more detail about it?

The error is cause by generic type name cannot determine if given T is Date.  
Type resolve itself is working good.
Once the `calcRefTypeName` could name it, everything works.

WoH:
I am worried about any `<T extends Date>` where Date is not the one we now assume it is.

WoH:
I briefly debugged this. We have declarations for Date, 6 of them actually. The fix should be in how we get the right one, not papering over a fix with another fix.

jackey8616:
> I briefly debugged this. We have declarations for Date, 6 of them actually. The fix should be in how we get the right one, not papering over a fix with another fix.

I do noticed that there is 6 declarations (5 interface, 1 value).
The error is cause by `getModelTypeDeclarations` omitted all typescript native library which `Date` is also included it.

Here comes 2 solutions:
1. escape like this:
```ts
      modelTypes = modelTypes.filter(modelType => {
        // remove types that are from typescript e.g. 'Account'
        switch (typeName) {
          case 'Date':
            return modelType;
          default: {
            const srcFileName = modelType.getSourceFile().fileName;
            return srcFileName.replace(/\\/g, '/').toLowerCase().indexOf('node_modules/typescript') <= -1;
          }
        }
      });
```
2. remove the filter
I checked the definition of Date, found out that inside es2015, it seems nothing critical.
Removing the filter is ok for all test cases.

Do you still remember anything about this native escape?

WoH:
> > I briefly debugged this. We have declarations for Date, 6 of them actually. The fix should be in how we get the right one, not papering over a fix with another fix.
> 
> I do noticed that there is 6 declarations (5 interface, 1 value).
> The error is cause by `getModelTypeDeclarations` omitted all typescript native library which `Date` is also included it.
> 
> Here comes 2 solutions:
> 1. escape like this:
> ```ts
>       modelTypes = modelTypes.filter(modelType => {
>         // remove types that are from typescript e.g. 'Account'
>         switch (typeName) {
>           case 'Date':
>             return modelType;
>           default: {
>             const srcFileName = modelType.getSourceFile().fileName;
>             return srcFileName.replace(/\\/g, '/').toLowerCase().indexOf('node_modules/typescript') <= -1;
>           }
>         }
>       });
> ```
> 2. remove the filter
> I checked the definition of Date, found out that inside es2015, it seems nothing critical.
> Removing the filter is ok for all test cases.
> 
> Do you still remember anything about this native escape?

Iirc there's a lot of bs in there. Basically all environments, as well as type declarations TS uses. Maybe we can decide based on the lib that is used from tsconfig. Otherwise we'd end up including a lot.

jackey8616:
> Iirc there's a lot of bs in there. Basically all environments, as well as type declarations TS uses. Maybe we can decide based on the lib that is used from tsconfig. Otherwise we'd end up including a lot.

I'm not sure should we keep this line or not.
I noticed that this line is exists at very beginning and add by Luke, and I've no idea why he wants to escape some type called Account like he commented with `// remove types that are from typescript e.g. 'Account'`.
Also, there is no type named `Account` in es5 ~ es2023. 

Shall we try remove this?

veiper93:
Is there any indication of whether and when this pull request will be merged?

jackey8616:
@WoH Any news?

WoH:
I have no clue, but removing and bumping a major version seems reasonable

veiper93:
Will this be reopened and hopefully merged, or is it closed for good?

GuidoGeo:
is there a change that there is a fix for this? in version 6.4 it still throws an error.

PR #1602, opened 2024-03-24
Await apiHandler to handle error in try/catch

### LINKED ISSUE #1598: ErrorHandlerMiddleware is ignore with 6.1.*
## Sorting

- **I'm submitting a ...**

  - [x] bug report
  - [ ] feature request
  - [ ] support request

- I confirm that I

## Expected Behavior

When an error is thrown, custom ErrorHandlerMiddleware catches the error and returns a response with according errorCode (400, 403, 429, etc.). Works as expected with 6.0.*

## Current Behavior

When an error is thrown, custom ErrorHandlerMiddleware is ignore and app hangs up, nothing happens. Breaks with upgrade to 6.1.4 and 6.1.5, and routes regenerated.

## Context (Environment)

Version of the library: 6.1.*
Version of NodeJS: 20.*

## Breaking change?

It seems that is related to the new way of generated routes that handles requests.
I have passed many hours to debug, but does not find something really more helpfull. Sorry !

VincentClair:
### All Submissions:

**Closing issues**

closes #1598

**Test plan**

Checks that a controller method without `async` keyword throwing an error is caught.

The real problem is if method is not a promise (`async` keyword), and error throw immediately when building promise, server is no more responding, and we get a time out (actually 5sec in tests).

The problem could resolve in two ways : 
- put the `buildPromise` int the try/catch  https://github.com/lukeautry/tsoa/blob/efdb468a5cedcb9bef345f1eed776e8fc621b443/packages/runtime/src/routeGeneration/templates/express/expressTemplateService.ts#L41
- wait for `apiHandler` to be resolved on generated template https://github.com/lukeautry/tsoa/blob/efdb468a5cedcb9bef345f1eed776e8fc621b443/packages/cli/src/routeGeneration/templates/express.hbs#L106
I have chosen to apply both, as I think code will be more robust.

### DIFF
--- a/packages/cli/src/routeGeneration/templates/express.hbs
+++ b/packages/cli/src/routeGeneration/templates/express.hbs
@@ -103,7 +103,7 @@ export function RegisterRoutes(app: Router) {
                 const controller = new {{../name}}();
               {{/if}}
 
-              templateService.apiHandler({
+              {{#if ../../iocModule}}await {{/if}}templateService.apiHandler({
                 methodName: '{{name}}',
                 controller,
                 response,
--- a/packages/cli/src/routeGeneration/templates/express.hbs
+++ b/packages/cli/src/routeGeneration/templates/express.hbs
@@ -79,7 +79,7 @@ export function RegisterRoutes(app: Router) {
             ...(fetchMiddlewares<RequestHandler>({{../name}})),
             ...(fetchMiddlewares<RequestHandler>({{../name}}.prototype.{{name}})),
 
-            {{#if ../../iocModule}}async {{/if}}function {{../name}}_{{name}}(request: ExRequest, response: ExResponse, next: any) {
+            async function {{../name}}_{{name}}(request: ExRequest, response: ExResponse, next: any) {
             const args: Record<string, TsoaRoute.ParameterSchema> = {
                 {{#each parameters}}
                     {{@key}}: {{{json this}}},
@@ -103,7 +103,7 @@ export function RegisterRoutes(app: Router) {
                 const controller = new {{../name}}();
               {{/if}}
 
-              {{#if ../../iocModule}}await {{/if}}templateService.apiHandler({
+              await templateService.apiHandler({
                 methodName: '{{name}}',
                 controller,
                 response,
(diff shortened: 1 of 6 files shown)


packages/cli/src/routeGeneration/templates/express.hbs
```diff
@@ -103,7 +103,7 @@ export function RegisterRoutes(app: Router) {
                 const controller = new {{../name}}();
               {{/if}}
 
-              templateService.apiHandler({
```
WoH, line 106: I think we can always await, making the apiHandler return A | Promise<A>

VincentClair: agree, but function must be always async: we could replace condition `{{#if ../../iocModule}}async {{/if}}`line 82 by async directly.

packages/runtime/src/routeGeneration/templates/express/expressTemplateService.ts
```diff
@@ -38,9 +38,9 @@ export class ExpressTemplateService extends TemplateService<ExpressApiHandlerPar
 
   async apiHandler(params: ExpressApiHandlerParameters) {
     const { methodName, controller, response, validatedArgs, successStatus, next } = params;
-    const promise = this.buildPromise(methodName, controller, validatedArgs);
 
     try {
+      const promise = this.buildPromise(methodName, controller, validatedArgs);
       const data = await Promise.resolve(promise);
```
VincentClair, line 44: is Promise.resolve mandatory : promise ca be waited directly, no? `const data = await promise;`

packages/cli/src/routeGeneration/templates/express.hbs
WoH: Yeah, that was the ultimate goal, while still making ts not complain about it.

jackey8616: > I think we can always await @jackey8616

Yes, I think that's good.

VincentClair: done

VincentClair:
Hello @WoH @jackey8616,
How can i help to validate this PR ? Is there something missing ?

WoH (approved):
No, lgtm 🚀

denizdogan:
When can we expect a new release with this fix included? Much appreciated

PR #1606, opened 2024-04-03
Refactor(resolver): clear logics

jackey8616:
- Splitting logics of TypeResolver into Transformer class.
- Clean up some nested logics in TypeResolver.
- Rewrite the flow of throwing error with throwUnless instead of if branch.

### DIFF
--- a/packages/cli/src/metadataGeneration/transformer/dateTransformer.ts
+++ b/packages/cli/src/metadataGeneration/transformer/dateTransformer.ts
@@ -1,11 +1,11 @@
-import * as ts from 'typescript';
+import type { Node } from 'typescript';
 import { Tsoa } from '@tsoa/runtime';
 
 import { Transformer } from './transformer';
 import { getJSDocTagNames } from '../../utils/jsDocUtils';
 
 export class DateTransformer extends Transformer {
-  public transform(parentNode?: ts.Node): Tsoa.DateType | Tsoa.DateTimeType {
+  public transform(parentNode?: Node): Tsoa.DateType | Tsoa.DateTimeType {
     if (!parentNode) {
       return { dataType: 'datetime' };
     }
--- a/packages/cli/src/metadataGeneration/transformer/dateTransformer.ts
+++ b/packages/cli/src/metadataGeneration/transformer/dateTransformer.ts
@@ -1,0 +1,28 @@
+import * as ts from 'typescript';
+import { Tsoa } from '@tsoa/runtime';
+
+import { Transformer } from './transformer';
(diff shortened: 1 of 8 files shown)


packages/cli/src/metadataGeneration/transformer/transformer.ts
```diff
@@ -0,0 +1,50 @@
+import * as ts from 'typescript';
+
+import { TypeResolver } from '../typeResolver';
+
+/**
+ * Transformer responsible to transforming native ts node into tsoa type.
+ */
+export abstract class Transformer {
```
WoH, line 8: What do you think about the Transformers registering themselves via a shouldHandle(node) function?

jackey8616: Can you give me more details about it? I have no clue about shouldHandle function. Oops.

jackey8616:
@WoH 
I think we should refactor it in pieces.
This PR would only includes these commits.

packages/cli/src/metadataGeneration/transformer/dateTransformer.ts
```diff
@@ -0,0 +1,28 @@
+import * as ts from 'typescript';
+import { Tsoa } from '@tsoa/runtime';
```
WoH, line 2: Note for the future, we should probably remove this namespace and use a module.

packages/cli/src/metadataGeneration/transformer/dateTransformer.ts
```diff
@@ -0,0 +1,28 @@
+import * as ts from 'typescript';
```
WoH, line 1: Please use this as an opportunity to avoid the star import in the new files.

packages/cli/src/metadataGeneration/transformer/dateTransformer.ts
jackey8616: Once we decouple everything from GIGANTIC type resolver, I think we can do this.

packages/cli/src/metadataGeneration/transformer/transformer.ts
```diff
@@ -0,0 +1,51 @@
+import type { HasModifiers } from 'typescript';
+import { SyntaxKind, getModifiers } from 'typescript';
```
WoH, line 2:
```suggestion
import { SyntaxKind, getModifiers, type HasModifiers } from 'typescript';
```

Weird that Eslint doesn't handle this.

WoH (approved):

Issue #1619, opened 2024-04-29
Using combination of @UploadedFile() and @UploadedFiles() decorators throws MulterError: Unexpected field


Yoronex:
## Sorting

- **I'm submitting a ...**

    - [x] bug report
    - [ ] feature request
    - [ ] support request

- I confirm that I
    - [x] used the [search](https://github.com/lukeautry/tsoa/search?type=Issues) to make sure that a similar issue hasn't already been submit

## Expected Behavior

The docs about [file uploads](https://tsoa-community.github.io/docs/file-upload.html) suggest that it is possible to use a combination of `@UploadedFiles()` and `@UploadedFile()` in the example.

```ts
  @Post("uploadFile")
  public async uploadFile(
      @FormField() title: string,
      @FormField() description: string,
      @UploadedFiles() files: Express.Multer.File[],
      @UploadedFile() file: Express.Multer.File,
  ): Promise<void> {
    console.log(files);
  }
```

## Current Behavior

A multer error is thrown and no HTTP response is sent to the client (in Swagger UI, the endpoint loads forever).

```
MulterError: Unexpected field
    at wrappedFileFilter (C:\Users\roy.kakkenberg\WebstormProjects\arvisan-backend\node_modules\multer\index.js:40:19)
    at Multipart.<anonymous> (C:\Users\roy.kakkenberg\WebstormProjects\arvisan-backend\node_modules\multer\lib\make-middleware.js:107:7)
    at Multipart.emit (node:events:514:28)
    at Multipart.emit (node:domain:488:12)
    at HeaderParser.cb (C:\Users\roy.kakkenberg\WebstormProjects\arvisan-backend\node_modules\busboy\lib\types\multipart.js:358:14)
    at HeaderParser.push (C:\Users\roy.kakkenberg\WebstormProjects\arvisan-backend\node_modules\busboy\lib\types\multipart.js:162:20)
    at SBMH.ssCb [as _cb] (C:\Users\roy.kakkenberg\WebstormProjects\arvisan-backend\node_modules\busboy\lib\types\multipart.js:394:37)
    at feed (C:\Users\roy.kakkenberg\WebstormProjects\arvisan-backend\node_modules\streamsearch\lib\sbmh.js:248:10)
    at SBMH.push (C:\Users\roy.kakkenberg\WebstormProjects\arvisan-backend\node_modules\streamsearch\lib\sbmh.js:104:16)
[...]

PR #1620, opened 2024-04-29
Fix Multer error when @UploadedFiles() is used twice in a route

Yoronex:
### All Submissions:

**Closing issues**

Closes https://github.com/lukeautry/tsoa/issues/1619.

There are not that many files present in the test folder to use for testing uploading files. Right now the same file is included multiple times, but the order is always different, so this should not matter.

The `routeGenerator.ts` now also adds two extra attributes to each file upload field, namely `maxCount` and `multiple`. The `maxCount` attribute is used by Multer to enforce only one file is uploaded to that field. However, HandleBars aims to be logic free, so a simple check (`maxCount == 1`) cannot be done in `hapi.hbs` to determine whether multiple files can be uploaded or not. Therefore, the `multiple` attribute is also added, but by design of the `json` function in Handlebars, this extra variable is also included in the Express and Koa `routes.ts`. Luckily, it does not do anything, but it is not very elegant. If there is a better way to do this, I would love to hear it!

**Test plan**

The unit tests that I added includes the case mentioned in the issue, but also two `@UploadedFiles()` statement to truly verify that these arrays are separated correctly.

The fix simply removes the `upload.array()` lines from any `routes.ts` files, because the same can be achieved by using `upload.fields()`.

### DIFF
--- a/packages/cli/src/routeGeneration/routeGenerator.ts
+++ b/packages/cli/src/routeGeneration/routeGenerator.ts
@@ -97,8 +97,9 @@ export abstract class AbstractRouteGenerator<Config extends ExtendedRoutesConfig
 
             const normalisedFullPath = normalisePath(`${normalisedBasePath}${normalisedControllerPath}${normalisedMethodPath}`, '/', '', false);
 
-            const uploadFilesWithDifferentFieldParameter = method.parameters.filter(parameter => parameter.type.dataType === 'file');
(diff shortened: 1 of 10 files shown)


WoH:
Sorry for the long delay, lgtm

douglasg14b:
Might be a regression here? https://github.com/lukeautry/tsoa/issues/1642

PR #1662, opened 2024-08-21
Fix typescript build error when using file upload

### LINKED ISSUE #1642: tsc build error with file uploads
When using either `@UploadedFile` or `@UploadedFiles` to configure a form post to do a file upload the built `routes.ts` file contains invalid code.

## Sorting

- **I'm submitting a ...**

  - [x] bug report
  - [ ] feature request
  - [ ] support request

- I confirm that I

## Expected Behavior

A handler with a file upload should stil compile to valid typescript which is compatible with multer type definitions:

```ts
  @Post('/')
  public async upload(
    @UploadedFile() file?: Express.Multer.File
  ): Promise<Attachment> { ... }
```

## Current Behavior

The built `routes.ts` is not valid against the multer type definitions. An excess property `multiple` is populated on an object literal that is not defined on the `multer.Multer.Field`

## Possible Solution

Remove this line https://github.com/lukeautry/tsoa/blame/d4a33015a7c0671cae8c24aae5afa3ea1fe44064/packages/cli/src/routeGeneration/routeGenerator.ts#L113 as it doesn't seem to add anything multer expects?

## Steps to Reproduce

We've hit this on all our repos that do file upload. https://github.com/digicatapult/sqnc-matchmaker-api/pull/385 is one example.

You can also just create a handler as described in the expected behaviour section above

## Context (Environment)

Version of the library: v6.3.1
Version of NodeJS: v20.x.x
Version of Typescript: v5.4.5
Version of Multer: 1.4.5-lts.1

## Detailed Description

When using either `@UploadedFile` or `@UploadedFiles` to configure a form post to do a file upload the built `routes.ts` contains
in the request handler the code:

```ts
upload.fields([{"name":"file","maxCount":1,"multiple":false}]),
```

The problem with this is the type of `multer.Multer.Field` doesn't actually contain a property `multiple` [see types package](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/f38be174e67baab75c611dc0588340b1d7151f4c/types/multer/index.d.ts#L308). Typescript therefore throw a build error:

```
src/routes.ts:1571:56 - error TS2353: Object literal may only specify known properties, and '"multiple"' does not exist in type 'Field'.

1571             upload.fields([{"name":"file","maxCount":1,"multiple":false}]),
```

I came across this [during an upgrade](https://github.com/digicatapult/sqnc-matchmaker-api/pull/385) to tsoa v6.3.0 but it also applied to the latest v6.3.1. This change was introduced in https://github.com/lukeautry/tsoa/pull/1620

Honestly maybe I'm missing something obvious as that field was clearly added in the above PR with intent. Looking at `multer`'s source code however I cannot see anything that references a property `multiple` on a field.

## Breaking change?

N/A

mattdean-digicatapult:
This PR fixes an issue where the generated `routes.ts` for an `express` or `koa` based project has invalid usage of the `multer` `upload.fields` method when uploading files. This is because  of an extra rendered field `multiple` which is used in the `hapi` template. Note I have also removed some conditional rendering when the property `uploadFiles` is present on the `action` as this is now never generated.

Note I think the best way to unit test this would be to enable type checking in the `tsc` build of usages against external libraries. This has very clearly been disabled intensionallty so I've left it as is, hence no tests in the checklist.

### All Submissions:

**Closing issues**

closes #1642

### DIFF
--- a/packages/cli/src/routeGeneration/templates/express.hbs
+++ b/packages/cli/src/routeGeneration/templates/express.hbs
@@ -85,10 +85,16 @@ export function RegisterRoutes(app: Router) {
             authenticateMiddleware({{json security}}),
             {{/if}}
             {{#if uploadFile}}
-            upload.fields({{json uploadFileName}}),
-            {{/if}}
-            {{#if uploadFiles}}
-            upload.array('{{uploadFilesName}}'),
+            upload.fields([
+                {{#each uploadFileName}}
+                {
+                    name: {{json name}},
+                    {{#if maxCount}}
+                    maxCount: {{maxCount}}
+                    {{/if}}
+                }{{#if @last}}{{else}},{{/if}}
+                {{/each}}
+            ]),
             {{/if}}
             ...(fetchMiddlewares<RequestHandler>({{../name}})),
             ...(fetchMiddlewares<RequestHandler>({{../name}}.prototype.{{name}})),
--- a/packages/cli/src/routeGeneration/templates/hapi.hbs
+++ b/packages/cli/src/routeGeneration/templates/hapi.hbs
@@ -89,13 +89,6 @@ export function RegisterRoutes(server: any) {
                   multipart: true,
                   allow: 'multipart/form-data'
                 },
-                {{else if uploadFiles}}
-                payload: {
-                  output: 'stream',
-                  parse: true,
-                  multipart: true,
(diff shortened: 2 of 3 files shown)


WoH (approved):

ClementValot:
Hi 👋 

Is this planned for a release soon?

PR #1676, opened 2024-09-09
Use type import for TsoaRoute to ensure with verbatumModuleSyntax

### LINKED ISSUE #1418: Set verbatimModuleSyntax to true
Pretty much #1044, but for the newer, slightly stricter flag.

## Sorting

- **I'm submitting a ...**

  - [ ] bug report
  - [x] feature request
  - [ ] support request

- I confirm that I

## Expected Behavior

```ts
// routes.ts
import type { FieldErrors, TsoaRoute, HttpStatusCodeLiteral, TsoaResponse } from '@tsoa/runtime';
import { Controller, ValidationService, ValidateError, fetchMiddlewares } from '@tsoa/runtime';
```

## Current Behavior

```ts
// routes.ts
import { Controller, ValidationService, FieldErrors, ValidateError, TsoaRoute, HttpStatusCodeLiteral, TsoaResponse, fetchMiddlewares } from '@tsoa/runtime';
```

## Possible Solution

Changing this line:

https://github.com/lukeautry/tsoa/blob/master/packages/cli/src/routeGeneration/templates/express.hbs#L4
https://github.com/lukeautry/tsoa/blob/master/packages/cli/src/routeGeneration/templates/hapi.hbs#L4
https://github.com/lukeautry/tsoa/blob/master/packages/cli/src/routeGeneration/templates/koa.hbs#L5

## Detailed Description

This stricter version requries *all* types to be explicitly imported as such. With `importsNotUsedAsValues` you could include type imports with your "normal" imports, but with `verbatimModuleSyntax`, this is no longer allowed.

For example:

```ts
// not allowed by neither `importsNotUsedAsValues` nor `verbatimModuleSyntax`
import { MyInterface } from 'whatever';
```

```ts
// fine with `importsNotUsedAsValues`, not allowed by `verbatimModuleSyntax`
import { someFunction, MyInterface } from 'whatever';
```

```ts
// fine with both
import { someFunction, type MyInterface } from 'whatever';
```

```ts
// fine with both
import { someFunction } from 'whatever';
import type { MyInterface } from 'whatever';
```

So just like #1044 before, this will require a small tweak to the handlebars template(s).

## Breaking change?

No

christiango:
### All Submissions:

**Closing issues**

closes #1418 

We had two options here: 

* Use inline type imports like import { type Foo } from 'bar'
* Use import type { Foo } from 'bar'

While the inline type imports look cleaner, they are only supported starting in TypeScript 4.5, so I used the syntax that works back to TypeScript 3.8 since that's what's already used in the repo. For my use case, I can use either approach, but I didn't want to impact any other users on Typescript < 4.5

**Test plan**

I patched my node_modules in my downstream project that uses TSOA and confirmed I no longer get compiler errors due to the correct import!

### DIFF
--- a/packages/cli/src/routeGeneration/templates/express.hbs
+++ b/packages/cli/src/routeGeneration/templates/express.hbs
@@ -1,7 +1,8 @@
 /* tslint:disable */
 /* eslint-disable */
 // WARNING: This file was auto-generated with tsoa. Please do not modify it. Re-run tsoa to re-generate this file: https://github.com/lukeautry/tsoa
-import { TsoaRoute, fetchMiddlewares, ExpressTemplateService } from '@tsoa/runtime';
+import type { TsoaRoute } from '@tsoa/runtime';
+import {  fetchMiddlewares, ExpressTemplateService } from '@tsoa/runtime';
 {{#each controllers}}
 // WARNING: This file was auto-generated with tsoa. Please do not modify it. Re-run tsoa to re-generate this file: https://github.com/lukeautry/tsoa
 import { {{name}} } from '{{modulePath}}';
--- a/packages/cli/src/routeGeneration/templates/hapi.hbs
+++ b/packages/cli/src/routeGeneration/templates/hapi.hbs
@@ -1,7 +1,8 @@
 /* tslint:disable */
 /* eslint-disable */
 // WARNING: This file was auto-generated with tsoa. Please do not modify it. Re-run tsoa to re-generate this file: https://github.com/lukeautry/tsoa
-import { TsoaRoute, fetchMiddlewares, HapiTemplateService } from '@tsoa/runtime';
+import type { TsoaRoute } from '@tsoa/runtime';
+import { fetchMiddlewares, HapiTemplateService } from '@tsoa/runtime';
 {{#each controllers}}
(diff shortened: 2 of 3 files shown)


WoH:
What do you think about keeping it in one line, `import { type T, t } from 't'`?

christiango:
> What do you think about keeping it in one line, `import { type T, t } from 't'`?
> 

Hey, I included some details in the description as to why I didn't go down that route. While it doesn't impact my project, doing this would possibly change the minimum version requirement for typescript, so I aligned with existing usages to keep this compatible with older typescript versions.

I'm open to either approach personally, but I think this is the one least likely to cause issues for anyone

WoH:
> > What do you think about keeping it in one line, `import { type T, t } from 't'`?
> > 
> 
> Hey, I included some details in the description as to why I didn't go down that route. While it doesn't impact my project, doing this would possibly change the minimum version requirement for typescript, so I aligned with existing usages to keep this compatible with older typescript versions.
> 
> I'm open to either approach personally, but I think this is the one least likely to cause issues for anyone 

Sorry, I'm blind. In my defense, GH mobile hides it though.
We currently need 4.1+ for template literal types, but I agree that pushing it to 4.5 for little difference is probably not worth it.

WoH (approved):

PR #1692, opened 2024-10-02
perf: avoid useless create new empty object

cesco69:
### All Submissions:

**Closing issues**

**Test plan**

### DIFF
--- a/packages/runtime/src/routeGeneration/templates/express/expressTemplateService.ts
+++ b/packages/runtime/src/routeGeneration/templates/express/expressTemplateService.ts
@@ -103,8 +103,7 @@ export class ExpressTemplateService extends TemplateService<ExpressApiHandlerPar
   }
 
   protected returnHandler(params: ExpressReturnHandlerParameters) {
-    const { response, statusCode, data } = params;
-    const { headers } = params;
+    const { response, statusCode, data, headers } = params;
 
     if (response.headersSent) {
       return;
--- a/packages/runtime/src/routeGeneration/templates/express/expressTemplateService.ts
+++ b/packages/runtime/src/routeGeneration/templates/express/expressTemplateService.ts
@@ -105,14 +105,15 @@ export class ExpressTemplateService extends TemplateService<ExpressApiHandlerPar
   protected returnHandler(params: ExpressReturnHandlerParameters) {
     const { response, statusCode, data } = params;
     let { headers } = params;
-    headers = headers || {};
 
     if (response.headersSent) {
       return;
     }
-    Object.keys(headers).forEach((name: string) => {
-      response.set(name, headers[name]);
-    });
+    if( headers ){
+      Object.keys(headers).forEach((name: string) => {
+        response.set(name, headers[name]);
+      });
+    }
     if (data && typeof data.pipe === 'function' && data.readable && typeof data._read === 'function') {
       response.status(statusCode || 200);
       data.pipe(response);
--- a/packages/runtime/src/routeGeneration/templates/express/expressTemplateService.ts
+++ b/packages/runtime/src/routeGeneration/templates/express/expressTemplateService.ts
@@ -108,11 +108,11 @@ export class ExpressTemplateService extends TemplateService<ExpressApiHandlerPar
     if (response.headersSent) {
       return;
     }
-    if( headers ){
-      Object.keys(headers).forEach((name: string) => {
-        response.set(name, headers[name]);
-      });
-    }
+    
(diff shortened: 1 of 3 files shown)


jackey8616 (commented):
I think there is more then one places use this syntax like hapi & koa template service.
Can you make it right in once?

cesco69:
> I think there is more then one places use this syntax like hapi & koa template service. Can you make it right in once?

done

packages/runtime/src/routeGeneration/templates/express/expressTemplateService.ts
```diff
@@ -104,15 +104,16 @@ export class ExpressTemplateService extends TemplateService<ExpressApiHandlerPar
 
   protected returnHandler(params: ExpressReturnHandlerParameters) {
     const { response, statusCode, data } = params;
-    let { headers } = params;
-    headers = headers || {};
+    const { headers } = params;
```
jackey8616, line 107: I think L106, 107 can combine together like
`const { response, statusCode, data, headers } = params;

packages/runtime/src/routeGeneration/templates/express/expressTemplateService.ts
```diff
@@ -104,15 +104,16 @@ export class ExpressTemplateService extends TemplateService<ExpressApiHandlerPar
...
       return;
     }
-    Object.keys(headers).forEach((name: string) => {
-      response.set(name, headers[name]);
-    });
+    if( headers ){
+      Object.keys(headers).forEach((name: string) => {
+        response.set(name, headers[name]);
+      });
+    }
```
jackey8616, line 116: This if statement is also can combine with or operator like:
Object.keys(headers || {}).forEach ...

packages/runtime/src/routeGeneration/templates/express/expressTemplateService.ts
cesco69: yep

packages/runtime/src/routeGeneration/templates/express/expressTemplateService.ts
cesco69: I want avoid unecessary creation of new empty object

cesco69 (commented):
done

jackey8616:
https://github.com/lukeautry/tsoa/pull/1692#discussion_r1786227992
Can I have reason?

jackey8616:
> > [#1692 (comment)](https://github.com/lukeautry/tsoa/pull/1692#discussion_r1786227992) Can I have reason?
> 
> This version:
> 
> ```js
> if (headers) {
>   Object.keys(headers).forEach((name: string) => {
>     response.set(name, headers[name]);
>   });
> }
> ```
> 
> is more performant than:
> 
> ```js
> Object.keys(headers || {}).forEach((name: string) => { 
>   response.set(name, headers[name]);
> });
> ```
> 
> for the following reasons:
> 
> **Unnecessary Operations**: In the second version, `Object.keys(headers || {})` is always called, even when headers is null or undefined. If headers is falsy, || {} creates an empty object (`{}`), and then Object.keys processes it. This results in a redundant operation when headers is falsy, as there are no keys to iterate over.
> 
> In contrast, the first version avoids calling Object.keys altogether if headers is falsy, saving the overhead of creating an empty object and calling Object.keys.
> 
> **Control Flow Simplicity**: The first version has a clear conditional check (if (headers)), and only when headers is truthy does the code proceed to iterate through the keys. This makes it easier for the JavaScript engine to optimize the code because there's no need to run operations when they are unnecessary.
> 
> **Reduced Memory Usage**: In the second version, when headers is falsy, the code creates a new empty object (`{}`) that only serves to satisfy the Object.keys call. Although the memory impact of creating an empty object is small, it is still a waste when compared to the first version, which simply skips this step when not needed.

logically, `header` should not include the `undefined` or `null` value.
It should be represents empty with `{}` empty object.

In this situation, examine its existence is a logically wrong.
the why it design like version 1 is a historical reason, at the time of refactoring it I just simply have no time to think deeper but just trying to make it easier when reading, that's why I didn't wrap with anoterh if statement.

Since you raise the clean code issue, 
Let take the "more clean" solutions that we deal with the root cause: caller need to correctly pass the parameter without any incorrect type.
If we do so, we can save a if statement, stay clean.

cesco69:
done

jackey8616 (changes requested):
I'm not sure about these changes.
I thought you are going to add implicitly type on header.

PR #1701, opened 2024-10-13
refactor(specGenerator)/remove request-prop param

### LINKED ISSUE #1663: @RequestProp decorator adds a request-prop parameter in generated Spec
## Sorting

- **I'm submitting a ...**

  - [x] bug report
  - [ ] feature request
  - [ ] support request

- I confirm that I

## Expected Behavior
Using @RequestProp should not make any changes to the spec but should still work as expected. The request property in question is not a parameter that we expect users of our API to actually supply, so I'm not sure it makes sense to publish in the OpenAPI spec.

At the minimum, we should be able to toggle whether the @RequestProp makes it into the spec or not.

## Current Behavior

When using @RequestProp, it adds the decorated param and its type to the spec. It appears as a parameter for the related endpoint like so 

```yaml
      parameters:
        - in: request-prop
          name: propKey
```

However, this doesn't seem to be valid OpenAPI since `request-prop` is not a [supported parameter.in value](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#parameterIn)

## Possible Solutions
I think any of the following would work:

1) don't write the request-prop to the spec
2) allow the spec-writing behavior to be toggleable
3) allow the hidden decorator to be applied to a @RequestProp parameter

## Steps to Reproduce
```ts

type UserObject = { id: string; }

@Route("my-route")
public MyController extends Controller {
  
  @Get("")
  public async myEndpoint(
    @RequestProp("user") user: UserObject
  ) {
  }
}
```

chappelo:
### All Submissions:

**Closing issues**
closes #1663

adjusted spec generators to match `in` params (removed request-prop)

https://github.com/OAI/OpenAPI-Specification/blob/main/versions/2.0.md
https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.0.md

**Test plan**

added test cases for both SpecGenerator2 and SpecGenerator3 to ensure request-props aren't added to the spec.

### DIFF
--- a/packages/cli/src/swagger/specGenerator2.ts
+++ b/packages/cli/src/swagger/specGenerator2.ts
@@ -192,7 +192,7 @@ export class SpecGenerator2 extends SpecGenerator {
 
     pathMethod.parameters = method.parameters
       .filter(p => {
-        return ['request', 'body-prop', 'res', 'queries'].indexOf(p.in) === -1;
+        return ['request', 'body-prop', 'res', 'queries', 'request-prop'].indexOf(p.in) === -1;
       })
       .map(p => this.buildParameter(p));
 
--- a/packages/cli/src/swagger/specGenerator3.ts
+++ b/packages/cli/src/swagger/specGenerator3.ts
@@ -282,7 +282,7 @@ export class SpecGenerator3 extends SpecGenerator {
 
     pathMethod.parameters = method.parameters
       .filter(p => {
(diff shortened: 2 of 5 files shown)


jackey8616:
@chappelo 
Can you provide a positive test cases covers the `allowed` in value of parameters?
Both 2.0 and 3.0 OAI.

We should take this chances to cover all possible values.

jackey8616 (approved):
LGTM.

PR #1708, opened 2024-10-24
perf: skip validation on default value

cesco69:
performance improvement: if the provided value is the default value, no need to validate

### DIFF
--- a/packages/runtime/src/routeGeneration/templateHelpers.ts
+++ b/packages/runtime/src/routeGeneration/templateHelpers.ts
@@ -53,7 +53,10 @@ export class ValidationService {
         return value;
       }
     }
-
+    // performance improvement: if the provided value is the default value, no need to validate
+    if( property.default === value ){
+        return value;
+    }
     switch (property.dataType) {
       case 'string':
         return this.validateString(name, value, fieldErrors, property.validators as StringValidator, parent);
--- a/packages/runtime/src/routeGeneration/templateHelpers.ts
+++ b/packages/runtime/src/routeGeneration/templateHelpers.ts
@@ -54,7 +54,7 @@ export class ValidationService {
       }
     }
     // performance improvement: if the provided value is the default value, no need to validate
-    if( property.default === value ){
+    if(property.default === value){
         return value;
     }
     switch (property.dataType) {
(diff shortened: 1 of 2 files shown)


cesco69:
Are these test correct?
https://github.com/lukeautry/tsoa/blob/master/tests/unit/swagger/templateHelpers.spec.ts#L328
https://github.com/cesco69/tsoa/blob/patch-3/tests/unit/swagger/templateHelpers.spec.ts#L354

In both cases, set default value as string with type integer?

WoH:
> Are these test correct? https://github.com/lukeautry/tsoa/blob/master/tests/unit/swagger/templateHelpers.spec.ts#L328 https://github.com/lukeautry/tsoa/blob/master/tests/unit/swagger/templateHelpers.spec.ts#L354
> 
> In both cases, set default value as string with type integer?
> 
> ```
> { dataType: 'integer', default: '666' }
> ```
> 
> maybe
> 
> ```
> { dataType: 'integer', default: 666 }
> ```

They should be, I think since we get them as strings Code, we probably never coerce them to the target data type when serializing the spec.

tests/unit/swagger/templateHelpers.spec.ts
```diff
@@ -325,7 +325,7 @@ describe('ValidationService', () => {
   describe('Param validate', () => {
     it('should apply defaults for optional properties', () => {
       const value = undefined;
-      const propertySchema: TsoaRoute.PropertySchema = { dataType: 'integer', default: '666', required: false, validators: {} };
+      const propertySchema: TsoaRoute.PropertySchema = { dataType: 'integer', default: 666, required: false, validators: {} };
```
WoH, line 328: Can you revert these changes? At least while we don't check/fix when creating the metadata for validation.

packages/runtime/src/routeGeneration/templateHelpers.ts
```diff
@@ -53,7 +53,10 @@ export class ValidationService {
         return value;
       }
     }
-
+    // performance improvement: if the provided value is the default value, no need to validate
+    if( property.default === value ){
```
WoH, line 57:
```suggestion
    if(property.default === value){
```

WoH:
Nitpicks, otherwise LGTM

tests/unit/swagger/templateHelpers.spec.ts
cesco69: If revert it, test fails... it's ok?

WoH: The problem then is that this is the schema with the default being a string is what we would generate "in the real world".

cesco69: ah ok, so this PR can't work!

Issue #1256, opened 2022-05-26
Compatibility with Zod Generated Types


cgibson-swyftx:
When using a Zod validator and then passing it to TSOA, it throws this error: ` Error: No matching model found for referenced type infer.`

Types File
```
export const MyValidator = z.object({
  result: z.object({
    price: z.string().nonempty()
  }),
  code: z.number(),
  msg: z.string().nonempty()
})
export type MyResponse = z.infer<typeof MyValidator>
```

Then use it in TSO
```
 @Get()
  public async getRequest (): Promise<MyResponse> {
 
```

## Sorting

- **I'm submitting a ...**

  - [X] bug report
  - [ ] feature request
  - [ ] support request

- I confirm that I
  - [X] used the [search](https://github.com/lukeautry/tsoa/search?type=Issues) to make sure that a similar issue hasn't already been submit

## Expected Behavior

When running `yarn tsoa spec-and-routes` I expect TSOA to be able to use the inferred type generated by Zod.

## Current Behavior

It crashes with 
```
Generate routes error.
 Error: No matching model found for referenced type infer.
    at new GenerateMetadataError (/Users/caseygibson/Documents/Github/node_modules/@tsoa/cli/dist/metadataGeneration/exceptions.js:22:28)
```

## Context (Environment)

Version of the library: "^3.14.1"
Version of NodeJS: v14.17.4

- Confirm you were using yarn not npm: [X]

cgibson-swyftx:
Bump

cgibson-swyftx:
Bump

dhad1992:
bump

WoH:
I've labeled this to avoid the bot. Please vote instead of commenting, and feel free to open a PR to fix this.

therealpaulgg:
I have the same problem when using yup's inferred types.

tuchk4:
faced the same issue

direisc:
> I've labeled this to avoid the bot. Please vote instead of commenting, and feel free to open a PR to fix this.

@WoH or other person, can gives me a little help. I really like trying to solve this but I need a little help to understand where I need to looking for.

khuongtp:
I tried this and it worked (with yup) https://github.com/jquense/yup/issues/946#issuecomment-647051195

WoH:
> > I've labeled this to avoid the bot. Please vote instead of commenting, and feel free to open a PR to fix this.
> 
[...]

PR #1753, opened 2025-03-03
Add support for `typeof` keyword in types

pclements12:
### All Submissions:

- [ x ] Have you followed the guidelines in our [Contributing](https://github.com/lukeautry/tsoa/tree/master/docs/CONTRIBUTING.md) document?
- [ x ] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/lukeautry/tsoa/pulls) for the same update/change?
- [ x ] Have you written unit tests?
- [ x ] Have you written unit tests that cover the negative cases (i.e.: if bad data is submitted, does the library respond properly)?
- [ x ] This PR is associated with an existing issue?

**References Issue**

This attempts to fix at least one of the sub issues related to the ability to use Zod's `z.infer` that was identified in #1256 
https://github.com/lukeautry/tsoa/issues/1256#issuecomment-2649340573 

> 1. Typeof Name Calculation Issue
> [Error message: No matching model found for referenced type value.](https://github.com/lukeautry/tsoa/blob/master/packages/cli/src/metadataGeneration/typeResolver.ts#L949)

**Test plan**

Adds a test case to the metadata / parameter resolver for `typeof` types as a parameter or return type

Adds an (expensive) unit test for TypeResolver to explicitly enumerate all of the type combinations that the resolver supports in a clean room environment. Open to suggestions on making this more efficient by combining into a single virtual source file, but the context of type right next to test is very valuable imo.

I've also included some types in this test that still fail (currently skipped) and are related to the other issues identified in #1256

### DIFF
--- a/packages/cli/src/metadataGeneration/typeResolver.ts
+++ b/packages/cli/src/metadataGeneration/typeResolver.ts
@@ -303,6 +303,87 @@ export class TypeResolver {
       return new TypeResolver(this.typeNode.type, this.current, this.typeNode, this.context, this.referencer).resolve();
     }
 
+    if (ts.isTypeQueryNode(this.typeNode)) {
+      const isIntrinsicType = (type: ts.Type): boolean => {
+        const flags = ts.TypeFlags;
+        return (
+          this.hasFlag(type, flags.Number) ||
+          this.hasFlag(type, flags.String) ||
+          this.hasFlag(type, flags.Boolean) ||
+          this.hasFlag(type, flags.BigInt) ||
+          this.hasFlag(type, flags.ESSymbol) ||
+          this.hasFlag(type, flags.Undefined) ||
+          this.hasFlag(type, flags.Null)
+        );
+      };
+      const symbol = this.current.typeChecker.getSymbolAtLocation(this.typeNode.exprName);
+      if (symbol) {
+        // Access the first declaration of the symbol
+        const declaration = symbol.declarations?.[0];
+        throwUnless(declaration, new GenerateMetadataError(`Could not find declaration for symbol: ${symbol.name}`, this.typeNode));
+
+        if (ts.isVariableDeclaration(declaration)) {
+          const initializer = declaration.initializer;
+          const declarationType = this.current.typeChecker.getTypeAtLocation(declaration);
+          if (isIntrinsicType(declarationType)) {
+            return { dataType: this.current.typeChecker.typeToString(declarationType) as Tsoa.PrimitiveType['dataType'] };
+          }
+          if (initializer && (ts.isStringLiteral(initializer) || ts.isNumericLiteral(initializer) || ts.isBigIntLiteral(initializer))) {
+            return { dataType: 'enum', enums: [initializer.text] };
+          } else if (initializer && ts.isObjectLiteralExpression(initializer)) {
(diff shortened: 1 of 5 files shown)


WoH:
I'll get to it over the weekend!

tests/unit/swagger/definitionsGeneration/typeResolver.spec.ts
```diff
@@ -0,0 +1,512 @@
+import { expect } from 'chai';
```
WoH, line 1: I'd prefer if you could integrate the tests in this file with the existing tests

WoH: Spec Generators Tests most likely

packages/cli/src/metadataGeneration/typeResolver.ts
```diff
@@ -303,6 +303,87 @@ export class TypeResolver {
...
+              return undefined;
+            };
+
+            const isIgnored = (prop: ts.Symbol) => {
+              const declaration = getOneOrigDeclaration(prop);
+              return declaration !== undefined && getJSDocTagNames(declaration).some(tag => tag === 'ignore') && !ts.isPropertyAssignment(declaration);
+            };
+
+            const typeProperties: ts.Symbol[] = this.current.typeChecker.getPropertiesOfType(this.current.typeChecker.getTypeAtLocation(initializer));
+            const properties: Tsoa.Property[] = typeProperties
```
WoH, line 351: A lot of the property management should already exist in checker based resolutions.

Ideally we can merge them together.

tests/unit/swagger/definitionsGeneration/typeResolver.spec.ts
pclements12: This isn't really about spec-generation though. These tests are about type resolution. The spec generation tests are relevant after a type has already been resolved. None of them use the type resolver.

Currently the testing for type resolution is spread across the controller/method/parameter generator specs, but those are higher order integration tests than what I'm testing here. 
If we were going to merge them with existing tests, it would be somewhere in those. 

In my opinion the lack of easy testability for type resolutions is one of the reasons that there are quite a few type resolution bugs outstanding for the repo. This test may make it easier for folks to identify and fix those in a much more narrow focus area.

packages/cli/src/metadataGeneration/typeResolver.ts
pclements12: Can you elaborate on what you mean? Do you mean putting this code in a location that can be shared across other places where we do property management in this file?

WoH: This is taken from lines 194ff, right?

idan2468:
Hey, what is the status with this PR? @WoH @pclements12 
My team is also waiting for this PR, and it seems that quite a lot of people are also waiting for it.
Can we do something to speed up the process?

packages/cli/src/metadataGeneration/typeResolver.ts
pclements12: Yes, parts of it are taken from another section. Do you want me to extract that to a shared method is my question

pclements12: I haven't had time to spend on it lately. Happy for anyone else to contribute to clean up on this pr to get it across the line. It's functionally complete, but needs a little organizing

PR #1818, opened 2025-12-15
chore!: upgrade koa/router package

### LINKED ISSUE #1817: Upgrade @koa/router (breaking change) with v7
`@koa/router@15` now ships its own types, which are incompatible with `@types/koa__router`. 

`@types/koa__router@12.0.5`:

```ts
declare namespace Router {
  // ...
}

export = Router;
```

`@koa/router@15.0.0`:

```ts
export { /* ... */ R as Router,  /* ... */  R as default } from './layer.mjs';
```

The [code generated by the koa template](https://github.com/lukeautry/tsoa/blob/6595d721e241360c920966c275440c09bd78fdb8/packages/cli/src/routeGeneration/templates/koa.hbs#L19) does not work with the new types, so we should upgrade the import to `import type { Router as KoaRouter} from '@koa/router';` - a breaking change that we can sneak in with v7

jeengbe:
### All Submissions:

**Closing issues**

Closes #1817

This will require users to upgrade `@types/koa__router` -> `@koa/router`, which may include breaking changes outside the direct scope of tsoa - e.g. in the context or request object.

**Test plan**


jeengbe:
Perhaps I need to bump the versions for the tests to work in CI? They run just fine locally.

jeengbe:
I have no idea why this doesn't work. Have you seen this before, WoH?

jrabmer:
Hi, is there any update on this pull request? Would love to be able to upgrade to the new koa router package. 

I tried out this pull request locally on my machine (ubuntu) and the tests also seem to run fine. One thing i noticed is that the author seemed to have run `npm install` instead of `yarn install`. I am not sure if that is enough of a difference though. 

One thing that could be looked at is that the Author imports the default export of the koa router. This is `declare const Router: RouterConstructor;`. Perhaps a better import would be `type RouterInstance`. ? Or importing the default export as a value instead of a type would also maybe fix the problem

Issue #1825, opened 2026-01-03
bug: type imported via @xxx/... causes No declarations found for referenced type X


alexng353:
"@tsoa/runtime": "^7.0.0-alpha.0",

running into this error in this case:

```ts
// file 1
export enum MyEnum {
  // ...
}
```

```ts
// file 2
import { MyEnum } from "@lib/file1";

type MyBodyType = {
  enumthingy: MyEnum[];
}

// ...

@Post("/my-route")
async myFn(
  @Body body: MyBodyType
) {
  // ...
}
```

the actual error body
```sh
There was a problem resolving type of 'Share'.
Generate routes error.
 GenerateMetadataError: No declarations found for referenced type ResourcePermissions. 
 in 'SharingController.shareDirectory'
    at /home/alex/code/futurity/api/node_modules/@tsoa/cli/dist/metadataGeneration/methodGenerator.js:114:23
    at Array.map (<anonymous>)
    at MethodGenerator.buildParameters (/home/alex/code/futurity/api/node_modules/@tsoa/cli/dist/metadataGeneration/methodGenerator.js:106:14)
    at MethodGenerator.Generate (/home/alex/code/futurity/api/node_modules/@tsoa/cli/dist/metadataGeneration/methodGenerator.js:76:33)
    at /home/alex/code/futurity/api/node_modules/@tsoa/cli/dist/metadataGeneration/controllerGenerator.js:46:41
    at Array.map (<anonymous>)
    at ControllerGenerator.buildMethods (/home/alex/code/futurity/api/node_modules/@tsoa/cli/dist/metadataGeneration/controllerGenerator.js:46:14)
    at ControllerGenerator.Generate (/home/alex/code/futurity/api/node_modules/@tsoa/cli/dist/metadataGeneration/controllerGenerator.js:35:27)
    at /home/alex/code/futurity/api/node_modules/@tsoa/cli/dist/metadataGeneration/metadataGenerator.js:210:41
    at Array.map (<anonymous>)
error: "tsoa" exited with code 1
```

PR #1826, opened 2026-01-03
feat: implement tsconfig.json support for compiler options in config

alexng353:
- Added a new function to read and parse tsconfig.json for compiler
  options.
- Updated the validateCompilerOptions function to merge options from
  tsconfig.json and tsoa.json, with precedence given to tsoa.json.
- error handling for missing tsconfig.json files.
- unit tests to validate the new functionality

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -11,6 +11,7 @@ import { AbstractRouteGenerator } from './routeGeneration/routeGenerator';
 import { extname, isAbsolute, dirname, join } from 'node:path';
 import * as ts from 'typescript';
 import type { CompilerOptions } from 'typescript';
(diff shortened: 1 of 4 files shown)


alexng353:
addresses #1825 and makes devx better

WoH:
Thanks for this PR, I had this in the back of my head. Would you mind checking if we can use:

https://github.com/TypeStrong/tsconfig

or:

https://github.com/privatenumber/get-tsconfig?tab=readme-ov-file#how-can-i-use-typescript-to-parse-tsconfigjson

for the fallback?

Would be more consistent.

alexng353:
I am leaning towards the second option since we wouldn't have to add another dependency. What do you mean by "fallback," though?

alexng353:
*Left unsquashed for more easy reading of the diff

Issue #1375, opened 2023-02-01
GenerateMetadataError: Unknown type: IndexedAccessType with v5.0


Craigfis:
## Sorting

- **I'm submitting a ...**

  - [X] bug report
  - [ ] feature request
  - [ ] support request

- I confirm that I
  - [X] used the [search](https://github.com/lukeautry/tsoa/search?type=Issues) to make sure that a similar issue hasn't already been submit
I see similar issues against earlier versions of tsoa, which have been closed.

## Expected Behavior
Routes should be generated without error when running `yarn run tsoa routes`

## Current Behavior
It chokes on the following class with:
`There was a problem resolving type of 'SP<PT>'.
Generate routes error.
 GenerateMetadataError: Unknown type: IndexedAccessType`

`export interface DM{
  [PT.C]: CD;
  [PT.E]: ED;
  [PT.A]: AD;
}
export class SP<SPT extends PT> {
  public constructor(
    readonly pt: SPT,
    readonly details: DM[typeof pt],
  ) {
  }
}`

## Context (Environment)

Version of the library: 5.0.0
Version of NodeJS: 18.13.0

- Confirm you were using yarn not npm: [X]

douglasg14b:
Does this mean that any indexed types are not-supported? by TSOA at this time?

PR #1831, opened 2026-02-12
feat: add fallback resolution for indexed access types

### LINKED ISSUE #1622: GenerateMetadataError: Unknown type: IndexedAccessType
## Sorting

- **I'm submitting a ...**

  - [x] bug report
  - [ ] feature request
  - [ ] support request

- I confirm that I

I found several referneced, but they all appeard to be fixed...?

## Expected Behavior

When I tun `tsoa routes` the following controller fails:

```ts
export type BaseIVR3Config = {
    walkingSkeletonPath: WalkingSkeletonPath;
    bestMovies: BestMovies;
};

export type BaseIVR3ConfigKey = keyof BaseIVR3Config;

type WalkingSkeletonPath = 'A' | 'B';
type BestMovies = {
    title: string;
    year: number;
}[];

@Route('config2')
export class Config2Controller extends Controller {
    constructor(private useCases: ConfigUseCases) {
        super();
    }

    @Get('{key}')
    async getByKey(@Path() key: BaseIVR3ConfigKey): Promise<BaseIVR3Config[BaseIVR3ConfigKey] | 'error'> {
        const result = await this.useCases.getByKey(key);

        if (result.ok) {
            return result.value.value;
        }

        return 'error';
    }
}
```

## Current Behavior

This fails with:

> Generate routes error.
 GenerateMetadataError: Unknown type: IndexedAccessType

## Possible Solution

Unknown

## Steps to Reproduce

1. 2. 3. 4.

## Context (Environment)

Version of the library: `6.2.1`
Version of NodeJS: `18.13.0`

   - We are using `pnpm`

## Detailed Description

N/A See above

thazhemadam:
Indexed access patterns like `T[keyof U], MappedType[Union]`, and `(typeof x)[keyof typeof y]` previously threw
"Unknown type: IndexedAccessType".

Add a fallback that delegates to TypeScript's type checker to evaluate any indexed access type the compiler accepts, then recursively resolves the resulting simpler type node.

### All Submissions:

**Closing issues**

Closes https://github.com/lukeautry/tsoa/issues/1622.

This doesn't fix all forms of the `IndexedAccessTypeNode` errors. 
For instance, generic-dependent indexed access types (e.g., as in #1375) are not resolved by this change. In such cases, we bail out to avoid infinite recursion and throw the same error as before.

**Test plan**

Positive test cases:
  - `ForeignIndexedValue`: `(typeof x)[keyof typeof y]` where `x` and `y` are different objects. Asserts the schema resolves to `{ type: "string", enum: ["FOO"] }`.
  - `EventUnion`: `{ [K in Union]: ... }[Union]`, a mapped type indexed by a union alias. Asserts the schema resolves to `anyOf` with 2 object variants, each containing the expected properties.
  - `ConfigValue`: `ConfigMap[ConfigKey]` where `ConfigKey = keyof ConfigMap` (#1622 pattern). Asserts the schema resolves to anyOf with 2 variants (the heterogeneous value types of the interface).

One negative test case that documents the known limitation:

  - `GenericIndexed<PT.C>`: `DetailMap[T]` where `T` is a generic type parameter (#1375 pattern). Asserts that an error is cleanly thrown, rather than stack overflowing.


thazhemadam:
^bump for stale bot

PR #1841, opened 2026-03-14
feat: Support Type Aliases in Queries Decorator Fixes #1743

### LINKED ISSUE #1743: Support `type` instead of interface for `Queries`
My project uses [ts-rs](https://github.com/Aleph-Alpha/ts-rs) to take rust objects, and generate typescript types, to be used for an API client.

While the types are nice to have, I'd also like to be able to use `tsoa` to build an OpenAPI spec.

All was going well until I tried to use my form in a `Get` : 

```ts
/**
 * List all the media for your user
 *
 * `HTTP.GET /account/list_media`
 */
@Get("/account/list_media")
listMedia(
  @Queries() form: ListMedia = {},
  @Inject() options?: RequestOptions,
) {
  return this.#wrapper<ListMedia, ListMediaResponse>(
    HttpType.Get,
    "/account/list_media",
    form,
    options,
  );
}
```

And got this error:

`GenerateMetadataError: @Queries('form') only support 'refObject' or 'nestedObjectLiteral' types. If you want only one query parameter, please use the '@Query' decorator.`

Turns out, `tsoa` only supports interfaces, and not types, even though they are trivially the same:

```ts
// This file was generated by [ts-rs](https://github.com/Aleph-Alpha/ts-rs). Do not edit this file manually.

// Doesn't work
export type ListMedia = { page?: number; limit?: number };

// Works
export interface ListMedia {
  page?: number;
  limit?: number;
}
```

The strangest part about this, is that the return types (which are also all `type`), and `@Body()` seem to work perfectly fine. Only the `Queries()` doesn't work.

MV-GH:
closes #1743 

Supports Typescript type aliases by unrolling them for @Queries decorator

### DIFF
--- a/packages/cli/src/metadataGeneration/parameterGenerator.ts
+++ b/packages/cli/src/metadataGeneration/parameterGenerator.ts
@@ -302,58 +302,9 @@ export class ParameterGenerator {
     const parameterName = (parameter.name as ts.Identifier).text;
     const type = this.getValidatedType(parameter);
 
-    // Handle cases where TypeResolver doesn't properly resolve complex types
-    // like Zod's z.infer types to refObject or nestedObjectLiteral
-    if (type.dataType !== 'refObject' && type.dataType !== 'nestedObjectLiteral') {
-      // Try to resolve the type more aggressively for complex types
-      let typeNode = parameter.type;
-      if (!typeNode) {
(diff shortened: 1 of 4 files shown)


MV-GH:
@WoH Could someone take a look?

PR #1855, opened 2026-04-13
fix: prevent null coercion to false

### LINKED ISSUE #1820: Type order in nullable boolean types affects deserialization behavior for requests
I ran into a bug today when attempting to send a PATCH request with a property containing the following type would always cause the type to be deserialized incorrectly.

```js
export interface SomeModel {
    is_active?: boolean | null
}
```

This field would be backed by a ternary checkbox that can be set to true, false, or null explicitly. However, I noticed that all of my requests when attempting to set the value to `null` were instead being set to `false`. I confirmed that the payload being sent from the client did in fact contain `null` for the field, but then was being changed after the body was deserialized in the controller. After a bit of digging, I determined that the type needs to be both non-optional, AND that the order of the types in the union matters.

This change works as intended, although is a bit of a hack:
```js
export interface SomeModel {
    is_active: null | boolean
}
```

iffa:
When validating a union type containing both boolean and null (e.g. `boolean | null`), a null value was incorrectly coerced to false because validateBool processes null before the null subschema is ever tried.

### All Submissions:

**Closing issues**

Fixes #1820

### DIFF
--- a/packages/runtime/src/routeGeneration/templateHelpers.ts
+++ b/packages/runtime/src/routeGeneration/templateHelpers.ts
@@ -522,6 +522,19 @@ export class ValidationService {
       );
     }
 
+    // When value is null or undefined, prefer explicit null/undefined subschemas over
+    // coercive type matches. Without this, e.g. `boolean | null` with value null would
+    // coerce null to false (via validateBool) before ever reaching the null type.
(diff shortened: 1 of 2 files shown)


iffa:
Test failure seems to be flakiness, unrelated to changes here

PR #1878, opened 2026-09-01
feat: support TypeScript 6.0

### LINKED ISSUE #1877: Support TypeScript 6.0 so tsoa projects can upgrade
## Sorting

- **I'm submitting a ...**

  - [ ] bug report
  - [x] feature request
  - [ ] support request

- I confirm that I

## Expected Behavior

A project that has upgraded to TypeScript 6 can install tsoa and generate its spec and routes, the same way it can on TypeScript 5.

```ts
// tsoa should install alongside typescript@6 and generate this model as it does on TypeScript 5
interface CustomError extends Error {
  status: number;
}

type KeysMember<T = any> = {
  keys: keyof T;
};

class DefaultsClass {
  boolValue?: boolean;
}
```

## Current Behavior

`@tsoa/cli` declares `typescript: ^5.7.2` as a runtime dependency, so TypeScript 6 is not an allowed install. Beyond the version range, generation itself breaks on TypeScript 6 in three separate places:

1. **`keyof T` on a generic fails.** `TypeResolver` masks a *symbol*'s flags with `ts.TypeFlags.TypeParameter`. That only ever matched because `TypeFlags.TypeParameter` and `SymbolFlags.TypeParameter` both happened to be `262144`. TypeScript 6 renumbers `TypeFlags.TypeParameter` to `524288`, so the branch is never taken:

```
GenerateMetadataError: Could not determine the keys on T
This was caused by 'keys: keyof T;'
    at TypeResolver.resolveTypeOperatorNode (packages/cli/src/metadataGeneration/typeResolver.ts:444:15)
```

2. **Types declared only by TypeScript's libs fail to resolve.** When a name has several declarations, `getModelTypeDeclarations` drops the ones under `node_modules/typescript` so a user model shadowing a lib name wins. TypeScript 6 ships more lib declarations, so `Error` now has several lib declarations and no user declaration — the filter removes all of them:

```
GenerateMetadataError: Could not find declarations for type 'Error'.
    at calcReferenceType (packages/cli/src/metadataGeneration/typeResolver.ts:865:17)
```

3. **Generated specs silently change shape.** `MetadataGenerator` builds the program with `compilerOptions || {}`. Through TypeScript 5 an empty option bag meant `strictNullChecks: false`. TypeScript 6 enables strict null checking by default, so optional properties resolve as `T | undefined` and conditional types take the other branch. Optional `boolean` properties are emitted as `"type": "object"`, and mapped/conditional aliases resolve differently. This is the quietest of the three — no error is thrown, the spec is just wrong.

The repo's own configuration also no longer compiles on TypeScript 6: `downlevelIteration`, `baseUrl`, `esModuleInterop: false` and `moduleResolution: "node"` are now error-level deprecations (TS5101/TS5107), and `ignoreDeprecations: "6.0"` cannot be used to silence them while still supporting TypeScript 5, because TypeScript 5 rejects that value (TS5103). Separately, `@typescript-eslint` 6.16.0 only supports TypeScript `<5.4`; on TypeScript 6 its type resolution collapses and it reports ~1100 spurious errors, so the repo cannot be linted on TS 6 at all.

## Possible Solution

- Mask with `ts.SymbolFlags.TypeParameter` in the `keyof` branch — correct on every supported TypeScript version.
- Only apply the lib-declaration filter when it leaves at least one declaration behind.
- Default explicitly to `{ strictNullChecks: false }` when the user configures no `compilerOptions`, so a project that does not set them gets the same spec on TypeScript 5 and 6. Projects that do pass `compilerOptions` keep full control.
- Widen the `typescript` dependency to `^5.7.2 || ^6.0.0`, migrate the repo's tsconfigs/fixtures off the removed options, upgrade to `typescript-eslint` v8 (supports `>=4.8.4 <6.1.0`), and add `^6.0.0` to the CI matrix so both majors are covered.

## Steps to Reproduce

1. Install `typescript@6.0.3` in a project using tsoa (the `@tsoa/cli` dependency range has to be overridden to allow it).
2. Add a model with `keyof T` on a generic, or an interface extending `Error`.
3. Run `tsoa spec-and-routes`.
4. Generation fails with the errors above. Removing those models and re-running instead produces a spec where optional `boolean` properties are typed as `object`.

## Context (Environment)

Version of the library: `master` (7.0.0-alpha.0)
Version of NodeJS: 22, 24 and 25 (the versions in the CI matrix)

## Detailed Description

TypeScript 6.0 is the last release of the JavaScript-based compiler; npm's `latest` tag is now the native TypeScript 7, which does not expose the JS compiler API that `@tsoa/cli` uses to read a consumer's project. That makes TypeScript 6 the version tsoa users will be on for the foreseeable future, and today they cannot upgrade without dropping tsoa.

TypeScript 7 support is deliberately out of scope here — it is a different architecture and a much larger piece of work.

Happy to send a PR; I have the four points above implemented, with build, lint and the full test suite green on both TypeScript 5.9.3 and 6.0.3.

## Breaking change?

Not for consumers. The dependency range only widens, and the `strictNullChecks` default is chosen specifically so that projects which do not configure `compilerOptions` keep the spec they get today rather than silently changing output on a TypeScript 6 upgrade.

One judgement call worth maintainer input: projects that *do* pass `compilerOptions` with `strict`/`strictNullChecks` enabled will see optional properties resolve as `T | undefined` under TypeScript 6, because that is what their own configuration now means. The alternative — letting the TypeScript 6 defaults through and updating the expected specs — would make generated output differ between the TypeScript 5 and 6 CI legs, so I did not take it.

gabsong:
closes #1877

### All Submissions:

No new test files: the fixes are only observable on TypeScript 6, and the existing suite covers them once the CI matrix runs it there.

### What this does

TypeScript 6.0 is out and tsoa can neither build nor generate on it. This widens `typescript` to `^5.7.2 || ^6.0.3` in every package, adds `^6.0.3` to the CI matrix, and fixes what breaks:

- **`keyof T`.** A symbol's flags were masked with `ts.TypeFlags.TypeParameter`, which only ever matched because the two enums shared a value. 6.0 renumbers it, so generation failed with `Could not determine the keys on T`.
- **Lib-declared types.** 6.0's default lib declares `Error` twice, so the filter that prefers a project's own models over lib declarations of the same name removed every candidate, and `interface CustomError extends Error` failed with `Could not find declarations`. It now falls back to the lib declarations for error types only — `Map` and `Set` still fail loudly, because their lib members include computed names like `[Symbol.toStringTag]` that have no name to emit.
- **Compiler defaults.** `MetadataGenerator` keeps `strictNullChecks: false`, merged underneath whatever the caller passes. With it on, optional properties resolve as `T | undefined` and stop matching in mapped and conditional types: `Record<'id', string>` emits no properties at all, and a conditional replacing `boolean` silently leaves the original type. Everything else takes 6.0's defaults.
- **Lint.** typescript-eslint 6.16.0 supports TypeScript <5.4 and cannot lint the repo on 6.0 at all, so it moves to v8.
- **Repo config.** Options 6.0 turns into errors: `downlevelIteration`, `baseUrl`, `moduleResolution: "node"`, the `esModuleInterop` default flip, `export module` (TS1540), and a side-effect import of a controller that does not exist (TS2882).

### One known difference between the two versions

On 6.0 a model extending `Error` gains an optional `cause`, because the newer default lib declares it. That is the only difference between the specs the fixtures generate on 5.9.3 and 6.0.3. Pinning `lib` would remove it, but then anything declared in a newer lib stops resolving — a model referencing `ErrorOptions` fails outright — so 6.0's default wins.

### Test plan

The new matrix leg runs the entire existing suite on TypeScript 6, where it did not run before. Without the `keyof` and lib-declaration fixes `tests/prepare.ts` aborts before any test runs; without the `strictNullChecks` default, 40 definition-generation assertions fail.

Verified locally on both, building every package and running lint plus both suites:

| | TypeScript 5.9.3 | TypeScript 6.0.3 |
| --- | --- | --- |
| build | pass | pass |
| lint | pass | pass |
| `tests` | 1946 passing | 1946 passing |
| `tests/esm` | 2 passing | 2 passing |

### CI

One failure, `cannot post file more than default 8mb`, also fails on unmodified `master`: CI installs with `--no-lockfile`, so `multer` resolves to 2.3.0 rather than the locked 2.0.2, and 2.3.0 accepts a file of exactly `limits.fileSize` ([expressjs/multer#1407](https://github.com/expressjs/multer/pull/1407)) while the test uploads exactly that. One-line fix in the test, left out to keep this focused.


gabsong:
### CI note: one pre-existing test failure, not introduced by this PR

I ran this branch through CI on my fork (the workflow triggers `on: push`, so it runs there without needing the `action_required` approval upstream). One test fails:

```
Koa Server (with multerOpts)
  file upload
    cannot post file more than default 8mb:
  Error: Should raise error about file too large and status 500.

1945 passing, 1 failing
```

It is **not** caused by this PR. I pushed unmodified `master` (`f0f9aa79`, no changes at all) to the same fork as a control, and it fails the identical test:

| Ref | Job | Result |
| --- | --- | --- |
| this branch | `Build (25, ubuntu-latest, ^6.0.0)` | same failure (reproduced on re-run) |
| this branch | `Build (24, macos-latest, ^6.0.0)` | same failure |
| **unmodified `master` `f0f9aa79`** | `Build (25, macos-latest, ^5.0.0)` | **same failure, 1945 passing / 1 failing** |

So it reproduces on `master`, on TypeScript 5, and across both Node 24/25 and macOS/ubuntu — the OS and Node version that report it just depend on which sibling `fail-fast` cancels first.

Locally the full suite passes (1946 passing) on both TypeScript 5.9.3 and 6.0.3, which points at the environment rather than the code. My best guess at the mechanism, for whatever it is worth:

- The workflow installs with `yarn install --ignore-scripts --no-lockfile`, so CI resolves transitive dependencies fresh instead of from `yarn.lock`. A different resolved `busboy`/`@koa/multer` than the locked one (locally `busboy@1.6.0`, `@koa/multer@3.0.2`) would explain a local/CI split.
- The assertion is on an exact boundary: the test writes exactly `8 * 1024 * 1024` = 8388608 bytes and expects rejection against tsoa's default `limits.fileSize` of exactly 8388608, while the sibling test writes `8388607` and expects success. A limit check that trips on "exceeds" rather than "reaches" makes the passing/failing side of that boundary sensitive to the resolved busboy version.
- The test also still uses the removed-in-spirit `new Buffer(...)` (`DEP0005` is logged immediately before the failure).

I have deliberately not touched that test in this PR, since it is out of scope and fixing it would muddy the TypeScript 6 diff. Happy to open a separate issue for it, or to include a fix here if you would prefer that.

Aside from this one pre-existing failure, both matrix legs are green — build, lint and the full suite pass on TypeScript 5.9.3 and 6.0.3.

gabsong:
Correction to my previous comment: I guessed at `busboy` drift, and that was wrong. `busboy` resolves to `1.6.0` either way. The behaviour change is in **multer**, and I can now show it exactly.

`yarn.lock` pins `multer@2.0.2`; CI installs with `--no-lockfile`, so it resolves the current `multer@2.3.0`. Running the same upload against both, with `limits.fileSize = 8388608` (tsoa's default):

| file size | multer 2.0.2 (lockfile) | multer 2.3.0 (what CI installs) |
| --- | --- | --- |
| 8388607 (limit − 1) | 200 accepted | 200 accepted |
| **8388608 (exactly the limit)** | **500 `File too large`** | **200 accepted** |
| 8388609 (limit + 1) | 500 `File too large` | 500 `File too large` |

So multer moved the cut-off from "reject when the file *reaches* the limit" to "reject when it *exceeds* the limit". Anything genuinely over the limit is still rejected in both versions, so the upload cap itself is intact — the only case that flips is a file of exactly the limit, which is precisely what `cannot post file more than default 8mb` asserts:

```ts
writeFileSync('./moreThan8mb', new Buffer(8 * 1024 * 1024)); // === limits.fileSize
```

That makes the test dependent on multer's old off-by-one rather than on tsoa behaviour, and `--no-lockfile` is what exposes it. The one-line fix is to make the file genuinely larger than the limit, which passes on both versions:

```ts
writeFileSync('./moreThan8mb', Buffer.alloc(8 * 1024 * 1024 + 1));
```

(`Buffer.alloc` also drops the `DEP0005 Buffer() is deprecated` warning the test currently emits.)

I have left this out of this PR to keep the TypeScript 6 diff focused, and it is unrelated to these changes — it reproduces on unmodified `master` as shown above. Happy to send it as a separate PR, or to fold it in here if you would rather this branch go green.

